PR #1, opened 2026-08-27
Fix thinking-off chat template handling

rwl4:
## Summary

- Honor `enable_thinking: false` in the custom chat template.
- Match vLLM’s compatible `thinking` alias.
- Pre-close the empty `<think></think>` block and omit the reasoning-effort hint.
- Send `chat_template_kwargs` at the top level in the raw HTTP benchmark.
- Document raw HTTP versus OpenAI SDK `extra_body` behavior.
- Add regression coverage.

## Validation

- `python3 -m unittest tests/test_chat_template.py`
- Python compilation and `git diff --check`
- Rendered successfully with the checkpoint’s actual Transformers tokenizer for default thinking, both disabled-thinking flags, and explicit low reasoning effort.

Commit: eaf90d06bc8a1acd1489176d122e02b68fa6ec22

### DIFF
--- a/tests/test_chat_template.py
+++ b/tests/test_chat_template.py
@@ -1,0 +1,51 @@
+#!/usr/bin/env python3
+"""Regression checks for GLM-5.3 chat-template reasoning controls."""
+
+import unittest
+from pathlib import Path
+
+from jinja2 import Environment
+
+
+TEMPLATE = Path(__file__).parents[1] / "files" / "chat_template.jinja"
+
+
+def render_generation_prompt(**kwargs: object) -> str:
+    template = Environment(extensions=["jinja2.ext.loopcontrols"]).from_string(
+        TEMPLATE.read_text()
+    )
+    return template.render(
+        messages=[{"role": "user", "content": "hello"}],
+        tools=None,
+        add_generation_prompt=True,
+        **kwargs,
+    )
+
+
+class ChatTemplateTests(unittest.TestCase):
+    def test_thinking_defaults_on(self) -> None:
+        rendered = render_generation_prompt()
+        self.assertIn("<|system|>Reasoning Effort: Max", rendered)
+        self.assertTrue(rendered.endswith("<|assistant|><think>"), rendered)
+
+    def test_thinking_can_be_disabled(self) -> None:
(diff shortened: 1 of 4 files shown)


tests/test_chat_template.py
```diff
@@ -0,0 +1,51 @@
...
+        self.assertTrue(rendered.endswith("<|assistant|><think>"), rendered)
+
+    def test_thinking_can_be_disabled(self) -> None:
+        rendered = render_generation_prompt(enable_thinking=False)
+        self.assertNotIn("Reasoning Effort:", rendered)
+        self.assertTrue(rendered.endswith("<|assistant|><think></think>"), rendered)
+
+    def test_thinking_alias_matches_parser_behavior(self) -> None:
+        rendered = render_generation_prompt(thinking=False)
+        self.assertNotIn("Reasoning Effort:", rendered)
```
MiaAI-Lab, line 38: **[suggestion]** `test_thinking_alias_matches_parser_behavior` only renders `thinking=False` in isolation. The new template (and vLLM `Glm47MoeParser`) uses OR: if both aliases are present, either being true keeps thinking on. A later switch to AND (or a default-true `enable_thinking` injected alongside `thinking=False`) would still pass this test.

**Suggestion:** Add cases for `thinking=False, enable_thinking=False` (off) and a conflict such as `thinking=False, enable_thinking=True` (on), matching `bool(thinking) or bool(enable_thinking)` when either kwarg is set.

MiaAI-Lab (approved):
Approve and merge.

The thinking-off template path is correct: `enable_thinking` / vLLM `thinking` alias, pre-closed `<think></think>`, and top-level `chat_template_kwargs` in the raw HTTP bench. No correctness bugs. Residual risk is model-side (always-open `<think>` training), not a reason to hold this.

The one remaining suggestion is extra combined-kwargs coverage in `test_thinking_alias_matches_parser_behavior`; that can land later.

MiaAI-Lab:
Approved. Template and bench changes are correct; no blocking issues.

Merge recommendation: ship this. Residual thinking-off leakage would be model-side, not a defect in this PR. Combined-kwargs test coverage can follow in a later change.

PR #16, opened 2026-08-28
preflight: validate NCCL_IB_GID_INDEX on both nodes; document other-kit setup

AXR-Jim:
Independent reproduction on a second 2×GB10 kit (2026-08-28): decode landed in your published bands (structured 38–62 tok/s, prose 27.1) — but getting to first boot hit three kit-specific walls this PR turns into fast failures / documentation:

1. **Empty GID index kills the worker rank ~60 s in.** On our pair, gid 3 is populated on one node and all-zero on the other; the launch passed every preflight check, then died with `ibv_modify_qp` errno 61 ("No data available") deep in NCCL init. `preflight` now reads the entry on **both** devices and refuses with both GID tables printed, so the fix (pick the `::ffff:<ip>` RoCEv2 index valid on both nodes) is in hand in seconds instead of after a container autopsy.

2. **Per-node NIC vars silently win over generic NCCL env pins.** We first set `NCCL_SOCKET_IFNAME`/`NCCL_IB_HCA` (our devices are `enP2p1s0f1np1`/`roceP2p1s0f1` on *both* nodes) and watched the launcher use your kit's defaults anyway — the per-node `-e` flags come later in the docker args. `.env.example` now says so and shows how to list a kit's devices.

3. **`GPU_MEM_UTIL=0.87` needs ≥105.9 GiB free *after* vLLM's ~9 GiB init.** Nodes carrying resident services miss it by <1 GiB and fail the startup memory check; README notes 0.86/800k as the fitting fallback pair (that's what our kit runs).

No behavior change for your kit — the GID check passes trivially where the index is valid, everything else is comments/docs.

Thanks for the recipe — it took us from your tweet to a serving 800k-context brain in one evening, and these were the only three walls on the way.

https://claude.ai/code/session_01FJapN9zZpWqv2dBDNsJW9A

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -12,10 +12,20 @@ WORKER_IP=10.0.0.2
 
 # Direct CX7 QSFP (spark1 f1 ↔ spark2 f0 on this kit). NCCL cannot use the
 # 10.0.0.x loopback aliases — without these pins ncclCommInitRank hangs.
+# Other kits: names differ (e.g. enP2p1s0f1np1/roceP2p1s0f1 on both nodes of
+# some Spark pairs) — list yours with `ls /sys/class/infiniband` + `ip -br a`.
+# These per-node values are what the containers get; exporting generic
+# NCCL_SOCKET_IFNAME / NCCL_IB_HCA does NOT override them.
 HEAD_CX7_IF=enp1s0f1np1
 WORKER_CX7_IF=enp1s0f0np0
 HEAD_CX7_IB=rocep1s0f1
 WORKER_CX7_IB=rocep1s0f0
+# NCCL_IB_GID_INDEX (default 3) must be a POPULATED entry on BOTH devices —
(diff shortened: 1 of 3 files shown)


MiaAI-Lab:
Reviewed on the recipe 2×GB10 kit (spark1/spark2).

The GID preflight is the right fail-closed check: an empty `NCCL_IB_GID_INDEX` really does pass every existing probe and then die ~60s into NCCL. On this pair gid 3 is populated on both CX7 devices (`::ffff:10.0.22.1` / `::ffff:10.0.22.2`), so the new check passes and does not change launch. Docs on per-node `HEAD_CX7_*`/`WORKER_CX7_*` beating generic `NCCL_SOCKET_IFNAME`/`NCCL_IB_HCA`, and the 0.86/800k memory-fit fallback, match the footguns.

No decode-path change. Merging.

PR #20, opened 2026-08-29
Fix warm-restart JSON stdout contamination

### LINKED ISSUE #15: sitecustomize executes overlay patches at every python3 startup — pollutes command substitution (warm-restart crash)
Found on the 2x DGX Spark kit (EXL3, image ghcr.io/miaai-lab/glm-5.3-flash-2x-dgx-sparks:exl3, b~c91754f).

**Repro:** after a warm container restart, the head crashes in vLLM argparse:
```
vllm serve: error: argument --speculative-config/-sc: Value glm53: video placeholders aligned to encoder grid_t (Glm5Next→glm46v)
{"method":"dflash",...} cannot be converted to <function loads ...>
```

**Root cause:** the image ships a `sitecustomize.py` (`/usr/lib/python3.12/sitecustomize.py`) that executes `/opt/glm53/patch_glm_video_placeholders.py` (bind-mounted from `overlay/`) at **every** python3 startup. On warm restarts the patch takes the already-applied branch and prints `glm53: video placeholders aligned to encoder grid_t (Glm5Next→glm46v)` to **stdout**. `start.sh` builds `--speculative-config` via command substitution (`ARGS+=(--speculative-config "$(python3 -c 'import json,os ...')")`), so the print lands inside the JSON string → invalid JSON → crash loop under the unless-stopped restart policy.

**Proof:** `docker exec <head> python3 -c 'import json; print(json.dumps({"clean": True}))'` outputs the patch line before the JSON on a warm container; a fresh `docker run` of the image is clean (the polluter is the mounted overlay copy, not the baked one).

**Fix that worked for us:** route all overlay patch-script prints to stderr (6 files in `overlay/`), e.g. `print(..., file=sys.stderr)`. Cold start unaffected; warm restarts now parse cleanly. (One-liner per print; happy to send a PR.) Also worth auditing the other overlay patches for stdout prints for the same reason.

knapcio:
Fixes #15.

The runtime video overlay is imported through `sitecustomize`, so its status message was prepended to every `python3 -c` stdout stream. That can corrupt the JSON generated by launcher command substitutions during a warm restart.

This change:
- routes all video-overlay diagnostics to stderr;
- runs the two stdlib-only JSON command substitutions with `python3 -S` as defense in depth;
- adds a regression test that checks both properties.

Verified locally with the new regression, Python compilation, `bash -n`, and `git diff --check`. Verified on the live pinned image after mounting the patch: stdout was exactly `{"clean":true}` while the overlay status line appeared only on stderr (72 bytes).

### DIFF
--- a/overlay/patch_glm_video_placeholders.py
+++ b/overlay/patch_glm_video_placeholders.py
@@ -6,6 +6,7 @@
 # processor wrapper still routed 5.3 through _get_video_second_idx_glm4v.
 from __future__ import annotations
 
+import sys
 from pathlib import Path
 
 
@@ -92,7 +93,10 @@ def apply() -> None:
 
     Glm4vProcessingInfo._construct_video_placeholder = _construct_video_placeholder
     Glm4vProcessingInfo._glm53_video_t_aligned = True
-    print("glm53: video placeholders aligned to encoder grid_t (Glm5Next→glm46v)")
+    print(
+        "glm53: video placeholders aligned to encoder grid_t (Glm5Next→glm46v)",
+        file=sys.stderr,
+    )
 
 
 def _install_import_hook() -> None:
@@ -113,7 +117,7 @@ def _install_import_hook() -> None:
         try:
             apply()
         except Exception as exc:
-            print(f"glm53: video apply() failed: {exc!r}")
+            print(f"glm53: video apply() failed: {exc!r}", file=sys.stderr)
         finally:
             applying = False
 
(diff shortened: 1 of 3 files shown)


MiaAI-Lab (approved):
Approved. This is the right fix for #15.

Warm restart was stuffing the video-overlay status line into `--speculative-config` because `sitecustomize` / `glm53_video.pth` runs on every `python3`, and `start.sh` captured that stdout via command substitution. Routing overlay diagnostics to stderr and building the two stdlib JSON snippets with `python3 -S` both close that hole; either would have been enough.

Head and worker inner scripts are both updated. `json`/`os` do not need site. Overlay behavior is unchanged aside from the file descriptor. Static regression test is a reasonable lock given this repo has no CI.

Merge and relaunch from the updated tree (`./start.sh`). A stale `docker restart` without the new bind-mounted overlay will still crash.

MiaAI-Lab:
Merging this as the #15 warm-restart crash fix.

After merge: pull `main` and relaunch with `./start.sh` so the bind-mounted overlay and regenerated inner scripts pick up stderr routing + `python3 -S`. No image rebuild required. Cold start is unchanged; the path this unblocks is warm restart with `SPEC_METHOD=dflash`.

Issue #19, opened 2026-08-29
API stalls (10-40s) under concurrent structured output


lilianmoraru:
API stalls (10-40s) under concurrent structured output: xgrammar grammar matcher vs GLM XML tool format + DFlash2 speculative + `max_num_batched_tokens=1024` cold-prefill queue

Attached head logs and a reproduction script that can be thrown into the `tests/` directory (see the `Reproducing` section):
* [logs.txt](https://github.com/user-attachments/files/31578439/logs.txt)
* [repro_grammar_stall.py](https://github.com/user-attachments/files/31578445/repro_grammar_stall.py)

Below is the analysis done by Muse Spark 1.2:
---
## Environment

- Repo commit: `f3043c9` (`vllm/vllm-openai:glm53-flash-arm64-cu130` overlay)
- Image: `ghcr.io/miaai-lab/glm-5.3-flash-2x-dgx-sparks:exl3` (`glm53-exl3-head`/`worker`)
- vLLM: `0.1.dev20051+g487ecf187` (`logs.txt:11`)
[...]

PR #21, opened 2026-08-29
Backport XGrammar termination fix for speculative batches

knapcio:
Addresses the matcher-after-stop defect in #19.

This is a fail-closed, idempotent runtime backport of merged vLLM PR #52805 / commit `12f64b39`. It stops `accept_tokens()` and `validate_tokens()` at the first terminating token, treats later advances as harmless, and clears the cached termination flag on reset. The patch is wired to both ranks and the image build.

Tests cover exact pinned anchors, source drift, idempotence, rollback/reset behavior, both-rank launcher wiring, Python compilation, and shell syntax.

Live verification on the pinned image:
- both ranks applied the patch at startup;
- one schema request plus 16 concurrent schema requests all returned valid strict JSON with `finish_reason=stop`;
- head logs contained zero `Failed to advance FSM` / matcher-after-termination warnings;
- a second patch invocation reported the exact already-present state.

Scope is intentionally narrow: this does not reinterpret an invalid request combining GLM XML tool output with a JSON response schema, and it does not remove cold-prefill queue time.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -446,6 +446,8 @@ COPY overlay/patch_scheduler_decode_floor.py /opt/glm53/patch_scheduler_decode_f
 COPY tests/test_scheduler_decode_floor.py /opt/glm53/test_scheduler_decode_floor.py
 COPY overlay/patch_hybrid_prefix_hit.py /opt/glm53/patch_hybrid_prefix_hit.py
 COPY tests/test_hybrid_prefix_hit.py /opt/glm53/test_hybrid_prefix_hit.py
+COPY overlay/patch_xgrammar_termination.py /opt/glm53/patch_xgrammar_termination.py
+COPY tests/test_xgrammar_termination.py /opt/glm53/test_xgrammar_termination.py
 RUN python3 /opt/glm53/patch_model_overrides.py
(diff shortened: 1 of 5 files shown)


MiaAI-Lab:
Reviewed and live-checked on the pinned `exl3` image (TP=2, DFlash2 k=7).

The overlay is a source-exact backport of vLLM #52805 (`12f64b39`) onto `487ecf187`: fail-closed, idempotent, both ranks. Host tests and a throwaway container apply against the real `backend_xgrammar.py` passed.

After restarting this pair with the overlay mounted:
- head and worker both logged `patched backend_xgrammar.py (vLLM #52805: stop XGrammar token batches at termination)`
- live source has the marker, stop-at-termination loop, and `reset()` clearing `_is_terminated`
- 1 strict JSON schema request → valid `{"city":"Paris","country":"France"}`, `finish_reason=stop`
- 16 concurrent same schema → 16/16 valid JSON, all `stop`, wall 3.9s
- zero `Failed to advance FSM` / `grammar_matcher` lines on either rank

Scope still holds: this does not fix #19's cold-prefill queue vs `max_num_batched_tokens=1024`, and it does not make GLM XML tools plus a JSON `response_format` a valid request.

Merging.

PR #26, opened 2026-08-29
fix: allow a per-rank RoCEv2 GID index (HEAD_GID / WORKER_GID)

im0xMagnus:
Complements #16 rather than competing with it. That PR validates `NCCL_IB_GID_INDEX` on both nodes and tells you to pick an index valid on both — which is the right first step. This covers the case where **no such index exists**.

On our 2× GB10 pair the RoCEv2 index whose GID matches the node's own fabric IP is **4 on the head and 3 on the worker**. With a single global value one rank advertises a GID from the wrong subnet and TP init dies:

```
initialize_model_parallel -> init_model_parallel_group
RuntimeError: NCCL error: unhandled system error
```

`#16`'s preflight would correctly refuse to boot this kit; it would just have nothing to suggest. `HEAD_GID` / `WORKER_GID` give it an answer.

```bash
HEAD_GID=4
WORKER_GID=3
```

Both default to `NCCL_IB_GID_INDEX`, so leaving them unset is byte-identical to current behaviour. Implementation relies on docker honouring the last `-e` for a repeated key.

Verified on this kit: head and worker come up with `NCCL_IB_GID_INDEX=4` / `=3` respectively, TP=2 initialises, 1M context serving.

Thanks for the recipe — it took us from DS4 NVFP4 at ~10 tok/s to GLM-5.3 EXL3 at 1M context in a day.

https://claude.ai/code/session_016wA3kowHQUVCEPXWhwpw5p

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -20,12 +20,16 @@ HEAD_CX7_IF=enp1s0f1np1
 WORKER_CX7_IF=enp1s0f0np0
 HEAD_CX7_IB=rocep1s0f1
 WORKER_CX7_IB=rocep1s0f0
-# NCCL_IB_GID_INDEX (default 3) must be a POPULATED entry on BOTH devices —
-# some pairs have gid 3 all-zero on one node, which kills the worker rank
-# ~60 s in (ibv_modify_qp errno 61). preflight now verifies this. Inspect:
-#   cat /sys/class/infiniband/<dev>/ports/1/gids/{0,1,2,3}
-# and pick the index carrying the ::ffff:<ip> RoCEv2 entry on both nodes.
+# The RoCEv2 GID index is per-NIC: each node needs the index carrying its OWN
+# ::ffff:<ip> entry. An all-zero entry kills that rank ~60 s in (ibv_modify_qp
+# errno 61); preflight verifies each rank against its own device. Inspect:
+#   cat /sys/class/infiniband/<dev>/ports/1/gids/{0..7}
+# If one index works on both nodes, set NCCL_IB_GID_INDEX. If the nodes need
+# different indices (e.g. head 4 / worker 3), set them per rank instead —
+# both default to NCCL_IB_GID_INDEX when unset.
 # NCCL_IB_GID_INDEX=2
+# HEAD_GID=4
+# WORKER_GID=3
 
 # --- weights -----------------------------------------------------------
 # Durable public mirror (same bytes as brandonmusic snapshot 5ab363a8).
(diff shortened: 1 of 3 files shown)


MiaAI-Lab:
Thanks for the report and the verification on your pair — the gap is real. #16 only covers “pick one GID index that is populated on both nodes,” and some kits (yours: head=4, worker=3) have no such index.

This cannot land as-is: it conflicts with current `main` (#16 merged a few minutes after this PR opened), and even after a mechanical conflict resolve the new knobs would not boot your kit.

Please rebase onto current `main` and teach preflight the per-rank indices if you want this in this repo:

1. Validate `HEAD_GID` on the head HCA and `WORKER_GID` on the worker HCA, not a single `NCCL_IB_GID_INDEX` on both nodes.
2. Dump more than gids 0–3 when it refuses (at least through 7 — your head index is 4).
3. Set `NCCL_IB_GID_INDEX` once per container instead of repeating `-e` and relying on last-wins.
4. Update the #16 README / `.env.example` text that currently says to pick one index valid on both.

Unset, `HEAD_GID` / `WORKER_GID` should still default to `NCCL_IB_GID_INDEX` so existing kits are unchanged.

Until that rebase lands, please keep using your fork on your pair only — do not copy `HEAD_GID=4` / `WORKER_GID=3` onto a kit where both GIDs are 3.

im0xMagnus:
Rebased onto current `main` and reworked to all four points. Also found the real reason a single index cannot work here.

**The common index is RoCE v1.** Full tables from this pair (`rocep1s0f0` both nodes, fabric `10.100.128.0/24`):

| idx | head | type | worker | type |
|---|---|---|---|---|
| 1 | `fe80::…7ca2` | RoCE v2 | `fe80::…7db9` | RoCE v2 |
| 2 | `::ffff:0a64:8001` | **IB/RoCE v1** | `::ffff:0a64:8002` | **IB/RoCE v1** |
| 3 | *empty* | — | `::ffff:0a64:8002` | **RoCE v2** |
| 4 | `::ffff:0a64:8001` | **RoCE v2** | *empty* | — |

Index 2 is populated on both, so the merged non-zero check accepts it — but it is v1, and the recipe pins `NCCL_IB_ROCE_VERSION_NUM=2`. The usable v2 entries are at head 4 / worker 3. That is the case this PR covers.

**Changes**

1. preflight validates `HEAD_GID` against the head HCA and `WORKER_GID` against the worker HCA, not one index against both
2. refusal dump covers gids 0–7 and now prints each entry's `gid_attrs/types`, so a v1 entry is not mistaken for a usable one
3. `NCCL_IB_GID_INDEX` is set once per container from that rank's value — removed from `nccl_common`, added to each container's own `-e` list, no last-wins
4. README / `.env.example` updated: pick a RoCE v2 entry per node; indices need not match

`HEAD_GID`/`WORKER_GID` still default to `NCCL_IB_GID_INDEX`, so kits with one valid index are byte-identical.

**Verified on this pair:** new validation passes with head 4 / worker 3; the old single-index rule with `NCCL_IB_GID_INDEX=3` correctly refuses (head gid3 empty). Live containers run `NCCL_IB_GID_INDEX=4` / `=3` and serve 1M context at TP=2.

Not yet exercised: the refusal path on a kit that legitimately has one shared v2 index — behaviour there is unchanged by construction, but I have not run it.

PR #28, opened 2026-08-29
Honor inline MAX_NUM_SEQS override

knapcio:
## Summary

- Preserve caller-exported MAX_NUM_SEQS across .env loading, using the launcher's existing override pattern.
- Add a dependency-free regression proving caller value 4 wins over a synthetic .env value 2.

## Why

The launcher documents caller exports as higher priority than .env, but MAX_NUM_SEQS was omitted from the capture/restore list. On a live benchmark restart, MAX_NUM_SEQS=4 ./start.sh restart silently launched vLLM with max-num-seqs 2 because .env contained the older deployment value. That makes launch tuning and benchmark reproduction misleading.

## Validation

- bash -n start.sh
- python3 tests/test_start_overrides.py
- git diff --check
- Live two-DGX-Spark restart with the existing image: generated command, vLLM parsed config, and both container environments reported max-seqs 4; health passed after 510 seconds; all 20 post-ready warmups including concurrency 3 and 4 passed; both ranks had zero restarts and no fatal, FSM, or CUDA OOM log matches.

The real .env was not read or changed.

### DIFF
--- a/start.sh
+++ b/start.sh
@@ -63,6 +63,7 @@ _cli_fused="${EXL3_FUSED_MOE-}"
 _cli_image="${IMAGE-}"
 _cli_util="${GPU_MEM_UTIL-}"
 _cli_lm="${LANGUAGE_MODEL_ONLY-}"
+_cli_max_num_seqs="${MAX_NUM_SEQS-}"
 set -a
 # shellcheck disable=SC1091
 source "$SCRIPT_DIR/.env"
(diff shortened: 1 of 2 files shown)


MiaAI-Lab (approved):
LGTM — merge.

Caller exports already beat `.env` for the `_cli_*` allowlist; `MAX_NUM_SEQS` was the hole that made `MAX_NUM_SEQS=4 ./start.sh restart` launch with 2. Same capture/restore as the other knobs, and both ranks already take the restored value via docker `-e`.

Checked: unpatched `main` still prints 2; the new regression prints 4. `bash -n start.sh` is clean. Live two-node restart in the PR body already covered generated command, parsed config, both container envs, health, and C=3/4 warmups.

Not this PR: the same allowlist still misses other launch knobs (`DFLASH_TOKENS`, `MAX_MODEL_LEN`, `MAX_NUM_BATCHED_TOKENS`, `CG_ESTIMATE`).

PR #30, opened 2026-08-29
feat: optional VLLM_API_KEY bearer auth (env-var, no argv leakage)

danielbanu:
## What

Adds optional bearer-token auth to the OpenAI-compatible API via the native `VLLM_API_KEY` env var, matching the pattern used in the DeepSeek V4 Flash DSpark deployment.

## Why

The cluster is exposed through a reverse proxy on a LAN-reachable head node. Without auth, any LAN client could issue completions. `VLLM_API_KEY` is natively read by vLLM as a fallback when `--api-key` is absent on the CLI, so:

- the key never lands in `argv` or the `non-default args` startup log
- no new CLI flag plumbing needed
- empty value = unauthenticated, fully backward-compatible

## How

- `start.sh`: reads `VLLM_API_KEY` (empty default), passes it as container env to the **head** container only (worker runs `--headless`, serves no API traffic). Deliberately excluded from the generic env-loop so it doesn't show in worker process listings.
- `on_ready`: logs an `auth:` status line and includes the Bearer header in the printed quick-test curl when a key is configured.
- `/health` stays unguarded (outside vLLM's GUARDED_PREFIX `/v1|/v2|/inference|/cohere`), so boot polling and `./start.sh status` work without the key.
- `.env.example`: documents the variable.

## Testing

- Deployed live on a 2x DGX Spark (TP=2) cluster this morning: boot poll, `/health`, unauthenticated `/v1` -> 401, authenticated `/v1/chat/completions` -> 200 with the Bearer header.
- `./start.sh status` and boot-shape warmup unaffected (warmup script picks the key up via `GLM53_WARMUP_BEARER="${VLLM_API_KEY:-}"`).

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -85,6 +85,13 @@ LANGUAGE_MODEL_ONLY=0
 SKIP_MM_PROFILING=1
 LIMIT_MM='{"image":4,"video":1}'
 
+# --- API auth --------------------------------------------------------------
+# OpenAI-compatible bearer token. vLLM reads the VLLM_API_KEY env var natively
+# (fallback when --api-key is absent on the CLI), so the key never lands in
+# argv / the `non-default args` startup log. Empty = unauthenticated.
+# Same single-key semantics as the DeepSeek V4 Flash DSpark deployment.
+# VLLM_API_KEY=
+
 # --- JIT caches / boot warmup ----------------------------------------------
 # Overlay ~/.triton and ~/.tilelang die on container recreate. These host
 # dirs are bind-mounted to /root/.triton/cache and /root/.tilelang/cache.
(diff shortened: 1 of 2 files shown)


MiaAI-Lab (approved):
Opt-in only: empty `VLLM_API_KEY` is current open `/v1`. vLLM treats a missing/empty key as no auth (`if key` filter), so this is backward-compatible. `/health` stays unguarded (outside GUARDED_PREFIX), so boot poll, `./start.sh status`, and warmup keep working. Warmup already takes `GLM53_WARMUP_BEARER`.

One nit, not blocking: the body says head-only, but `serve_env` still injects the key into the worker `docker run` over SSH. The worker is `--headless` and does not need it. Fine to land as-is; a follow-up can drop that line if we want the secret off the worker SSH argv.

Issue #32, opened 2026-08-29
Field report: 2× ThinkStation PGX — 990k cold validated needle-exact, prefix page model confirmed, decode numbers


Tutanka01:
Third-kit report, in the spirit of the "Running on a different 2×Spark kit" section — **ThinkStation PGX pair**, TP=2 over CX7, `:exl3` image as pulled (no source changes, host-side tweaks only), DFlash2 k=7, 1M serve.

## Long context, cold, needle protocol

3 needles at 5/50/95%, exact sha256 match on retrieval, API healthy after every run:

| Target | Actual prompt | TTFT | Prefill e2e |
|---|---:|---:|---:|
| 200k | 200,005 | 229.5 s | 871.3 tok/s |
| 500k | 500,011 | 598.0 s | 836.1 tok/s |
| 900k | 900,007 | 1138.4 s | 790.6 tok/s |
| **990k (fresh boot)** | **990,007** | **1231.9 s** | **803.6 tok/s** |

Two things we liked: the prefill curve is basically flat (−9% from 200k to 990k — sparse-MLA doing its job), and the 900k run went through with ~1.6M tokens of earlier runs' pages still resident in the 1.75M pool. The 990k row is the full 1M window on a fresh boot.

## Decode

Your `bench_decode.py`, median of 5 × 400 tokens, temp 0, thinking off:

- structured: **66.7 tok/s** median, accept 0.959 / 6.71 per step (per-pos 1.0 / 1.0 / 1.0 / 0.98 / 0.97 / 0.90 / 0.89)
- prose: 25.2 tok/s, accept 0.305

Above your published bands — we changed nothing in the serve path, so probably kit-to-kit variance + the probabilistic draft sampler rather than anything clever on our side.

## Prefix cache

Confirmed the 3584-token page model empirically: an 8.4k-token conversation's warm turn hits **exactly 7168 = 2×3584** (hit 0.8541, efficiency 1.0 against the page model), warm TTFT 2.5s vs 10.3s true cold (**4.1×**). PR incoming with the bench we built for this — needed anyway since there's no reset endpoint on the image (separate issue).

## One trade-off data point

[...]

Issue #31, opened 2026-08-29
No cache-reset endpoint on the :exl3 image (404 everywhere) + /tokenize is root-mounted and validates `prompt`


Tutanka01:
Hey! Benchmarking the `:exl3` image on our 2× GB10 kit (ThinkStation PGX pair, image pulled 2026-08-29, zero source changes) and hit two API-surface things worth documenting.

## 1. There is no cache-reset endpoint anywhere

```
POST /reset_prefix_cache     → 404
POST /v1/reset_prefix_cache  → 404 (tried it just in case)
POST /v1/flush_cache         → 404
POST /flush_cache            → 404
```

So the only way to a genuinely cold prefix cache is a container restart. Not tragic, but it makes cold benchmarking "fun": our first prefix-cache rerun measured a "cold" 8.4k-token prefill at **1.9s vs 10.3s an hour earlier** — turned out the second run reused the same content and everything was a cache hit 🙃 Took some `/metrics` arithmetic to untangle (the `vllm:prefix_cache_*` deltas don't lie, the bench did).

**Workaround we shipped client-side** (PR to follow): salt the filler content with a per-invocation session id so every cold is genuinely cold regardless of cache state. Restart stays the strict guarantee. If exposing `/reset_prefix_cache` on this build is cheap, that'd be lovely — otherwise maybe a README note that cold runs need a restart?

## 2. `/tokenize` is mounted at the root and validates `prompt`, not `text`

`/v1/tokenize` → 404. Root `/tokenize` with `{"model", "text"}` → 400 complaining about missing `prompt` / `messages`. The working shape:

```
POST /tokenize  {"model": "GLM-5.3-Flash-EXL3", "prompt": "..."}  → {"count": N, ...}
```

Cost us a port of our long-context bench (it probed the SGLang shapes first). Documenting so the next person doesn't chase ghosts.

Happy to send logs/details if useful.

PR #33, opened 2026-08-29
tests: prefix-cache bench (page-model hit ratio, reset-less colds)

Tutanka01:
Adds `tests/bench_prefix_cache.py` — a stdlib-only client bench for the prefix-caching table in the README. We needed it reproducible on our kit, and it doubles as the measurement tool for the 3584-token page model.

## What it does

- **Unique-content cold/warm chat pairs** per run; reports cold TTFT, warm TTFT, speedup
- **Hit ratio from the `vllm:prefix_cache_{hits,queries}` counter deltas**, snapshotted around the warm turn *only* — a window covering both turns silently halves the ratio (we stepped on that one; the arithmetic to catch it: `0.4279 × (8359+8410) ≈ 7168` = exactly the 2 full pages you'd expect on the warm turn alone)
- **`hit_efficiency`** scores measured hits against the page model (`floor(tokens/3584)×3584`): 1.0 means every full page was reused. On our kit: 8.4k conversation → warm hits exactly **7168 = 2×3584**, eff 1.0, warm TTFT 2.5s vs 10.3s true cold (**4.1×**)
- **Per-invocation content salt** — since this build exposes no cache-reset endpoint (see #31), repeated runs would otherwise silently hit the previous session's pages and fake their colds (observed: identical content measured 10.3s cold then 1.9s "cold" an hour later)
- `--concurrent N` for isolated chats in flight, `--min-hit-ratio` for CI-ish gating, `--no-reset-prefix-cache` for servers that do expose an endpoint

Stdlib only. Degrades cleanly (exit 1, no traceback) when the server is unreachable; `/metrics` absence just disables the hit-ratio columns. `results/` is gitignored since the bench writes JSON artifacts there.

Verified on 2× ThinkStation PGX (field data in #32).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,6 @@ __pycache__/
 *.py[cod]
 .pytest_cache/
 
+
+# bench artifacts (prefix-cache runs write here)
+results/
--- a/README.md
+++ b/README.md
@@ -247,6 +247,20 @@ This 1M boot: **1,670,157** tokens / **1.67×** / 638 GPU blocks (padded
 slot-share still applied). The 900k process measured 1,754,237 / 690 blocks
 on the same recipe; the delta is leftover UMA, not a slot-share collapse.
 
+Re-measure (see also `tests/bench_prefix_cache.py`):
+
+```bash
+# unique-content cold/warm pairs; hit ratio from the vllm:prefix_cache_*
+# counter deltas; hit_efficiency scores against the 3584-token page model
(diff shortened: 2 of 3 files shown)


MiaAI-Lab (approved):
Bench-only: `tests/bench_prefix_cache.py` plus README/`results/` gitignore. No serve-path or overlay changes. Matches the #32 field protocol (3584 page model, warm-turn metrics window, salt because this image has no cache reset).

Issue #22, opened 2026-08-29
Deployment fixes: hf-CLI probe fails outside PATH, SKIP_DOWNLOAD not honored for DFlash2 check, + portability nits


hsrakri:
**Field notes + proposed fixes from a real 2× DGX Spark deployment (mixed-account kit).**

First — thank you for this recipe; it's the reason GLM-5.3-Flash EXL3 came up cleanly on our two Sparks. While deploying we hit a few concrete issues (items 1–3 below, prioritized) and drafted minimal patch proposals for each, plus a handful of additional robustness/portability improvements. Happy to open a PR for any of these if useful — just say which.

---

Contribution-ready improvements for the GLM-5.3-Flash EXL3 2× DGX Spark bring-up recipe. Items 1–3 were hit by a real deployer on a mixed-account 2× Spark kit and are prioritized. Each item is a minimal patch — no rewrites.

---

## 1. `hf` CLI probe hard-fails when the CLI lives outside PATH (and still runs under `SKIP_DOWNLOAD=1`)

**File/location:** `start.sh` — the `command -v hf` / `command -v huggingface-cli` probe in the preflight/download section, and the DFlash2 presence check (~line 430).

**Problem:** The script dies with `neither hf nor huggingface-cli found` when the `hf` binary is installed in a venv not on PATH (e.g. `~/.hf-cli/venv/bin/hf`) — **even when the 164 GB of weights are already present on disk**. Worse, the DFlash2 check at ~line 430 re-runs the same probe even when `SKIP_DOWNLOAD=1` was set, so the "skip downloads" escape hatch doesn't actually skip the failure.

**Proposed change:**

1. Add an `HF_BIN` override with venv-path probing and a `huggingface_hub` python fallback, resolved once near the config block:

```bash
HF_BIN="${HF_BIN:-}"
if [ -z "$HF_BIN" ]; then
    for _c in hf huggingface-cli "$HOME/.local/bin/hf" "$HOME/.hf-cli/venv/bin/hf" /opt/hf-cli/venv/bin/hf; do
        if command -v "$_c" >/dev/null 2>&1; then HF_BIN="$_c"; break; fi
    done
fi
if [ -z "$HF_BIN" ] && python3 -c 'import huggingface_hub' >/dev/null 2>&1; then
    HF_BIN="python3 -m huggingface_hub.commands.huggingface_cli"
fi
```

[...]

PR #34, opened 2026-08-29
start.sh: bring-up robustness — implements #22 items 1/2/4 + worker cache writability preflight

Tutanka01:
Implements three of @hsrakri's #22 items (the problem descriptions were spot on; here's a working implementation, field-tested on a mixed-account kit where the worker runs a different OS user), plus one new preflight from a real deploy failure. Fixes nothing else — no serve-path changes.

## 1. HF CLI resolution (#22 item 1)

`resolve_hf_bin()` probes `hf` / `huggingface-cli`, a few common venv paths, then falls back to `python3 -m huggingface_hub.commands.huggingface_cli`. `HF_BIN` overrides everything and may contain arguments. The old probe died even when the 164 GiB were already on disk and `SKIP_DOWNLOAD=1` was set.

## 2. Revision-keyed sync marker (#22 item 2)

`.glm53-exl3-synced` inside each synced repo folder, keyed on the **snapshot commit** (`refs/main`, with the same repair fallback as `ensure_refs_main`) rather than `MODEL_REVISION` — so a `MODEL_FALLBACK` adoption re-syncs correctly since the marker lives per-cache-folder. `FORCE_SYNC=1` (or deleting the marker on the worker) forces the walk. No more full size+mtime re-verification of ~164 GiB on every `./start.sh`.

## 3. Worker death detection in the /health wait (#22 item 4)

The wait loop now inspects the worker container and bails after **3 consecutive non-running answers (~30 s)** with the existing failure-log dump, instead of polling `/health` for the full `READY_TIMEOUT` while a dead worker rank makes that mathematically hopeless. Transient ssh/docker hiccups are tolerated (only consecutive failures count).

## 4. (new) Worker cache writability preflight

Hit on a real mixed-account deploy: a **root-owned `~/.cache/huggingface` on the worker** (left over from a prior sudo/docker prepare) failed the sync mid-flight with a bare `mkdir: Permission denied` — after the image was already shipped and verified. Preflight now does `mkdir -p … && test -w …` on the worker and dies early with the exact `chown` command to run.

## Also

- `.env.example` documents `HF_BIN` and `FORCE_SYNC`
- `tests/test_bringup_robustness.py` pins the anchors, same static style as `test_warm_restart_stdout.py`

No serve-path changes: container args/env are byte-identical (inner scripts and docker runs untouched). GID handling intentionally left alone since #26 covers it. Their existing text-contract tests pass (`test_start_overrides`, `test_warm_restart_stdout`, `test_xgrammar_termination`) — plus the new one.

Verified on a 2× GB10 ThinkStation PGX pair (field data in #32).

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -40,6 +40,10 @@ MODEL_REVISION=25a44fdbf16862a46b7cc9921142c6c81350af2f
 # HF cache root on the head (worker uses ${WORKER_HOME}/.cache/huggingface)
 # HF_HOME=/home/you/.cache/huggingface
 # HF_TOKEN=hf_xxx
+# HF CLI override when `hf` lives outside PATH (venv installs). May contain
+# arguments. Without it start.sh falls back to `python3 -m
+# huggingface_hub.commands.huggingface_cli` when huggingface_hub is importable.
+# HF_BIN=/home/you/.hf-cli/venv/bin/hf
 
 # --- serve -------------------------------------------------------------
 # Public GHCR image. ./start.sh docker-pulls this on the head, then on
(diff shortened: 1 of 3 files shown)


MiaAI-Lab (approved):
Bring-up only: HF CLI resolution, revision-keyed worker sync marker, fail-fast on a dead worker rank, and writable-cache preflight. Inner scripts / docker serve args are unchanged.

One operational note: `.glm53-exl3-synced` is a snapshot-hash match, not a shard count. A corrupted worker cache that still has the marker will skip the 164 GiB rsync — `FORCE_SYNC=1` (or delete the marker) is the escape hatch. Acceptable for this recipe.

PR #37, opened 2026-08-29
expose /reset_prefix_cache on the head API (GLM53_EXPOSE_CACHE_RESET overlay, #31)

### LINKED ISSUE #31: No cache-reset endpoint on the :exl3 image (404 everywhere) + /tokenize is root-mounted and validates `prompt`

Tutanka01:
Closes #31 (the cache-reset half in code; the `/tokenize` half turns out to be stock behavior of this vLLM, so it's documented instead — details below).

## What was actually wrong

The route exists in this build! `vllm/entrypoints/serve/dev/cache/api_router.py` ships `/reset_prefix_cache`, `/reset_mm_cache` and `/reset_encoder_cache` — but `build_app` only mounts it behind `VLLM_SERVER_DEV_MODE=1`:

```python
if envs.VLLM_SERVER_DEV_MODE:
    from vllm.entrypoints.serve import register_vllm_dev_api_routers

    register_vllm_dev_api_routers(app)
```

…and dev mode would also drag in the rest of the dev surface (`/sleep`, `/rlhf`, `/rpc`, `/server_info`) plus its own security warning. Not a trade we want on a network-facing bench server, so the endpoint sat invisible and cold benchmarking needed a restart.

## The fix

`overlay/patch_cache_reset.py`, same house style as the other overlays (fail-closed, idempotent, single anchored replace, `compile()`-checked):

- adds a narrow `elif os.getenv("GLM53_EXPOSE_CACHE_RESET", "0") == "1":` branch to `build_app` that attaches **only** the cache-reset router
- `VLLM_SERVER_DEV_MODE=1` keeps precedence (full dev set, as upstream)
- `GLM53_EXPOSE_CACHE_RESET=0` / unset → stock surface, restart stays the only reset

`start.sh` wires it like the others: `CACHE_RESET_PATCH_HOST` override, runtime `:ro` mounts, both inner scripts, flag forwarded through the shared env block (default `1` — the prefix-cache bench is a first-class workflow here, and a cache flush is mild next to what an unauthenticated caller can already do to an open `:8888`; the README says to set `0` on shared kits). The Dockerfile bakes the patch and runs the new test in the build verify chain.

## Auth caveat worth knowing (upstream, not ours)

While tracing this I noticed `AuthenticationMiddleware` only guards `GUARDED_PREFIX = ("/v1", "/v2", "/inference", "/cohere")`. Root-mounted routes — stock `/tokenize`, `/detokenize`, and now `/reset_prefix_cache` — answer without the bearer even with `VLLM_API_KEY` set. Documented in the README; flagging it here in case it deserves an upstream issue.

## The `/tokenize` half of #31

No code change needed: this is stock for this vLLM version. It's mounted at the root (`/v1/tokenize` is genuinely 404) and the request model validates `prompt` (or `messages` for the chat shape), not `text`. The README now documents the exact working shape so the next person doesn't port against SGLang shapes and chase ghosts:

```bash
curl -s http://127.0.0.1:8888/tokenize -H 'Content-Type: application/json' \
     -d '{"model": "GLM-5.3-Flash-EXL3", "prompt": "hello world"}'
# -> {"count":2,"max_model_len":1000000,"tokens":[14978,1879],"token_strs":null}
```

## Receipts (2× GB10 ThinkStation PGX pair, stock `:exl3` image + runtime overlay, 2026-08-29)

Before (stock image, the exact four probes from #31): all `404`.
After `./start.sh restart`:

```
POST /reset_prefix_cache                        -> 200 {"success":true}
POST /reset_prefix_cache?reset_external=true    -> 200 {"success":true}
POST /reset_mm_cache                            -> 200
```

And the salt workaround from #33 now gets the real thing — `tests/bench_prefix_cache.py` auto-detects the endpoint (`reset=on`) and colds stay cold across reruns, no session salt needed:

```
  run 1 chat 0: cold=10.818s warm=9.202s speedup=1.176 hit=0.8196 eff=1.0 coldhit=0.0 prompt_tokens=8711
  run 2 chat 1: cold=10.726s warm=2.806s speedup=3.823 hit=0.8196 eff=1.0 coldhit=0.0 prompt_tokens=8711
  run 3 chat 2: cold=10.73s  warm=3.037s speedup=3.533 hit=0.8196 eff=1.0 coldhit=0.0 prompt_tokens=8711
```

Cold TTFT ~10.7 s on every rerun (the contaminated "cold" from #31 was 1.9 s), `coldhit=0.0` before every cold turn — the reset demonstrably drains the page pool. Run 1's warm is post-boot JIT noise; runs 2–3 are steady state.

One infra note: `BUILD=1` pulls `vllm/vllm-openai:glm53-flash-arm64-cu130`, which isn't reachable outside your build infra, so external kits validate via the runtime-overlay path (that's what the receipts above use). The build path itself is exercised by the same test file the Dockerfile runs — I executed it inside the container against the installed file, green.

## Tests

New `tests/test_cache_reset_endpoint.py` (same static style as `test_xgrammar_termination.py`): pins the exact `build_app` anchor, asserts the flag semantics by actually exec'ing the patched `build_app` against fake modules (off / on / dev-precedence), idempotence, merged-upstream acceptance, fail-closed on drifted/partial/duplicated anchors, the installed-file check, and the launcher/Dockerfile wiring.

All existing text-contract tests stay green: `test_start_overrides`, `test_warm_restart_stdout`, `test_xgrammar_termination`, `test_bringup_robustness`.

GID handling untouched (still #26's territory).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -448,6 +448,8 @@ COPY overlay/patch_hybrid_prefix_hit.py /opt/glm53/patch_hybrid_prefix_hit.py
 COPY tests/test_hybrid_prefix_hit.py /opt/glm53/test_hybrid_prefix_hit.py
 COPY overlay/patch_xgrammar_termination.py /opt/glm53/patch_xgrammar_termination.py
 COPY tests/test_xgrammar_termination.py /opt/glm53/test_xgrammar_termination.py
+COPY overlay/patch_cache_reset.py /opt/glm53/patch_cache_reset.py
+COPY tests/test_cache_reset_endpoint.py /opt/glm53/test_cache_reset_endpoint.py
 RUN python3 /opt/glm53/patch_model_overrides.py
 RUN python3 /opt/glm53/patch_dflash2.py
 RUN python3 /opt/glm53/patch_glm_eagle3.py
@@ -456,9 +458,11 @@ RUN python3 /opt/glm53/patch_suppress_stops_in_reasoning.py
 RUN python3 /opt/glm53/patch_scheduler_decode_floor.py
 RUN python3 /opt/glm53/patch_hybrid_prefix_hit.py
 RUN python3 /opt/glm53/patch_xgrammar_termination.py
+RUN python3 /opt/glm53/patch_cache_reset.py
 
 RUN EXL3_SELFCHECK_GPU=0 python3 /opt/glm53/test_exl3_overlay.py \
     && python3 /opt/glm53/test_suppress_stops.py \
     && python3 /opt/glm53/test_scheduler_decode_floor.py \
     && python3 /opt/glm53/test_hybrid_prefix_hit.py \
-    && python3 /opt/glm53/test_xgrammar_termination.py
+    && python3 /opt/glm53/test_xgrammar_termination.py \
+    && python3 /opt/glm53/test_cache_reset_endpoint.py
--- a/README.md
+++ b/README.md
@@ -262,8 +262,43 @@ python3 tests/bench_prefix_cache.py --runs 3
 Note the page math: hits are **block-aligned to the 3584-token hybrid MLA
 page**, so a warm prompt only ever reuses `floor(tokens / 3584) × 3584`
 tokens — the 7168 / 10752 / 14336 hit rows above are exactly 2 / 3 / 4 full
-pages. And since this build exposes **no cache-reset endpoint**, the bench
-salts its filler content per invocation so every cold is genuinely cold.
+pages. The bench resets the prefix cache between colds
+(`POST /reset_prefix_cache`, see the API surface notes below) and salts its
+filler content per invocation on top — repeated runs stay genuinely cold
+even if the reset route is disabled.
+
+## API surface notes (this build, 2026-08-29)
+
+Two things that cost us time (#31), documented so the next person does not
+chase ghosts:
+
+**Cache reset.** `overlay/patch_cache_reset.py` mounts the upstream dev
+cache router on the head API server, so a genuinely cold prefix cache no
+longer needs a container restart:
+
+```bash
+curl -s -X POST http://127.0.0.1:8888/reset_prefix_cache    # -> {"success": true}
+```
+
+It returns `{"success": bool}` and reports `false` while blocks are still
+held (running requests, in-flight async KV offload) — retry after they
+drain. `GLM53_EXPOSE_CACHE_RESET=0 ./start.sh` restores the stock surface
+(restart is then the only reset); `VLLM_SERVER_DEV_MODE=1` still mounts the
+whole dev set (`/sleep`, `/rlhf`, `/rpc`, `/server_info`) if ever needed.
+Auth caveat: the bearer middleware only guards `/v1`, `/v2`, `/inference`,
+`/cohere` (upstream `GUARDED_PREFIX`), so root-mounted routes — the stock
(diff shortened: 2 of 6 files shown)


plotarmordev:
### Review follow-up — `09e172915794`

The cache-reset feature is useful; the hold is for default safety and integration, not missing performance numbers.

Please make the feature opt-in by default and align the launcher/README with that choice. Its root-mounted routes are outside the normal bearer-auth boundary, so the current launcher default should not enable them implicitly on keyed/shared deployments. Also rebase the wiring onto main's ordered-overlay/artifact-validation path rather than retaining a parallel set of copy/mount checks.

Next action: author source changes; we then check the rebased overlay and both-rank wiring. No new GPU benchmark is requested.

plotarmordev (approved):
Exact-head source review approved. Cache reset defaults off, independent DEV_MODE precedence and root-route bearer caveat remain explicit, ordered overlay/artifact handling and unconditional both-rank gate contracts are retained, and redundant source pins are removed. Main supplied CPU fixture/parity receipts; the final test-only cleanup leaves launcher/parity bytes unchanged. Installed vLLM routes and live auth were not exercised. Author ancestry preserved by normal integration; no merge performed. Main owns cumulative safechecks and SIX-GlmLandingB alone owns landing.

plotarmordev (approved):
Reviewed cumulative normal integration with current main, preserving allocator/host-template and fair-v5 behavior. Main exact-head CPU reset and rank-parity receipts pass; incidental source/default/mount-count assertions removed, actual safety and rank behavior retained. Full outgoing content and tree reviewed; no owner-local deployment changes. Repository has zero configured workflows (not a green or unavailable CI run). Qualification is bounded to inert CPU fixtures, not actual installed vLLM/API/auth/node or GPU proof. Source and cumulative CPU gates satisfied for normal merge.

PR #38, opened 2026-08-29
start.sh: validate serving numeric knobs before restart stops the service

hecisaza:
## What

`GPU_MEM_UTIL`, `MAX_MODEL_LEN`, `MAX_NUM_SEQS`, and `MAX_NUM_BATCHED_TOKENS` are
assigned defaults and passed straight into both rank commands with **no type,
range, or normalization check**. A typo fails deep in vLLM — and because
`./start.sh restart` runs `stop` before any vLLM parse error can surface, a bad
value **takes down a healthy 2-node deployment before it is diagnosed**.

## Fix

`validate_numeric_config()`, called at the top of the `start|restart` dispatch,
**before** `restart) stop; start`. Invalid serving config is rejected while the
service stays up; `stop`/`status`/`logs`/`help` remain usable. Integer knobs are
normalized (leading zeros stripped as strings, oversized digit strings rejected
before bash arithmetic can overflow), so no unbounded input reaches `$(( ))`.

Proposed contract (happy to adjust bounds):

| Variable | Accepted |
|---|---|
| `GPU_MEM_UTIL` | decimal `0 < v <= 1` |
| `MAX_MODEL_LEN` | integer 1–1,000,000 (the recipe's documented 1M ceiling) |
| `MAX_NUM_SEQS` | integer 1–4,096 |
| `MAX_NUM_BATCHED_TOKENS` | integer 1–8,388,608 |

The two larger maxima follow the accepted DeepSeek recipe precedent
(MiaAI-Lab/DeepSeek-v4-Flash-DSpark-2x-DGX-Spark#124) — happy to tune to whatever
GLM contract you prefer.

## Tests

`tests/test_numeric_config.py`: accepted/rejected matrix incl. `.87`, leading-zero
normalization (`01000000`→`1000000`), an **actual carriage-return byte**, a
2^64-1 overflow string, and a wiring assertion that validation precedes the
`restart) stop`. Mutation-checked: deleting the wiring fails the suite.
`bash -n` clean; 0 new ShellCheck findings vs `main`.

## Notes

Rebased onto current `main` (applies cleanly post-#34). Kept separate from a
memory-preflight change I'm preparing so each lands independently.

### DIFF
--- a/start.sh
+++ b/start.sh
@@ -226,6 +226,40 @@ log()  { printf '\033[1;36m[glm53-exl3]\033[0m %s\n' "$*"; }
 warn() { printf '\033[1;33m[glm53-exl3]\033[0m %s\n' "$*" >&2; }
 die()  { printf '\033[1;31m[glm53-exl3]\033[0m ERROR: %s\n' "$*" >&2; exit 1; }
 
+# GLM53 numeric config guard (begin)
+_glm53_canonical_positive_int() {
+    local name="$1" value="$2" maximum="$3" canonical
+    if ! [[ "$value" =~ ^[0-9]+$ ]]; then
+        echo "$name must be a positive base-10 integer (got: $value)" >&2
+        return 2
+    fi
+    canonical="$value"
+    while [ "${canonical#0}" != "$canonical" ]; do canonical="${canonical#0}"; done
+    [ -n "$canonical" ] || canonical=0
+    if [ "$canonical" = 0 ] \
+       || [ "${#canonical}" -gt "${#maximum}" ] \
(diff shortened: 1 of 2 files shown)


MiaAI-Lab (approved):
Verified against current main: patch applies (hunk offsets only), `bash -n start.sh` clean, `tests/test_numeric_config.py` PASS, plus existing `test_start_overrides` / `test_bringup_robustness` / `test_warm_restart_stdout`. Guard runs on start|restart before `stop; start`, so a bad knob cannot take down a healthy pair. Integer overflow/CR/leading-zero cases covered. Safe to merge.

PR #39, opened 2026-08-29
preflight: check unified memory vs GPU_MEM_UTIL before bring-up

hecisaza:
## What

`preflight()` checks disk, ports, and *containers* on the worker, but not **free
unified memory**. On GB10 the GPU shares the ~122 GiB pool with the host, so a
host process holding memory (**not** a container — the existing "other
containers" check misses it) leaves less than `GPU_MEM_UTIL × MemTotal` free. The
worker rank then dies **mid-bring-up** with a `ValueError: Free memory on device
... less than desired` — *after* the image pull, weight load, and sync, in a
hard-to-read worker log.

## Fix

`preflight_memory()`, run **after the port checks, before image/download/sync**:
- parses `MemTotal` + **`MemAvailable`** from `/proc/meminfo` (KiB),
- compares **raw KiB** against `GPU_MEM_UTIL × MemTotal + headroom` on **both** nodes,
- fails early with a node-specific, actionable message.

`MemAvailable` (not `MemFree`) so it does not false-fail after a large weight
`rsync` fills reclaimable page cache — that distinction is load-bearing on this
recipe. Headroom defaults to 2 GiB, overridable via
`GLM53_PREFLIGHT_MEMORY_HEADROOM_KIB`.

## Open questions for you

- **Headroom size** — 2 GiB is provisional. A healthy boot here shows ~3.7 GiB
  slack over the request on both nodes, but one boot isn't a universal margin.
- **Fatal vs override** — currently fatal (with the env override above). Given the
  failure is delayed and expensive, fatal-by-default seemed the more useful
  policy, but happy to make it a warning if you prefer.
- Known limitation documented, not fixed: a time-of-check/time-of-use window
  (another process can allocate after preflight); swap is not counted as UMA.

## Honesty note

This establishes the gap by the **formula + unguarded pass-through + missing
check**, plus a healthy boot's own `Free memory on device (~109.6/121.69 GiB)`
line matching `util × total`. It does **not** claim to reproduce a specific
prior OOM.

## Tests

`tests/test_preflight_memory.py`: ordinary matrix, boundary cases (one KiB under
threshold, exact-equal), `MemAvailable`-vs-`MemFree` fixture, and a wiring
assertion that the check runs after port checks. Mutation-checked: deleting the
calls fails the suite. `bash -n` clean; 0 new ShellCheck findings vs `main`.

## Notes

Rebased onto current `main` (applies cleanly on top of the merged #34, which also
touched `preflight()`). Separate from #38 (numeric-knob validation) so each lands
independently.

### DIFF
--- a/start.sh
+++ b/start.sh
@@ -292,6 +292,48 @@ check_port_free() {
     fi
 }
 
+# GLM53 preflight memory guard (begin)
+read_meminfo_kib() {
+    local source_file="${1:-/proc/meminfo}"
+    awk '
+      /^MemTotal:/ { total=$2 }
+      /^MemAvailable:/ { available=$2 }
+      END {
+        if (!total || !available) exit 1
+        print total, available
+      }
+    ' "$source_file"
+}
+
+preflight_memory() {
+    local label="$1" total_kib="$2" available_kib="$3" util="$4"
+    local headroom_kib="${GLM53_PREFLIGHT_MEMORY_HEADROOM_KIB:-2097152}"
+    local total_gib available_gib requested_gib headroom_gib
+
+    if ! [[ "$total_kib" =~ ^[0-9]+$ && "$available_kib" =~ ^[0-9]+$ && "$headroom_kib" =~ ^[0-9]+$ ]]; then
+        echo "PREFLIGHT FAIL [$label]: invalid memory reading" >&2
+        return 2
+    fi
+    if ! [[ "$util" =~ ^(0([.][0-9]+)?|[.][0-9]+|1([.]0+)?)$ ]] \
+       || ! awk -v u="$util" 'BEGIN { exit !(u > 0 && u <= 1) }'; then
+        echo "PREFLIGHT FAIL [$label]: GPU_MEM_UTIL must be greater than 0 and at most 1: $util" >&2
+        return 2
+    fi
+
+    total_gib=$(awk -v k="$total_kib" 'BEGIN { printf "%.2f", k/1048576 }')
+    available_gib=$(awk -v k="$available_kib" 'BEGIN { printf "%.2f", k/1048576 }')
+    requested_gib=$(awk -v t="$total_kib" -v u="$util" 'BEGIN { printf "%.2f", (t*u)/1048576 }')
+    headroom_gib=$(awk -v k="$headroom_kib" 'BEGIN { printf "%.2f", k/1048576 }')
+
+    if ! awk -v a="$available_kib" -v t="$total_kib" -v u="$util" -v h="$headroom_kib" \
+        'BEGIN { exit !(a >= (t*u)+h) }'; then
+        echo "PREFLIGHT FAIL [$label]: MemAvailable=${available_gib} GiB of ${total_gib} GiB; GPU_MEM_UTIL=${util} requests ${requested_gib} GiB plus ${headroom_gib} GiB headroom" >&2
+        return 1
+    fi
+    echo "PREFLIGHT OK [$label]: MemAvailable=${available_gib}/${total_gib} GiB; request=${requested_gib} GiB plus ${headroom_gib} GiB headroom"
+}
+# GLM53 preflight memory guard (end)
+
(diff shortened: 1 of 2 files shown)


plotarmordev:
### Review follow-up — `d5fa7d3ff7c5`

The unified-memory preflight addresses a real missing check, and using MemAvailable is appropriate. Your documented heuristic limits and healthy-boot observation are acknowledged.

The unresolved choice is ours: fatal versus warning-only behavior, and the proposed 2 GiB headroom. We have not asked you for further evidence, so the premature `awaiting-test-evidence` label is being removed. The merge hold now means a maintainer policy decision, not an unanswered author request.

No action is required from you until we choose that policy. The implementation check should be the scoped CPU boundary/validation tests; a new GPU experiment is not a prerequisite merely to review this guard.

plotarmordev (approved):
Exact-head review: `39c677908bb637988135c67f83631701c9b705ed`.

Approved for the scoped fatal unified-memory preflight. The maintainer policy decision is resolved: retain fatal refusal on either node and the existing tunable 2 GiB headroom; no warning-only mode and no new contributor/GPU evidence request. `MemAvailable` accounts for current kernel-held/unavailable memory more appropriately than `MemFree`, but remains an estimate: this is conservative admission, not reservation, a device-memory measurement, or an OOM guarantee; the allocation-time race remains.

Reviewed the actual guard/parser, both-node call sequence, and carried boundary tests. The comparison uses raw KiB before display rounding. Helper numeric/util failures return 2; shortage returns 1; missing/zero/unreadable meminfo follows the existing fatal read-error path (1), not a universal status-2 claim. Main supplied successful CPU receipts for integer/fractional/custom-headroom boundaries, invalid input/read failures, and an inert full-launcher refusal trace on either node with no image/download/sync/launch/removal stages. The healthy fixture intentionally reaches a fake download and stops there; it is not a successful serving boot. I reviewed these receipts rather than rerunning tests in this concurrent wave.

Publication: plain fast-forward of the unchanged original source branch from `d5fa7d3ff7c5a1794122e541bb3a09d5d320228c`; author history preserved, no force/rebase/squash. The source-order text assertion has been removed, so the original PR body’s mutation-checked wiring-test statement is historical and does not describe this head; the inert action trace is the supplied ordering evidence. Full publishing worktree and entire pushed range were scanned for secrets/deployment-local content.

No merge performed. Current main `e6e7869d589698e7f3d90e4bbe766f49e3870708` is newer than the integrated/CPU-checked anchor; read-only merge-tree inspection is textually clean but is not cumulative verification. SIX-GlmLandingB owns the sole GLM merge lane and Main owns any cumulative exact-merged-head safechecks before landing. No live endpoint, container, SSH, GPU, model, or service operations were used for this review.

plotarmordev (approved):
Approved cumulative head `84d78e818db39495abe8ec7074e78da5eef5ff37`, normal integration of current main `a9bbbadc4c7253870d0d519c3d445e4ee7b97540` with the previously approved PR39 head `39c677908bb637988135c67f83631701c9b705ed`. Full delta against current main remains only the 52-line memory guard/call addition plus its CPU boundary test; external #186/#187 and merged #70/#41/#75 are ancestors and their content is preserved.

Fresh exact-head checks passed: bash syntax, tests/test_preflight_memory.py, fractional/custom-headroom/invalid-input helper cases, and the inert full-launcher admission harness. Both head and worker shortage cases exit 1 with zero image/download/sync/launch/removal stage records. The healthy fixture reaches stubbed image/download work and intentionally exits 1 at fake download; this is not live boot qualification. No real container, SSH, model or node operations were executed.

Secret/content gate: scanned the whole clean worktree and full publication range; only existing upstream examples/historical paths and fixture credentials matched, with no new owner-local deployment content. Fatal refusal and tunable 2 GiB default policy remain approved. No unresolved implementation request at this head.

Hosted CI is not green: repository reports zero workflows and this PR has no check runs. Scoped local evidence is the landing evidence; account-capacity authorization was not needed to bypass a failed check. Main owns the follow-up to rerun applicable CI against merged main when hosted capacity/checks become available. Conservative admission/TOCTOU limitations remain; no OOM guarantee or pre-stop restart memory validation is claimed.

PR #42, opened 2026-08-29
fix: make container health checks pipefail-safe

jmporchet:
## Summary
- Prevent `pipefail` from treating running Docker containers as stopped when `grep -q` closes early.
- Apply the safe state checks to head and worker startup monitoring.
- Add a regression test for the pipefail-sensitive checks.

## Verification
- `bash -n start.sh start-cluster-watchdog.sh`
- `python3 tests/test_bringup_robustness.py`
- `python3 tests/test_start_overrides.py`
- `python3 tests/test_warm_restart_stdout.py`
- Live TP=2 launch completed successfully; API health and chat endpoint verified.

## Notes
- The local `.env_bak` and `start-cluster-watchdog.sh` files were intentionally not included.


knapcio:
The pipefail-safe `docker inspect` substitutions look good, but the adjacent `check_port_free` escape change introduces a regression.

Inside the existing double-quoted shell string, `"[:.]${port}\$"` passes `[:.]8000$` to grep, where `$` is the end anchor. The proposed `"[:.]${port}\\$"` passes `[:.]8000\$`, which matches a literal dollar sign instead. A normal listener therefore stops matching:

```text
sample: 0.0.0.0:8000
old pattern rc=0
new pattern rc=1
```

I reproduced this with the exact `awk | grep -qE` pipeline on current main. Please keep the original single-backslash pattern and add a functional assertion that a representative `0.0.0.0:8000` or `[::]:8000` listener is detected. The container-state changes can then remain isolated to the pipefail fix.

jmporchet:
Good catch. Restored the single-backslash end-anchor in `check_port_free` (`f038622`) and added `test_check_port_free_detects_representative_listeners`, which pins the source escape and asserts `0.0.0.0:8000` / `[::]:8000` match through the same bash double-quote expansion. Pipefail container-state checks are unchanged.

jmporchet:
Follow-up on the "head container exited" / `No such container` failure.

That message was not a pipefail false negative on a live container. Journal + a local repro show a concurrent `docker rm` (second `start.sh stop`/`restart`, or systemd `ExecStop` racing ahead of KillMode) deleting the containers under `wait_for_health`. `collect_failure_logs` then prints exactly `No such container`.

Pushed `35ecb5a`: flock across start/restart, stop steals that lock, head gets the same 3-strike inspect window as the worker, and the watchdog refuses to launch over an in-flight start. Local systemd unit now uses `ExecStopPost=./start.sh stop` instead of `ExecStop` (already `daemon-reload`'d here).

Live check after the recovery start: health passed at 480s, API healthy on :8888, both containers up.

plotarmordev:
### Review follow-up — `35ecb5a495cc`

Your follow-up distinguished the concurrent container-removal race from the original pipefail hypothesis, and the earlier port-matching correction is acknowledged.

Please remove start-cluster-watchdog.sh from this PR: it contains deployment-specific paths/node identity and is separate from the launcher race fix. Rebase the locking/health-check changes onto current main, keeping unrelated lifecycle work out.

We own the decision on stop-versus-in-flight-start termination semantics. Before merge, the rebased change needs scoped start/stop/race checks; neither the original title nor a syntax check alone establishes those semantics. The hold is for these source/integration issues, not an unspecified performance request.

plotarmordev (approved):
Exact-head source review approved for the decided TP2 bounded-refusal policy. Kernel flock ownership gates mutations; start/restart fail fast, stop waits up to 30 seconds then refuses, no advisory PID is signalled, and the log follower closes fd 9. Watchdog is removed, inspect substitutions and the resetting head three-miss window remain. Main supplied corrected isolated lifecycle/health evidence with an owned decoy and a 30.01-second refusal receipt; these are safe seams, not live container or GPU qualification. Per-checkout/TP2 limits remain, and current-main textual integration is clean but cumulative runtime proof belongs to Main and the sole landing owner SIX-GlmLandingB. No merge performed.

plotarmordev (approved):
Approved exact cumulative 0e7c50fca98548e66fbd190547ca35163c7c3af6. Source review preserves per-checkout TP2 kernel-flock ownership, fail-fast start/restart, bounded stop refusal without PID signalling, log follower fd closure, pipefail-safe inspection, and resetting three-miss health detection. Main exact-head lifecycle, launcher parity and extra-env CPU fixtures pass; shipped stop bound refused at 30.02 seconds with owned holder alive, metadata/inode intact and no recorded docker/SSH calls. Incidental log-wording assertions removed; behavioral checks retained. Full tree and outgoing-object content/provenance gate reviewed; synthetic fixture credential only, no owner-local deployment additions. Repository has zero configured workflows, not a green hosted CI result. This does not qualify actual nodes, containers or GPUs; Main authorized normal merge under those limits.

PR #44, opened 2026-08-30
Honor remaining inline launch overrides

knapcio:
## Summary

- Preserve caller-exported `DFLASH_TOKENS`, `MAX_MODEL_LEN`, and `CG_ESTIMATE` across `.env` loading using the established launcher pattern.
- Retain regression coverage for `MAX_NUM_BATCHED_TOKENS`, whose capture/restore landed independently in #40 while this follow-up was being prepared.
- Add a dependency-free table-driven regression proving all four values beat conflicting synthetic `.env` values, including the operationally important caller `CG_ESTIMATE=0`.

## Why

The review on #28 noted that these launch knobs still fell outside the caller-export allowlist. Silent `.env` precedence makes one-off tuning and benchmark reproduction misleading.

Current main already fixes the batch-token case through #40, so this PR does not duplicate that code. It adds only the three remaining capture/restores and tests the complete four-knob contract.

## Validation

- `python3 tests/test_start_overrides.py`
- `python3 tests/test_bringup_robustness.py`
- Existing dependency-free chat-template, suppress-stops, warm-restart, and XGrammar tests
- `bash -n` on every tracked shell file
- In-memory compilation of all 26 tracked Python files
- `git diff --check`

No live restart was performed: this is launcher precedence only, the downstream rank plumbing is unchanged, and #28 already live-validated the same capture/restore mechanism. The real `.env` was not read or changed.

### DIFF
--- a/start.sh
+++ b/start.sh
@@ -67,6 +67,9 @@ _cli_image="${IMAGE-}"
 _cli_util="${GPU_MEM_UTIL-}"
 _cli_lm="${LANGUAGE_MODEL_ONLY-}"
 _cli_max_num_seqs="${MAX_NUM_SEQS-}"
+_cli_dflash_tokens="${DFLASH_TOKENS-}"
+_cli_max_model_len="${MAX_MODEL_LEN-}"
+_cli_cg_estimate="${CG_ESTIMATE-}"
 set -a
 # shellcheck disable=SC1091
 source "$SCRIPT_DIR/.env"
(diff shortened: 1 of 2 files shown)


plotarmordev:
Closing as subsumed by #92, which fixes the whole caller-export-vs-.env class rather than three more inline override names. Thank you for pushing on launch-override consistency — the general rule is the better fix. Reopen if #92 doesn't land or leaves a gap this one covers.

PR #59, opened 2026-08-30
fine-grained prefix-cache hits: exempt KpoolTailManager from the partial hash veto

simonbissonnette:
Upstream disables fine-grained prefix-cache hits whenever any cache manager lacks supports_fine_grained_hash_lookup and is not already hash-grained. On the GLM-5.3 hybrid the only such manager is KpoolTailManager, a 1-block/req circular scratch that never sees block hashes and already opts out of the hybrid hit min. Its veto forces block-aligned (3584-token) hits, so every follow-up re-prefills up to a full hybrid block of already-cached history.

The exemption ships ON (it is a fix, not a feature, consistent with the other hybrid patches in this repo). GLM53_FINE_GRAINED_APC=0 reverts to block-aligned-only hits. The MLA + mamba(align) managers both report supports_fine_grained_hash_lookup=True and reconcile the hit at hash_block granularity (64 tokens measured) instead of the 3584-token scheduler block.

start.sh wires the patch + env to BOTH ranks, the head container takes an explicit -e list, not serve_env (the worker-only env flow silently left the head vetoed). Adds a [glm53-fgapc] boot diagnostic. Idempotent, fail-closed on upstream layout drift.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -134,6 +134,12 @@ ABLIT_INCLUDE_MTP=1
 # After /health, burn DFlash2 BLOCK / sampler / kpool shapes so TP=2 does
 # not JIT mid-collective. 0 = skip (nonfatal either way).
 GLM53_BOOT_SHAPE_WARMUP=1
+# [fork] 1 = exempt the KpoolTail scratch manager from the partial-hash-hit
+# veto so MLA/mamba(align) prefix hits reconcile at hash granularity
+# (follow-ups re-prefill the new tail only, not a full 3584-token block).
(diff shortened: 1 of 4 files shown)


plotarmordev:
Closing as superseded: the same KpoolTailManager partial-hash veto exemption is carried forward by #125/#84, which are newer, measured, and default-off where this one is unmeasured and default-on. Thank you for finding this veto path first — the analysis here informed the newer work. Feel free to reopen if those PRs stall and this becomes the active line again.

PR #63, opened 2026-08-31
fix(template): emit Reasoning Effort unconditionally to keep the prefix cache

sovereignbrah:
## Problem

The reasoning-effort head line is gated on `thinking_enabled`:

```jinja
{%- if thinking_enabled and effective_reasoning_effort is not none -%}<|system|>Reasoning Effort: {{ effective_reasoning_effort | capitalize }}{%- endif -%}
```

It renders at roughly token 2, immediately after `[gMASK]<sop>` and **before** the tools block. So toggling thinking adds or removes ~7 tokens at the very head of the prompt.

vLLM chains prefix-cache block hashes forward from token 0 over token ids, so a divergence that early invalidates the entire prompt. On this kit the MLA `KpoolTailManager` only counts block-aligned hits, so there is no partial credit either — it is a total miss, not a degraded one.

Rendering the current template on a 4-message conversation with a system prompt, thinking on vs off:

| | shared prefix |
|---|---|
| without tools | 22 / 219 chars |
| with tools | 22 / 855 chars |

Agent traffic always carries a system prompt and tools, so in practice **every thinking toggle is a full re-prefill**.

## Change

Drop the `thinking_enabled` guard so the line is always emitted. Both shapes then share an identical head, and the thinking-off prompt becomes a *strict extension* of the thinking-on prompt, differing only by the trailing `</think>`. After the change both cases above share 100% of the on-prompt.

This also restores the behaviour of the upstream `zai-org/GLM-5.3-Flash` template, which emits this line unconditionally — the guard is a local addition here.

The control itself is unchanged: thinking is still disabled by the closed `<think></think>` block in the generation prompt. Only its cache cost changes. With thinking off the prompt still reads `Reasoning Effort: Max`, which is inert — the model cannot reason through an already-closed block.

## Measured

2× GB10, TP=2, EXL3 4bpw, DFlash2 k=7, 1M context, prefix caching on. 21,642-token prompt, `max_tokens=1`:

| step | effort | time |
|---|---|---|
| cold, first sight | high | **26.54 s** |
| warm, same effort | high | 0.62 s |
| **toggle high → none** | none | **0.42 s** |
| toggle none → high | high | 0.62 s |
| toggle high → low | low | 0.60 s |

Every toggle now hits warm. Before the change, row 3 cost a full re-prefill.

## Tests

Two existing tests in `tests/test_chat_template.py` asserted the line was *absent* when thinking was off, which encoded the divergence. They now assert the head stays stable while the tail carries the toggle.

Added `PrefixStabilityTests`: the off-shape must be a strict extension of the on-shape, with and without tools. The with-tools case is the one that actually regressed, since the tools block renders after the effort line.

Also registers a `tojson` filter accepting `ensure_ascii` in the test environment, so the template renders under bare Jinja2 the way vLLM's renderer supplies it. Without it these tests error out on the tools path.

```
$ python3 -m unittest tests.test_chat_template
Ran 7 tests — OK
```

## Known limitation, not addressed here

Changing effort *level* (`low` ↔ `high`) still rewrites the head, since the level word is in that line. Making that cache-stable too would mean moving the instruction to the last user turn, the way NVIDIA's Nemotron 3 templates do — a larger change with a real question attached about whether effort adherence survives the move, given the model was trained with it at the head. Deliberately out of scope for this PR.

### DIFF
--- a/files/chat_template.jinja
+++ b/files/chat_template.jinja
@@ -5,7 +5,7 @@
 {%- set thinking_enabled = true -%}
 {%- endif -%}
 {%- set effective_reasoning_effort = reasoning_effort if reasoning_effort is defined and reasoning_effort in ['low', 'high'] else 'max' -%}
-{%- if thinking_enabled and effective_reasoning_effort is not none -%}<|system|>Reasoning Effort: {{ effective_reasoning_effort | capitalize }}{%- endif -%}
+{%- if effective_reasoning_effort is not none -%}<|system|>Reasoning Effort: {{ effective_reasoning_effort | capitalize }}{%- endif -%}
 {%- set clear_thinking = clear_thinking if clear_thinking is defined else false -%}
 {%- if tools -%}
 {%- macro tool_to_json(tool) -%}
--- a/tests/test_chat_template.py
+++ b/tests/test_chat_template.py
@@ -1,25 +1,63 @@
 #!/usr/bin/env python3
 """Regression checks for GLM-5.3 chat-template reasoning controls."""
 
+import json
 import unittest
 from pathlib import Path
 
 from jinja2 import Environment
 
 
+def _tojson(value, ensure_ascii=False, indent=None, **_kwargs):
+    """vLLM's renderer supplies a tojson accepting ensure_ascii; bare Jinja2
+    does not. Register the same shape so the template renders standalone."""
+    return json.dumps(value, ensure_ascii=ensure_ascii, indent=indent)
+
+
+def _environment() -> Environment:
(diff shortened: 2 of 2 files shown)


plotarmordev (approved):
Exact-head validation passed: clean merge onto post-PR86 main, 7 focused template tests, deterministic byte-prefix checks, and real GLM-5.3 tokenizer verification with and without tools. Thinking-off is exactly thinking-on plus one </think> token; parser semantics are unchanged and the effort-line behavior restores upstream template parity. Non-blocking documentation follow-up: README's statement that thinking-off omits the reasoning-effort hint becomes stale and should be corrected.

PR #64, opened 2026-08-31
Codex/ChatGPT /v1/responses compatibility (v2: receive-delegate, versioned marker, fail-loud)

bpc-oss:
## Codex/ChatGPT `/v1/responses` compatibility for the TP2 recipe (v2, rebased on `6961fa0`)

(@plotarmordev — apologies for the placeholder body on the previous iteration; this is the real description with receipts. Force-pushed the branch to the rebased v2.)

### Problem

Codex/ChatGPT clients POST Codex-private item types to `/v1/responses` (`agent_message`, `custom_tool_call`, `compaction`, `item_reference`, reasoning items carrying `encrypted_content`). Stock pydantic validation rejects them, which surfaces as 400s that look like model failures.

A v1 overlay attempted to normalize them via an ASGI middleware, but its `_receive` returned `{"type": "http.disconnect"}` **unconditionally** after replaying the body — uvicorn aborted the exchange mid-stream and `/v1/responses` answered with a **null body**. Our first receipts only asserted HTTP 200, which is how that shipped. That design is abandoned.

### What v2 does

- `overlay/patch_codex_compat.py` (idempotent, fail-closed, MARK/anchor style):
  - `api_server.py`: `_CodexCompatMiddleware` — buffers the request body, normalizes Codex-private items **before pydantic**, replays the (possibly rewritten) body **once**, then **delegates to the real `receive()`** so uvicorn still observes `http.disconnect`.
  - `responses/utils.py` + `responses/protocol.py`: dict-form item fallbacks + widened input union.
  - Default `max_output_tokens=32768` injected for `/v1/responses` when absent.
  - **Versioned idempotency marker `__codex_compat_v2`**; a pre-v2 baked middleware now **fails loud** (`SystemExit`) instead of being silently skipped — the stale-marker revival path (null-body bug coming back via a baked old copy) is closed.
- `start.sh`: `CODEX_PATCH_HOST`, overlay-order entry, worker scp, two `-v` mounts (5 wiring lines).
- `tests/test_codex_compat.py`: marker + ast gate.

### Receipts (kit: 2x DGX Spark TP2, image `:exl3-instanttensor`, repo `6961fa0`)

In-image apply:
```
api_server: patched OK
resp_utils: encrypted_content stripped
resp_utils: fallback patched
resp_protocol: patched OK
CODEX_COMPAT_PATCH DONE
TEST_CODEX_COMPAT_PASS
```

Idempotency + stale-bake rejection (in-image):
- re-run -> `api_server: already patched` (skip, no double patch)
- marker stripped to simulate a stale baked copy -> `SystemExit: codex-compat: api_server carries a PRE-V2 _CodexCompatMiddleware … Refusing to silently skip` (fail-loud, verified)

Live boot on the kit: `CODEX_COMPAT_PATCH DONE` in head logs.

`/v1/responses` compatibility — **HTTP status AND body validated** (model `GLM-5.3-Flash-EXL3`):

| case | http | body |
|---|---|---|
| plain input | 200 | `resp_…` status=completed |
| `agent_message` item | 200 | completed |
| `custom_tool_call` (input object) | 200 | completed |
| `compaction` + `item_reference` | 200 | OK |
| reasoning `encrypted_content` + `include` | 200 | OK |

Non-Codex traffic untouched: the middleware matches only `POST /v1/responses` (line-verified); chat completions unaffected.

### Known limits

- `custom_tool_call.input` must be an object — a string input 400s from pydantic (same as stock behavior for malformed payloads).
- An independent read-only review of this patch also found the tail-delegate ordering depends on the loader applying weights before the hook (currently holds; a post-load hash assertion is tracked separately).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -449,7 +449,9 @@ COPY tests/test_hybrid_prefix_hit.py /opt/glm53/test_hybrid_prefix_hit.py
 COPY overlay/patch_xgrammar_termination.py /opt/glm53/patch_xgrammar_termination.py
 COPY tests/test_xgrammar_termination.py /opt/glm53/test_xgrammar_termination.py
 COPY overlay/patch_kpool_tail_slotmap.py /opt/glm53/patch_kpool_tail_slotmap.py
+COPY overlay/patch_codex_compat.py /opt/glm53/patch_codex_compat.py
 COPY tests/test_kpool_tail_slotmap.py /opt/glm53/test_kpool_tail_slotmap.py
+COPY tests/test_codex_compat.py /opt/glm53/test_codex_compat.py
 COPY overlay/ablit_runtime.py /opt/glm53/ablit_runtime.py
 COPY overlay/patch_ablit.py /opt/glm53/patch_ablit.py
 COPY tests/test_ablit.py /opt/glm53/test_ablit.py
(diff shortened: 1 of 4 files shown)


plotarmordev:
Closing as unreviewable as submitted: the PR body is literally "test body", with no measurements or rationale to evaluate. No judgment on the underlying idea — Codex/ChatGPT responses-API compatibility may well be worth doing. Please reopen with a real description and evidence if you pick this up.

PR #66, opened 2026-08-31
fix: keep VLLM_API_KEY out of argv and off the headless worker

hecisaza:
## What
Restores the two security properties merged PR #30 promised: the bearer token is **env-only** (never in argv) and **head-only** (the worker runs `--headless` and serves no API). Current `main` violates both.

## The regression (reproduced on `main` @ 688b7ab)
- `serve_env+=" -e VLLM_API_KEY='...'"" interpolates the secret into the **worker's** remote docker command — it lands in the local `ssh` argv and the remote `docker` argv (visible in `/proc/*/cmdline`, subject to the host's `hidepid`), on a node that never needs the credential.
- The head's `-e VLLM_API_KEY="$VLLM_API_KEY"` puts the value in the docker client argv.

Safe reproducer (dummy key, no real host) — the key is captured verbatim from `/proc/<pid>/cmdline` of the ssh process:
```
DUMMY_KEY='dummy-key-NOT-A-SECRET'
REMOTE_CMD="docker run -e VLLM_API_KEY='$DUMMY_KEY' example/image /start.sh"
timeout 3s ssh -T -o BatchMode=yes -o 'ProxyCommand=sleep 10' dummy.invalid "$REMOTE_CMD" & sleep 0.2
tr '\0' ' ' < /proc/$!/cmdline   # -> contains the dummy key
```

## Fix
- Worker: drop the credential from the env loop entirely.
- Head: `VLLM_API_KEY="$VLLM_API_KEY" docker run ... -e VLLM_API_KEY ...` — docker reads the value from its own environment; argv carries only the variable name. vLLM's native env-var behavior is unchanged.

## Test
`tests/test_api_key_argv.py` locks both properties on `launch_cluster` (fails on unfixed main). Full suite: 26 passed.

Precisely-scoped claim: this removes the secret from ordinary process argv and the unnecessary worker copy; same-owner/root can still read a process's environment, as with any env-var secret.

### DIFF
--- a/start.sh
+++ b/start.sh
@@ -1100,12 +1100,8 @@ launch_cluster() {
              ABLIT ABLIT_METHOD ABLIT_DIRECTION ABLIT_LAYERS ABLIT_ALPHA ABLIT_INCLUDE_MTP; do
         serve_env+=" -e $v='${!v:-}'"
     done
-    # VLLM_API_KEY is read by the head (rank 0) API server for bearer auth; the
-    # worker runs --headless so it only needs the var for argv-parity, and
-    # start.sh below passes it explicitly on the head. Keep it out of the
-    # generic loop so the key never shows in process listings of either node
-    # beyond the container env (same as the DeepSeek deployment).
-    serve_env+=" -e VLLM_API_KEY='${VLLM_API_KEY:-}'"
+    # The worker is headless and serves no API, so do not propagate the API
+    # credential into its remote docker command or container environment.
 
     log "starting worker on ${WORKER_SSH} (NCCL if=${WORKER_CX7_IF} hca=${WORKER_CX7_IB}) ..."
     worker_ssh "docker run -d --name '$CONTAINER_WORKER' \
@@ -1139,7 +1135,7 @@ launch_cluster() {
         --entrypoint bash '$IMAGE' /start.sh" >/dev/null
 
     log "starting head (vLLM API :${PORT}; NCCL if=${HEAD_CX7_IF} hca=${HEAD_CX7_IB}) ..."
-    docker run -d --name "$CONTAINER_HEAD" \
+    VLLM_API_KEY="$VLLM_API_KEY" docker run -d --name "$CONTAINER_HEAD" \
         --gpus all --network host --ipc=host --shm-size 32g --stop-timeout 60 \
         --device /dev/infiniband --cap-add IPC_LOCK \
         --ulimit memlock=-1 --ulimit stack=67108864 \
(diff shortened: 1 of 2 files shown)


plotarmordev:
**Status:** merge-ready after one test swap. Production hunks accepted as-is.

**Verified:** merged your head onto current main locally (clean, no conflicts) and reviewed every hunk. The production change is exactly right: the worker's remote docker command no longer carries `VLLM_API_KEY` at all, and the head passes a value-less `-e VLLM_API_KEY` with the credential assigned into the docker client process environment, so the value never appears in argv on either rank. Empty auth stays empty. Ran the affected launcher checks on the merged tree — `test_api_key_argv.py`, `test_start_overrides.py`, `test_bringup_robustness.py`, `test_numeric_config.py`, `test_warm_restart_stdout.py`, `bash -n start.sh` — all pass.

**Change requests:**

1. Replace the source-string test with a behavioral stubbed-launch test. The submitted assertions in `tests/test_api_key_argv.py` only match strings against the extracted `launch_cluster` source (`not in` / `in`); they never execute argument construction, so a refactor that keeps the strings but breaks the behavior (or vice versa) passes/fails wrongly. We have this as a local patch (our local rework `f650776`) and can push it onto your branch or you can apply it. It is ~80 lines, so summarizing its assertions rather than inlining:
   - Extracts `launch_cluster` from `start.sh` and actually executes it with a recorder `docker` shim first on `PATH` and stubbed `worker_ssh`/`scp`; no real docker/ssh is invoked.
   - Runs twice: once with a configured bearer containing spaces and `$`-literal characters, once with empty auth.
   - Asserts exactly one head `docker run`; the recorder's **environment** carries the configured value (proving the value-less `-e VLLM_API_KEY` inherits correctly); argv contains `-e VLLM_API_KEY` with no value attached.
   - Asserts the configured value appears in no captured argv on either rank, in no worker command, and in neither stdout nor stderr; `VLLM_API_KEY` is absent from the worker command entirely; empty auth stays empty on the head and reaches the worker not at all.
   - Shortens the historical docstring to one line.

Keep every production hunk unchanged — no trim warranted there; the new worker-side comment explaining why the credential is omitted is worth retaining.

**Waiting on:** nothing live from us — this was verified host-side with command stubs. Merge-ready as soon as the test swap lands.

plotarmordev:
Merged via https://github.com/MiaAI-Lab/GLM-5.3-Flash-EXL3-2x-DGX-Sparks/pull/154 (rebased onto current main with maintainer fixes; your commits and authorship are preserved). Thank you for the contribution — closing this one in favour of the merged replacement.

PR #67, opened 2026-08-31
fix: pin the DFlash2 drafter revision — mutable Hub main silently drifts weights

hecisaza:
## What
Adds `DFLASH_REVISION` (documented in `.env.example`) so fresh and warm installs resolve the **same drafter bytes**. Today the launcher runs `hf download $DFLASH_MODEL` with no `--revision`.

## Why it matters (verified against the Hub API)
`incoai/GLM-5.3-Flash-DFlash2` has shipped **three different `model.safetensors`** under the same model ID and filename:

| commit | date (UTC) | LFS sha256 (prefix) |
|---|---|---|
| `7d74cdd` | 2026-08-27 | `8931dc52…` |
| `dc77ff1` | 2026-08-28 | `b33c0347…` |
| `bf582e4` | 2026-08-31 | `b038e1d9…` |

A warm cache keeps serving whatever it downloaded; a fresh clone gets today's `main`. Two deployments of the same recipe run different drafters while the README's performance receipts describe only one of them.

## Fix
- `DFLASH_REVISION` env knob, default **`dc77ff1`** — the checkpoint the deployment behind the 2026-08-30 `DFLASH_DRAFT_TP=2` receipts is pinned to (verified via `refs/main` in that deployment's HF cache). If you'd rather adopt `bf582e4`, benchmark it and change pin + receipts together — the defect is the silent drift, not the specific checkpoint.
- `--revision` passed to `hf download`; resolution prefers the pinned snapshot even if `refs/main` points elsewhere; `model.safetensors` required in the selected snapshot; worker sync marker keyed on the pin so a ref flip can't skip the rsync.
- Operators can override or empty `DFLASH_REVISION` to track `main` deliberately.

## Test
`tests/test_dflash_revision_pin.py` drives `download_dflash`/`resolve_dflash_dir` against a mock `hf` CLI and asserts the exact `--revision` argv and pinned resolution path (fails on unfixed main). Full suite: 27 passed.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -83,6 +83,8 @@ SERVED_MODEL_NAME=GLM-5.3-Flash-EXL3
 # Rollback: SPEC_METHOD=mtp
 SPEC_METHOD=dflash
 DFLASH_MODEL=incoai/GLM-5.3-Flash-DFlash2
+# Exact checkpoint behind the published 2026-08-30 TP=2 receipts.
+DFLASH_REVISION=dc77ff1c99eeb2df044ee3d4f0094eb033fee410
 DFLASH_TOKENS=7
 # Drafter TP. 2 = measured keep on this 2× GB10 kit (shard draft across ranks).
 # 1 = rank 0 only (skip CX7 on a 2.3 GiB draft). Empty = inherit TP.
--- a/start.sh
+++ b/start.sh
@@ -153,6 +153,10 @@ MTP_TOKENS="${MTP_TOKENS:-2}"
 SPEC_METHOD="${SPEC_METHOD:-dflash}"
 DFLASH_MODEL="${DFLASH_MODEL:-incoai/GLM-5.3-Flash-DFlash2}"
 DFLASH_CACHE_NAME="${DFLASH_CACHE_NAME:-models--${DFLASH_MODEL//\//--}}"
+# Receipt-matched DFlash2 checkpoint used by the 2026-08-30 TP=2 results.
+# A mutable Hub main has already changed weights, so fresh and warm installs
+# must resolve the same snapshot unless the operator deliberately overrides it.
+DFLASH_REVISION="${DFLASH_REVISION:-dc77ff1c99eeb2df044ee3d4f0094eb033fee410}"
 DFLASH_TOKENS="${DFLASH_TOKENS:-7}"
 # 2 = shard the ~2.3 GiB DFlash2 drafter across TP (C4 keep, 2026-08-30:
 # idle 8k 938 / 16k 972 / 100k 997; decode structured 65.1 / prose 27.1).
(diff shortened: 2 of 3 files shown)


plotarmordev:
**Status:** rework, small. The pin itself is accepted — the default `dc77ff1c99eeb2df044ee3d4f0094eb033fee410` is byte-identical to the revision our 2x GB10 TP=2 E3 kit currently serves, so adopting it is zero-drift for us.

**Verified:** merged your head onto current main locally (clean) and reviewed the full diff. Confirmed the main-side defect: `download_dflash` calls `hf download` with no `--revision`, and a missing draft `refs/main` is synthesized from the latest snapshot directory rather than a pinned commit — a fresh sync can change decode behavior overnight. Your mock-HF download test and the affected launcher suite (`test_start_overrides.py`, `test_bringup_robustness.py`, `test_numeric_config.py`, `test_warm_restart_stdout.py`, `bash -n start.sh`) all pass on the merged tree. We also exercised the real resolver/sync functions against fixture caches.

**Change requests** (line numbers refer to your PR merged onto current main):

1. **Fail closed when the pinned dir is absent — `start.sh:463`.** `ensure_dflash_refs_main` only honors a nonempty `DFLASH_REVISION` if `snapshots/$DFLASH_REVISION` already exists; otherwise it silently falls back to `refs/main`. A pinned launch with downloads skipped can then silently select different weights — the exact drift this PR exists to prevent. Remove only the directory-existence conjunct; the existing selected-snapshot weights-file checks then fail closed. Our strengthened resolver test caught this: with the pinned directory missing, the submitted resolver returned the stale snapshot and old sync marker instead of failing.
2. **`DFLASH_REVISION` must join the caller-export capture/restore (`start.sh:60-122`).** Today an explicit caller value loses to `.env`. Add a setness-aware capture next to the speculative-method capture and restore after `.env`, following the existing convention in that block.
3. **`start.sh:185`: use `${DFLASH_REVISION-...}`, not `${DFLASH_REVISION:-...}`.** The PR body promises explicit empty means unpinned; `:-` replaces explicit empty with the default pin. Combined with (2), explicit empty then survives either caller configuration or `.env`.
4. **Remove the dead completion check (`start.sh:917-919`).** `selected="$(resolve_dflash_dir)"` discards the resolved path and `have=1` makes the following failure check unreachable. Replace those three statements with `resolve_dflash_dir >/dev/null`.
5. **Replace the static test.** `test_pin_is_declared_and_threaded_through` only asserts source strings. Keep the submitted mock-HF download test; swap the static one for behavioral tests.
6. **Docs:** add an `.env.example` comment stating overrides are exact commits and empty follows cached main (refresh required to update it).

We have all of this as a local patch (our local rework `37211db`) and can push it onto your branch or you can apply it. It is ~170 lines across three files, so describing the hunks precisely:

- `start.sh` (7 insertions, 5 deletions): setness-aware capture `_cli_dflash_revision_set`/`_cli_dflash_revision` before `.env` is sourced and restore after it; `:-` → `-` at the default assignment; drop the `[ -d ... ]` conjunct in `ensure_dflash_refs_main`; replace the dead check in `download_dflash` with `resolve_dflash_dir >/dev/null`.
- `.env.example` (+1): the override-semantics comment from (6).
- `tests/test_dflash_revision_pin.py`: two new behavioral tests replacing the static one — (a) an override/empty precedence matrix (4 cases: unset→default pin; caller override beats `.env`; caller explicit empty beats `.env`; `.env` explicit empty survives) run through the real start.sh preamble against a temp `.env`; (b) a resolver/sync-marker test against a fixture repo with a stale `refs/main`: the pinned snapshot resolves, the sync marker keys on the pin, and removing the pinned weights file or the pinned directory exits with the failure status.

Both new tests **failed against the submitted PR** before the corresponding fixes (stale-snapshot fallback; caller override replaced by the default pin) and pass after — so they pin the actual defects, not the wiring.

**Waiting on:** nothing live from us. We also merged this together with #66 and #68 locally — no conflicts, combined suite passes — so the three can land in any order. Once the above lands, this is merge-ready.

plotarmordev:
Merged via https://github.com/MiaAI-Lab/GLM-5.3-Flash-EXL3-2x-DGX-Sparks/pull/155 (rebased onto current main with maintainer fixes; your commits and authorship are preserved). Thank you for the contribution — closing this one in favour of the merged replacement.

Issue #52, opened 2026-08-30
start.sh weight check trusts hub-cache refs/main and fails on a fully-staged pinned snapshot


hughmadden:
`model_dir()` in `start.sh` resolves the model snapshot via the hub-cache `refs/main`. An `hf download --revision <pin>` can leave `refs/main` pointing at a newer upstream commit whose snapshot was never materialized locally — after which a fully-staged pinned snapshot fails the boot with:

```
ERROR: config.json missing in .../snapshots/<other-sha> — re-run with REFRESH_WEIGHTS=1
```

…which would re-download all 164 GiB. We hit this on a fresh 2× Spark deploy where the Hub repo's `main` moved between weight staging and first launch (staged pin `25a44fd…`, `refs/main` pointing at `024db9f7…` with no snapshot directory).

Suggested fix: when `MODEL_REVISION` is set and `snapshots/$MODEL_REVISION/config.json` exists, prefer it over `refs/main` (or rewrite `refs/main` to the pin). Our workaround was exactly that — `printf '<pin>' > refs/main` on the head, plus the same on the worker's rsynced cache (the stale ref gets copied too). Happy to PR the small change if useful.

Unrelated small note from the same deploy, in case it helps others: iterating on locally-derived images is much faster with `SKIP_SHIP=1` plus building the same Dockerfile on the worker from shared storage, since `docker save | ssh load` re-ships every layer even when only a small derived layer changed.

Thanks for the recipe — it's been rock solid otherwise; we're running it as a daily driver with DFlash2 and (via some vLLM-side patches we're upstreaming) a node-local disk KV tier on top.

PR #68, opened 2026-08-31
fix: scope weight completeness to the active snapshot (cross-snapshot false-pass)

hecisaza:
## What
`count_shards()` counts every `*.safetensors` recursively under `snapshots/`, but `resolve_model_dir()` later resolves exactly **one** snapshot via `refs/main`. The completeness check can therefore be satisfied by files the launcher will never load.

## Reproduced false-pass (fixture, from `main` @ 688b7ab)
- `refs/main -> active`; `snapshots/active/` holds `config.json` + **1** shard; `snapshots/old/` holds **119** stale shards
- `count_shards` returns **120** → "weights already present (120 shards)" → download/sync skipped → resolver selects the incomplete `active` → vLLM fails minutes later mid-load instead of seconds early with an actionable message

`download_dflash` has the same shape (any `model.safetensors` anywhere under `snapshots/` satisfies the guard) — fixed separately in the DFlash revision-pin PR, which requires the weight in the selected snapshot.

## Fix
Resolve the ref first (`refs/main`, falling back to the newest snapshot, matching the launcher's existing style), then count only regular files in that one snapshot directory.

## Relationship to #52
Distinct but adjacent: #52 is `refs/main` pointing **away** from a fully-staged pinned snapshot; this is the count being satisfied **across** snapshots even when the selected ref has `config.json`. Happy to align the resolver with whatever shape #52 settles on. A stronger follow-up could parse `model.safetensors.index.json` and require every `weight_map` filename to exist instead of trusting any count.

## Test
`tests/test_snapshot_scoped_count.py`: the 1-active/119-stale fixture must count **1** (mutation-checked — unfixed main returns 120), plus an empty-repo → 0 case. Full suite: 27 passed.

### DIFF
--- a/start.sh
+++ b/start.sh
@@ -317,7 +317,15 @@ worker_ssh() { ssh -T -o BatchMode=yes -o ConnectTimeout=15 "$WORKER_SSH" "$@"; 
 usage() { sed -n '2,40p' "${BASH_SOURCE[0]}" | sed 's/^# \{0,1\}//'; }
 
 count_shards() {
-    find "$1/snapshots" -name '*.safetensors' 2>/dev/null | wc -l | tr -d '[:space:]' || true
+    local repo_path="$1" ref
+    ref="$(cat "$repo_path/refs/main" 2>/dev/null || true)"
+    [ -n "$ref" ] || ref="$(ls -1t "$repo_path/snapshots" 2>/dev/null | head -n 1 || true)"
+    if [ -z "$ref" ]; then
+        printf '0'
+        return
+    fi
+    find "$repo_path/snapshots/$ref" -maxdepth 1 -type f -name '*.safetensors' 2>/dev/null \
+        | wc -l | tr -d '[:space:]' || true
 }
 
 ensure_refs_main() {
--- a/tests/test_snapshot_scoped_count.py
+++ b/tests/test_snapshot_scoped_count.py
@@ -1,0 +1,71 @@
+#!/usr/bin/env python3
+"""Regression guard: weight completeness is scoped to the active snapshot.
+
+A repo-wide count can accept 119 stale shards in snapshots/old plus one shard
+in the active snapshot as "120 present", then resolve the incomplete active
+snapshot and fail deep in vLLM load. Related: issue #52 (stale refs/main).
+"""
+
+from __future__ import annotations
+
+import re
+import subprocess
+import tempfile
+from pathlib import Path
+
+ROOT = Path(__file__).resolve().parents[1]
+START = ROOT / "start.sh"
+
+
+def function(name: str) -> str:
+    text = START.read_text(encoding="utf-8")
+    match = re.search(rf"(?ms)^{re.escape(name)}\(\) \{{\n.*?^\}}\n", text)
+    assert match, f"missing function {name}"
+    return match.group(0)
+
+
+def run_bash(script: str) -> subprocess.CompletedProcess[str]:
+    return subprocess.run(
+        ["bash", "-c", script], capture_output=True, text=True, check=False
+    )
+
+
+def test_target_completeness_is_scoped_to_active_snapshot() -> None:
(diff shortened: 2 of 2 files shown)


plotarmordev:
**Status:** merge-ready after a one-liner. The scoping fix itself is accepted.

**Verified:** merged your head onto current main locally (clean). Reproduced the main-side defect with a fixture (incomplete active snapshot + 119 stale shards in sibling snapshots): main's `count_shards` counts 120 and the completeness check false-passes; your snapshot-scoped version counts 1. Your two behavioral tests pass on the merged tree, along with the affected launcher suite.

**Change requests:**

1. **`start.sh:432`: `find` needs `-L`.** The submitted `find "$repo_path/snapshots/$ref" -maxdepth 1 -type f -name '*.safetensors'` does not follow symlinks, so on a real Hugging Face cache — where snapshot entries are symlinks into `blobs/` — a *complete* snapshot counts **0** shards and the completeness check false-fails. Your tests only create regular files, so they miss it. Our linked-cache regression failed on your submitted PR exactly this way (`assert result.stdout.strip() == "1"` got `"0"`).
2. Add the symlink fixture test: a valid blob link counts, a dangling link does not.

We have this as a local patch (our local rework `9d2ece2`) and can push it onto your branch or you can apply it — full diff below:

```diff
diff --git a/start.sh b/start.sh
--- a/start.sh
+++ b/start.sh
@@ -429,7 +429,7 @@ count_shards() {
         printf '0'
         return
     fi
-    find "$repo_path/snapshots/$ref" -maxdepth 1 -type f -name '*.safetensors' 2>/dev/null \
+    find -L "$repo_path/snapshots/$ref" -maxdepth 1 -type f -name '*.safetensors' 2>/dev/null \
         | wc -l | tr -d '[:space:]' || true
 }
 
diff --git a/tests/test_snapshot_scoped_count.py b/tests/test_snapshot_scoped_count.py
--- a/tests/test_snapshot_scoped_count.py
+++ b/tests/test_snapshot_scoped_count.py
@@ -65,7 +65,28 @@ def test_empty_repo_counts_zero() -> None:
         assert result.stdout.strip() == "0", result.stdout
 
 
+def test_cached_blob_links_count_but_dangling_links_do_not() -> None:
+    with tempfile.TemporaryDirectory() as tmp:
+        repo = Path(tmp)
+        active = repo / "snapshots" / "active"
+        active.mkdir(parents=True)
+        (repo / "refs").mkdir()
+        (repo / "refs" / "main").write_text("active", encoding="utf-8")
+        (repo / "blobs").mkdir()
+        (repo / "blobs" / "present").touch()
+        (active / "present.safetensors").symlink_to("../../blobs/present")
+        (active / "missing.safetensors").symlink_to("../../blobs/missing")
+        result = run_bash(
+            "set -euo pipefail\n"
+            + function("count_shards")
+            + f"count_shards {str(repo)!r}\n"
+        )
+        assert result.returncode == 0, result.stderr
+        assert result.stdout.strip() == "1", result.stdout
+
+
 if __name__ == "__main__":
     test_target_completeness_is_scoped_to_active_snapshot()
     test_empty_repo_counts_zero()
+    test_cached_blob_links_count_but_dangling_links_do_not()
     print("snapshot-scoped count guard OK")
```

Keep snapshot selection, depth limiting, and the zero result for an absent repository exactly as submitted; retain both of your behavioral tests unchanged.

**Waiting on:** nothing live — host-verified. Merge-ready once the one-liner and the fixture test land. (Also merged cleanly alongside #66 and #67 in our local combined branch; the three can land in any order.)

plotarmordev:
Merged via https://github.com/MiaAI-Lab/GLM-5.3-Flash-EXL3-2x-DGX-Sparks/pull/153 (rebased onto current main with a maintainer fix; your commits and authorship are preserved). Thank you for the contribution — closing this one in favour of the merged replacement.

PR #69, opened 2026-08-31
GB10: opt-in SpinCondition busy_loop 1 s -> 2 ms (GLM53_SPINWAIT_2MS)

rexffan:
## What

Opt-in patch shrinking vLLM's `SpinCondition.busy_loop_s` from **1 s to 2 ms** inside the serving container, gated by `GLM53_SPINWAIT_2MS=1`. **Default OFF — merging changes nothing** until an operator opts in.

## Why

The bundled vLLM ships `shm_broadcast.py` with `busy_loop_s: float = 1`: after every read, each reader thread `sched_yield()`-spins for a full second before parking on the zmq notify poller. On a 20-core GB10 those spinners compete with NCCL proxy threads and the engine loop for cores.

Same-silicon precedent: the DeepSeek-V4-Flash DSpark image was patched 1 s → 2 ms in Aug 2026 and validated on four Sparks ([analysis](https://nacyot.github.io/artifacts/vllm-spin-wait-gb10/)). The newer `SpinCondition` in this tree keeps the zmq wake-up path, so the shrink is safe: readers just park on the poller 2 ms after the last read and the writer's ping wakes them.

## Measured

On a 4x DGX Spark TP=4 deployment of this kit's lineage (CRS804 switch fabric), spinwait-only A/B window: no regression in any decode tier; code 4-stream aggregate +8%; 70k-token cold-prefill TTFT −19%. These are TP=4 numbers — treat as directional for TP=2, which is why the default is off.

## How

Wired exactly like `patch_kpool_tail_slotmap`: overlay file, `*_PATCH_HOST` var, worker scp, both container mounts, both patch-exec blocks, one entry in the shared env array, `.env.example` documentation. The patch is idempotent and fails closed on anchor drift (a vLLM bump can't silently half-apply). `bash -n start.sh` passes.

## Verify on your pair

```bash
GLM53_SPINWAIT_2MS=1 ./start.sh
# boot log should show: "spinwait: busy_loop_s 1 -> 0.002 patched" on both ranks
docker exec <head> grep -n "busy_loop_s: float" \
  /usr/local/lib/python3.12/dist-packages/vllm/distributed/device_communicators/shm_broadcast.py
```

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -155,3 +155,8 @@ NCCL_IB_GID_INDEX=3
 # still enabled. 1 = upstream default.
 CG_ESTIMATE=1
 NCCL_DEBUG=WARN
+
+# GB10 core relief: shrink vLLM SpinCondition reader spin from 1 s to 2 ms.
+# Readers park on the zmq notify poller sooner instead of sched_yield-spinning,
+# freeing cores for NCCL proxy threads. Validated on 4x TP=4; A/B on your pair.
+# GLM53_SPINWAIT_2MS=1
--- a/overlay/patch_spinwait_gb10.py
+++ b/overlay/patch_spinwait_gb10.py
@@ -1,0 +1,52 @@
+#!/usr/bin/env python3
+"""GB10: shrink SpinCondition busy_loop_s 1 s -> 2 ms (opt-in via GLM53_SPINWAIT_2MS=1).
+
+The vLLM build in this image ships shm_broadcast.py with SpinCondition
+busy_loop_s defaulting to 1 second: after every read, each reader thread
+sched_yield()-spins for a full second before falling back to the zmq notify
+socket.  On GB10 (20-core Grace, unified memory) those spinning readers
+compete with NCCL proxy threads and the engine loop for cores.
+
+Precedent on the same silicon: the DeepSeek-V4-Flash DSpark image was patched
+from 1 s to 2 ms in 2026-08 and validated on four Sparks (analysis:
+https://nacyot.github.io/artifacts/vllm-spin-wait-gb10/).  The newer
+SpinCondition here keeps a zmq wake-up path, so shrinking the spin window is
+safe: readers just park on the poller 2 ms after the last read instead of
+1 s, and the writer's notify ping wakes them.
+
+Measured on a 4x DGX Spark TP=4 deployment of this kit (CRS804 switch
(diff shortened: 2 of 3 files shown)


zltnxyz:
**TP=2 A/B results from a 2×DGX Spark EXL3 deployment (ABLIT=1, DFlash2 k=7, recipe @ 493cb88)** — can confirm this patch on the same silicon.

**Protocol:** identical C=4 × ≤400-token completion probe before/after activation, hottest acpitz zone sampled every 3 s on the worker node (asus). Same `.env` otherwise; the only change between runs was `busy_loop_s 1 → 0.002` (applied via the same mechanism this PR proposes).

| Metric | Pre (stock, busy_loop_s=1) | Post (2 ms) | Δ |
|---|---|---|---|
| Worker peak temp under load | **91 °C** | **77 °C** | **−14 °C** |
| Aggregate throughput (C=4) | 42.2 tok/s | 44.3 tok/s | ≈ equal / slightly up |
| Worker load1 during decode | 1.4–1.8 | 0.7–1.1 | visibly lower |
| Post-load cooldown | 91 → 64 °C in ~90 s | 77 → 54 °C in ~90 s | faster recovery |

Notes:
- This replicates the DSpark-kit precedent exactly (DeepSeek-V4-Flash on 2×GB10, Aug 2026: worker 77.6 → 63.2 °C, zero throughput change; analysis linked in the PR description).
- Pre-patch worker peaks of 85–91 °C are uncomfortably close to GB10 thermal-shutdown territory (~87 °C per the linked article); 2 ms removes that headroom concern for us.
- Staging note for operators testing this: since `start.sh` recreates containers from the pinned image, a `docker exec sed` into a running container does **not** survive a recreate — only survives plain `docker restart`. We stage it as an overlay-style patch (`overlay/patch_vllm_spin_wait.py` + start.sh wiring: host var, head/worker mounts, worker scp, exec guards) so it re-applies at every boot; if this PR merges, that collapses into `GLM53_SPINWAIT_2MS=1`.
- No regressions observed: engine boots clean, 20/20 warmup, health/`/v1/models` normal, tool-calling unaffected.

Happy to provide the raw 3 s-interval thermal/load traces (both runs) if useful.

plotarmordev:
Superseded by #96, which replaces the fixed 2 ms boolean with the validated numeric GLM53_SPINWAIT_MS contract and includes exact-head tests/build wiring.

PR #70, opened 2026-08-31
docs: GB10 UVM livelock runbook + spec-acceptance gate script

rexffan:
## What

Docs + one read-only script, no serving-path changes:

- `docs/uvm-livelock-gb10.md` — field runbook for the GB10 UVM livelock class (prevention / detection / recovery), from two incidents on a 4x TP=4 deployment of this kit's lineage.
- `scripts/spec-accept-gate.sh` — post-boot check for the vllm#53030 signature (CUDA graphs silently pinning per-position spec acceptance at exactly 1.00). Reads `/metrics`, prints the decay curve, fails on the pinned shape.

## Why this might save someone's night

This failure class is the one tonyd2wild's OPEN-PROBLEMS #4 describes as "it never recovers". Our contribution is the part that was unknown there — **both of our incidents recovered without a power cycle**, and the triage that makes that possible:

- `/health` stays **200 the whole time** the engine is wedged (V1 `check_health` never probes workers) — the doc gives a `/metrics`-based stall detector that actually fires.
- After `docker kill`, `nvidia-smi` can keep reading ~96% util at idle wattage with nothing running. That reading is cosmetic: a 10-second CUDA matmul probe distinguishes it from a truly wedged driver, and in our two incidents the probe said healthy → plain restart, 18 minutes end to end, no machine-room trip.
- Prevention detail that cost us a mid-serve wedge: `vm.swappiness=0` stops new page-outs but does nothing about pages **already** in swap — a ~236k-token prefill touching week-old swapped pages triggered the spiral. Cycle residual swap (`swapoff -a && swapon -a`) while containers are down, and keep the swap file itself (no swap at all gets the worker killed on allocation spikes).

All numbers and symptoms are from DGX Spark GB10 hardware running this kit's image lineage.

### DIFF
--- a/docs/uvm-livelock-gb10.md
+++ b/docs/uvm-livelock-gb10.md
@@ -1,0 +1,78 @@
+# GB10 UVM livelock: prevention, detection, recovery
+
+Field notes from running this kit's lineage on DGX Spark (GB10, unified
+memory). Two incidents on a 4x TP=4 deployment (2026-08-26 unclean node
+reboots; 2026-08-31 mid-serve wedge) plus the recovery that worked. The
+failure class is the one tonyd2wild's OPEN-PROBLEMS #4 describes — this doc
+adds the part that was unknown there: it is survivable without a power cycle.
+
+## The failure
+
+If the kernel pages vLLM memory out (or back in) under memory pressure, the
+UVM driver can enter an unrecoverable spin:
+
+- every engine step stalls; `shm_broadcast.py` logs
+  `No available shared memory broadcast block found in 60 seconds` every 60 s
+- `nvidia-smi` shows **~96% GPU utilization at idle wattage (~12–20 W)** —
+  busy-looking, doing nothing
+- the stuck request stays in `num_requests_running` forever;
+  `VLLM_EXECUTE_MODEL_TIMEOUT_SECONDS` does not fire
+- worst case the node drops off SSH (sshd cannot fork) and needs a power cycle
+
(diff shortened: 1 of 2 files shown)


plotarmordev:
### Review follow-up — `72a0388571fd`

I ran the script with synthetic /metrics responses, without contacting a serving stack: pinned position-0 acceptance exited 1 with FAIL, a non-pinned case exited 0 with PASS, and fewer than 100 drafts exited 0 with SKIP. With 150 drafts but no per-position series, it exited 1 after only printing drafts_total, with no explanation.

Please make missing/mismatched metric series an explicit diagnostic error, and handle multiple label sets deliberately rather than feeding multiple values into a scalar denominator. Also replace the site-specific default URL with localhost or a required argument. The runbook remains useful; these are small script corrections, not a request for hardware evidence.

Next action: author corrections, then our same scoped fixture check before merge.

plotarmordev (approved):
SIX exact-head approval: 3bb97a4e7cebb7f54e741995970553c41ad5384f. Independently reviewed the parser and frozen runbook; corrected missing/ambiguous evidence, scientific/timestamp parsing, overflow refusal, and actual position-label boundaries. The inert shell gate suite passed at this exact head after normal integration of current public main; previously observed overflow/phantom-label cases now exit 2 without PASS/FAIL/SKIP. Complete publishing tree and pushed commit range were scanned for scope, secrets and deployment-local content; author history and source branch are preserved. Repository has no workflows/checks: no CI-green claim. No live endpoint, GPU, SSH, container or service operation; incident runbook observations remain author-reported.

PR #72, opened 2026-08-31
preflight: support dual-rail NCCL_IB_HCA comma lists in the per-rank GID check

ajbmachon:
## Problem

Since #26 the preflight builds sysfs paths from `HEAD_CX7_IB` / `WORKER_CX7_IB` verbatim:

```bash
gid_path="/sys/class/infiniband/${HEAD_CX7_IB}/ports/1/gids/${HEAD_GID}"
```

`NCCL_IB_HCA` accepts a comma list, and the launcher already passes the value straight through to NCCL, so a dual-rail config works end to end — but the sysfs path built from the raw list does not exist, `gid_head` comes back empty, and preflight `die`s on a healthy deployment.

**Repro** (any pair with both QSFP ports cabled):

```
HEAD_CX7_IB=rocep1s0f0,roceP2p1s0f0
WORKER_CX7_IB=rocep1s0f0,roceP2p1s0f0
./start.sh restart
# -> "head GID index 3 is EMPTY on rocep1s0f0,roceP2p1s0f0" + die, service left stopped at next start
```

## Fix

- Split each rank's `*_CX7_IB` on commas and require that rank's GID index to be populated on **each** listed device — NCCL applies one `NCCL_IB_GID_INDEX` to every HCA it opens, so all of them must carry a usable entry.
- The refusal dump now prints one GID table per device (with `gid_attrs` types, as before).
- Single-device configs are byte-identical in behavior (splitting one name is a no-op), and the per-rank `HEAD_GID` / `WORKER_GID` semantics from #26 are preserved.

## Why dual rail is worth keeping launchable (measured)

2x DGX Spark (GB10), direct QSFP cable, both rails 200 Gb/s. Same suite as the kit receipts (3 spec-plan/code tasks with 2–8k-token prompts + 20 HumanEval), the only change between arms is the second rail in `NCCL_IB_HCA`:

| metric | single rail | dual rail |
|---|---|---|
| TTFT, 2.2k-token prompt | 3.73 s | 3.55 s |
| TTFT, ~8k-token prompt | 11.21 s | 10.55 s |
| TTFT, code task | 2.97 s | 2.71 s |
| median decode tok/s | 49.6 | 49.4 |
| HumanEval pass@1 (20) | 20/20 | 20/20 |

First token is ~5–9 % faster on real prompts; decode is unchanged (the per-token allreduce is latency-bound, not bandwidth-bound). Port counters confirmed both rails carrying traffic during the run (~62 MB/s each).

## Verification

- `bash -n start.sh`
- `python3 tests/test_numeric_config.py` → PASS (guard block untouched)
- Positive path, live on the dual-rail pair: all 4 endpoints (2 devices × 2 nodes) validated at index 3, preflight passes.
- Refusal path, live with a deliberately empty index (7): both ranks detected empty, per-device GID tables printed for both rails.

https://claude.ai/code/session_01WDpCmmggXtL3Ffimh2u9f1

### DIFF
--- a/start.sh
+++ b/start.sh
@@ -394,11 +394,20 @@ preflight() {
     # rank ~60 s in with ibv_modify_qp errno 61 "No data available". The index is
     # per-NIC, so validate head and worker separately: some pairs share one good
     # index, others need different ones (HEAD_GID / WORKER_GID).
-    local gid_head gid_worker gid_path
-    gid_path="/sys/class/infiniband/${HEAD_CX7_IB}/ports/1/gids/${HEAD_GID}"
-    gid_head=$(cat "$gid_path" 2>/dev/null | tr -d ':0' || true)
-    gid_path="/sys/class/infiniband/${WORKER_CX7_IB}/ports/1/gids/${WORKER_GID}"
-    gid_worker=$(worker_ssh "cat '$gid_path' 2>/dev/null" | tr -d ':0' || true)
+    # CX7_IB may be a comma list (dual rail). NCCL applies one GID index to
+    # every HCA it opens, so each rank's index must be populated on EACH of
+    # its listed devices.
+    local gid_head=set gid_worker=set gid_path dev entry
+    for dev in ${HEAD_CX7_IB//,/ }; do
+        gid_path="/sys/class/infiniband/${dev}/ports/1/gids/${HEAD_GID}"
+        entry=$(cat "$gid_path" 2>/dev/null | tr -d ':0' || true)
+        [ -n "$entry" ] || gid_head=""
+    done
+    for dev in ${WORKER_CX7_IB//,/ }; do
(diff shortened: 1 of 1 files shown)


plotarmordev:
### Review follow-up — `b7149dd5cc09`

The dual-rail preflight fix is present on current main through merged #172, including the per-device GID diagnostics. Your contribution and dual-rail measurements are acknowledged.

This branch is now superseded, not waiting for an author rebase or another benchmark. The `duplicate` label refers to #172. No action is requested from you; deciding/recording closure in favor of #172 is ours.

plotarmordev:
Closing as superseded by merged #172 (0758c07): main now checks each selected HCA’s GID and reports per-device diagnostics. Thank you @ajbmachon for the original dual-rail fix; the historical receipts remain in this discussion. No author action is requested, and this closure claims no new tests or live qualification.

PR #83, opened 2026-08-31
Per-KV-cache-group prefix-cache retention: sparse drafter, dense MLA/mamba

nood-co1:
### Root cause (proven from code, reconciled with measurements)
Block ids are global across KV-cache groups in one LRU pool (642 usable on this kit). A cached 3584-token segment costs
**38 ids: 1 MLA + 4 mamba + 33 DFlash2 drafter** SWA blocks; the drafter blocks are hashed and freed *mid-prefill* to the LRU
tail — ahead of the MLA/mamba blocks that carry the hit (priority inversion; the hybrid `min()` discards the drafter's hit
length anyway). Knee = (642 − 33t)/(2·38 − 33) ≈ 14.9 segments/conversation (measured: 14 OK / 17 fail). The global
retention interval (PR #79) buys capacity but coarsens every group's hit grid (diverging subagent: 45 % at 14336, 0 % at 28672).

### Change
Overlay `patch_apc_per_group_retention.py` (coordinator only; fail-closed; composes with `patch_hybrid_prefix_hit.py` in either
order — tested): retention becomes per-group; the EAGLE min-exempt drafter group (derived from `eagle_group_ids`) gets
`VLLM_PREFIX_CACHE_RETENTION_INTERVAL_SWA` (auto/`0` = reachable boundaries only; validated ≤ 1e6, block-aligned, unconditional);
other groups keep the global value. Boot log prints the resolved vector. Launcher knob (both ranks, caller export wins incl.
explicit `0` over a blank `.env`), Dockerfile COPY/test/RUN, README/.env.example, 12-part host test.

### Live receipts (idle 2× GB10, `GLM53_APC_RETENTION_INTERVAL=` dense + `_SWA=0`, boot log `retention_by_group=[None,…,None,0]`)
| | dense (before) | global 14336 (PR #79) | **per-group (this PR)** |
|---|---|---|---|
| two 56K conversations | 0 % / 0 % (70 s re-reads) | 98 % | 98 % (1.9 s) |
| two 80K | — | 96 % | 96.4 % (4.0 s) |
| **two 120K** | — | untested | **98.5 % (2.7 s re-turns)** |
| diverging subagent (shares ~20K of a 55K root) | 3584 grid but capacity broken | 45 % (14336 grid) | **67.6 % (3584 grid)** |
| L5 decode rate after a warm diverging hit | — | — | 1.01× cold (pass ≥ 0.90) |
[...]

PR #84, opened 2026-08-31
Re-enable fine-grained (64-token) prefix-cache hits (GLM53_FINEGRAINED_APC)

nood-co1:
### Problem
The hybrid coordinator disables fine-grained hits for the whole model because `KpoolTailManager` — a per-request scratch that
opts out of prefix caching entirely — is included in the compatibility check. One veto silences three cooperating subsystems
that already operate at `hash_block_size` = 64 (mamba partial-tail snapshot producer/consumer, scheduler chunk stop). Every
hit snaps to the 3584-token block: a warm re-turn recomputes up to 3583 tokens (0.9–4.0 s measured).

### Change
Overlay `patch_apc_fine_grained_hits.py` (fail-closed, transactional; composes with `patch_hybrid_prefix_hit.py` both orders):
exclude non-participating managers from the veto; enforce the real invariant `hash_block_size % index_kpool == 0`
(index_kpool = 4, verified from live specs at init — violations refuse to boot; participating coarse managers still safely
disable). Effective-value receipt line at init; detects a future image where upstream already scopes the veto (vLLM main
e126687a) and no-ops. Kill switch `GLM53_FINEGRAINED_APC=0`; launcher validates 0/1; 195-check host test.

### Live receipts (on top of the per-group retention PR)
| | before | **with this PR** |
|---|---|---|
| re-turn hit, 45–120K conversations | 96.4–99.4 % | **99.9–100 %** |
| re-turn TTFT | 0.9–4.0 s | **0.3–0.5 s** |
| falsifiable hit-length prediction (P=33641, +9-token turn) | 32256 (3584 grid) | **33600 = the 64 grid, exactly as predicted** |
| L5 decode rate after a warm diverging hit | — | 0.99× cold |
| temp-0 logprob equivalence (3-cold-pair noise floor, `probe_prefix_equivalence.py` v3) | — | **PASS**: prose cold-vs-warm max Δlogprob 0.183 vs floor 0.176 (≤3×), code 0.47 vs 0.35; position-0 token identical throughout |

### Audit log
Codex design review (build-with-changes) → applied; final pass (fix-first: capability cross-check, strict ints, 0/1 kill
switch, mixed-layout safe-disable, receipt line, upstream-fixed detection) → applied; confirm: deploy-for-live-test.

---

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -155,3 +155,6 @@ NCCL_IB_GID_INDEX=3
 # still enabled. 1 = upstream default.
 CG_ESTIMATE=1
 NCCL_DEBUG=WARN
+
+# Fine-grained (64-token) prefix-cache hits (1 = on; 0 = upstream behaviour: hits snap to 3584-token blocks).
+GLM53_FINEGRAINED_APC=1
--- a/.env.example
+++ b/.env.example
@@ -54,6 +54,8 @@ MODEL_REVISION=25a44fdbf16862a46b7cc9921142c6c81350af2f
 # the worker if that node can reach GHCR; otherwise docker save
 # --platform | ssh docker load. No login required.
 # Overlay rebuild from this repo: BUILD=1 ./start.sh
+# start.sh also rebuilds once when Dockerfile/overlay change after git pull
+# (recipe stamp). Keep a pulled GHCR tag: SKIP_BUILD=1
 # Skip the pull (use whatever is already local): SKIP_PULL=1
 # Skip copying to the worker: SKIP_SHIP=1
 # Force the ~164 GiB rsync even when the worker revision marker matches:
@@ -76,6 +78,14 @@ EXL3_FUSED_MOE=1
 # EXL3_MOE_ROW_TILE=0
 # Fused temp rows/expert. 1024 lost the MNBT=1024 A/B (P2b); keep 128.
 # EXL3_TEMP_ROWS_FUSED=128
+# Sorted routing tier; 0 falls back to legacy scans unless a higher tier is on.
+# EXL3_FAT_SORTED=0
+# E1 batched tier: persistent scratch + combined gate/up; implies SORTED=1.
+# EXL3_FAT_BATCHED=0
+# E2 direct kernel (default on). Implies BATCHED=1 and SORTED=1.
+# Needs the patched extension — start.sh rebuilds when the overlay recipe drifts.
+# EXL3_FAT_KERNEL=0 restores the legacy fat-expert path without a rebuild.
+EXL3_FAT_KERNEL=1
 SERVED_MODEL_NAME=GLM-5.3-Flash-EXL3
 
 # Winner on this 2× GB10 kit with DFlash2: fused MoE + DFlash2 k=7.
@@ -90,8 +100,9 @@ DFLASH_DRAFT_TP=2
 MTP_TOKENS=2
 MAX_MODEL_LEN=1000000
 MAX_NUM_SEQS=4
-# P1 keep 2048. 3584/4096 lost (fat-expert LinearEXL3). Never 8192 (indexer smem).
-MAX_NUM_BATCHED_TOKENS=2048
+# Prefill chunk. E2 one-shot keep on this 2× GB10 kit (2026-09-01): 7168.
+# 2048/3548 were similar or slower on 100k–300k. Never 8192 (indexer smem).
+MAX_NUM_BATCHED_TOKENS=7168
 GPU_MEM_UTIL=0.87
 KV_CACHE_DTYPE=fp8
 # 0 = load vision tower (image + video). 1 = language-only.
@@ -140,6 +151,14 @@ GLM53_SUPPRESS_STOPS_IN_REASONING=1
 # When a decode seq is running, do not mix a peer prefill into that step
 # (issue #6). skip = decode-only steps; N>0 = cap mixed prefill tokens; 0 = off.
 GLM53_MIXED_PREFILL_CHUNK=skip
+# Sparse-indexer prefill gather workspace sizing.
+#   stock     = max_model_len * 40 entries (5036.40 MB locked at 1M, measured)
+#   rightsize = the legal per-step maximum, ~+26% KV cache. Opt-in.
+# See docs/DESIGN-indexer-workspace.md. Anything else is rejected at launch.
+GLM53_INDEXER_WORKSPACE=stock
+# SpinCondition reader busy-loop window: stock keeps vLLM's 1 s default;
+# 1..1000 selects milliseconds. This deployment's frozen sweep selected 16.
+GLM53_SPINWAIT_MS=stock
 # EngineCore execute timeout (seconds). Stock 300 is short for a cold JIT.
 # VLLM_EXECUTE_MODEL_TIMEOUT_SECONDS=1800
 
--- a/.env.tp4.example
+++ b/.env.tp4.example
@@ -1,0 +1,57 @@
+# GLM-5.3-Flash EXL3 — EXPERIMENTAL 4× DGX Spark (TP=4)
+# Copy to .env.tp4 and edit. ./start-tp4.sh does this on first run.
+#
+# Sourced AFTER .env, so these values win over the 2× TP=2 knobs
+# (TP=2, NNODES=2, DFLASH_DRAFT_TP=2, NCCL_CROSS_NIC=0). start.sh never
+# reads this file.
+
+# --- ranks (edit these) ------------------------------------------------
+HEAD_IP=10.0.0.1
+WORKER_IP=10.0.0.2
+WORKER2_IP=10.0.0.3
+WORKER3_IP=10.0.0.4
+# Mixed OS accounts (unset = same user as the head / WORKER_USER):
+# WORKER_USER=
+# WORKER2_USER=
+# WORKER3_USER=
+# WORKER2_HOME=/home/you
+# WORKER3_HOME=/home/you
+# WORKER2_SSH=you@10.0.0.3
+# WORKER3_SSH=you@10.0.0.4
+
+# --- fabric --------------------------------------------------------------
+# Four Sparks need a RoCE path among all ranks (QSFP ring on both CX7
+# ports, or a switch). NCCL cannot use the 10.0.0.x loopback aliases.
+# Per-rank pins are what the containers get.
+# Dual-port ring example (comma-separated HCAs / IFs):
+#   HEAD_CX7_IB=rocep1s0f0,rocep1s0f1
+#   HEAD_CX7_IF=enp1s0f0np0,enp1s0f1np1
+NCCL_CROSS_NIC=1
+HEAD_CX7_IF=enp1s0f1np1
+HEAD_CX7_IB=rocep1s0f1
+WORKER_CX7_IF=enp1s0f0np0
+WORKER_CX7_IB=rocep1s0f0
+WORKER2_CX7_IF=enp1s0f0np0
+WORKER2_CX7_IB=rocep1s0f0
+WORKER3_CX7_IF=enp1s0f0np0
+WORKER3_CX7_IB=rocep1s0f0
+# Per-rank RoCEv2 GID (empty = NCCL_IB_GID_INDEX). Inspect:
+#   cat /sys/class/infiniband/<dev>/ports/1/gids/{0..7}
+# HEAD_GID=3
+# WORKER_GID=3
+# WORKER2_GID=3
+# WORKER3_GID=3
+NCCL_IB_GID_INDEX=3
+NCCL_DEBUG=WARN
+
+# --- serve (TP=4) --------------------------------------------------------
+TP=4
+NNODES=4
+# Shard the DFlash2 drafter across all four ranks. 1 = rank 0 only.
+DFLASH_DRAFT_TP=4
+PORT=8888
+MASTER_PORT=29521
+CONTAINER_HEAD=glm53-exl3-tp4-head
+CONTAINER_WORKER=glm53-exl3-tp4-w1
+CONTAINER_WORKER2=glm53-exl3-tp4-w2
+CONTAINER_WORKER3=glm53-exl3-tp4-w3
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 .env
+.env.tp4
 .glm53-exl3-*.inner.sh
 logs/
 tool-eval-bench/
--- a/Dockerfile
+++ b/Dockerfile
@@ -370,6 +370,9 @@ PY
 # Python overlay (exl3.py) is copied AFTER the CUDA compile so edits do not
 # rebuild exllamav3_ext. The aarch64 stub patch must stay in this layer.
 COPY overlay/patch_exl3_ext_aarch64.py /opt/glm53/patch_exl3_ext_aarch64.py
+COPY overlay/patch_exl3_fat_kernel.py /opt/glm53/patch_exl3_fat_kernel.py
+COPY overlay/exl3_fat_gemm.cu /opt/glm53/exl3-fat-kernel/exl3_fat_gemm.cu
+COPY overlay/exl3_fat_gemm.cuh /opt/glm53/exl3-fat-kernel/exl3_fat_gemm.cuh
 
 ARG EXLLAMAV3_COMMIT=c5d9c657966ffeeaa9353f0cc899f18629da4a13
 ENV TORCH_CUDA_ARCH_LIST=12.1a
@@ -418,13 +421,14 @@ RUN set -eux; \
       | tar -xz -C /tmp/exllamav3 --strip-components=1; \
     python3 -c "from pathlib import Path; assert (Path('/tmp/exllamav3')/'exllamav3/modules/quant/exl3.py').is_file()"; \
     python3 /opt/glm53/patch_exl3_ext_aarch64.py /tmp/exllamav3/exllamav3/exllamav3_ext; \
+    python3 /opt/glm53/patch_exl3_fat_kernel.py /tmp/exllamav3/exllamav3/exllamav3_ext /opt/glm53/exl3-fat-kernel; \
     export CPATH="/usr/local/lib/python3.12/dist-packages/nvidia/cu13/include${CPATH:+:$CPATH}"; \
     export CPLUS_INCLUDE_PATH="/usr/local/lib/python3.12/dist-packages/nvidia/cu13/include${CPLUS_INCLUDE_PATH:+:$CPLUS_INCLUDE_PATH}"; \
     export C_INCLUDE_PATH="/usr/local/lib/python3.12/dist-packages/nvidia/cu13/include${C_INCLUDE_PATH:+:$C_INCLUDE_PATH}"; \
     cd /tmp/exllamav3; \
     TORCH_CUDA_ARCH_LIST=12.1a MAX_JOBS=8 \
       pip install --no-deps --no-build-isolation --no-cache-dir .; \
-    python3 -c "import torch; import exllamav3_ext; assert hasattr(exllamav3_ext, 'exl3_moe'), dir(exllamav3_ext); print('exllamav3_ext', exllamav3_ext.__file__, 'exl3_moe=yes')"; \
+    python3 -c "import torch; import exllamav3_ext; assert hasattr(exllamav3_ext, 'exl3_moe'), dir(exllamav3_ext); assert hasattr(exllamav3_ext, 'exl3_fat_gemm'), dir(exllamav3_ext); assert hasattr(exllamav3_ext, 'exl3_fat_gemm_scatter'), dir(exllamav3_ext); print('exllamav3_ext', exllamav3_ext.__file__, 'exl3_moe=yes fat_gemm=yes')"; \
     rm -rf /tmp/exllamav3 /root/.cache/pip
 
 # Keep this AFTER the CUDA compile layer so Python-only hook edits do not
@@ -452,6 +456,10 @@ COPY overlay/patch_xgrammar_termination.py /opt/glm53/patch_xgrammar_termination
 COPY tests/test_xgrammar_termination.py /opt/glm53/test_xgrammar_termination.py
 COPY overlay/patch_kpool_tail_slotmap.py /opt/glm53/patch_kpool_tail_slotmap.py
 COPY tests/test_kpool_tail_slotmap.py /opt/glm53/test_kpool_tail_slotmap.py
+COPY overlay/patch_spinwait.py /opt/glm53/patch_spinwait.py
+COPY tests/test_spinwait_patch.py /opt/glm53/test_spinwait_patch.py
+COPY overlay/patch_indexer_workspace.py /opt/glm53/patch_indexer_workspace.py
+COPY tests/test_indexer_workspace.py /opt/glm53/test_indexer_workspace.py
 COPY overlay/ablit_runtime.py /opt/glm53/ablit_runtime.py
 COPY overlay/patch_ablit.py /opt/glm53/patch_ablit.py
 COPY tests/test_ablit.py /opt/glm53/test_ablit.py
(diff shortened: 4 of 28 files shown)


a2001eric-web:
Tested successfully on a production-like 2x DGX Spark deployment.

Integration tested:

- Mia `main` at `493cb88`
- this PR at `df8942ba` (integrated together with #83)
- GLM-5.3-Flash EXL3/TR3 4bpw, TP=2, FP8 KV
- DFlash2 `k=7`, draft TP=2
- `MAX_MODEL_LEN=524288`, `GPU_MEM_UTIL=0.82`, `MAX_NUM_SEQS=4`
- `GLM53_FINEGRAINED_APC=1`

Boot confirmed:

```text
[glm53-apc-finegrained] Fine-grained prefix-cache hits ENABLED
effective alignment=64 tokens (hash_block_size; scheduler_block_size=3584)
scratch groups checked: {'KpoolTailManager': 4}
```

The live request had 119,995 prompt tokens and the repeated request reported exactly 119,936 cached tokens:

```text
floor(119,995 / 64) * 64 = 119,936
```

119,936 is 64-aligned and is **not** a multiple of 3,584, so this is direct wire-usage evidence that the effective hit grid changed from the scheduler block to the 64-token hash block. Cold TTFT was 174.522s; warm TTFT was 0.581s.

Combined with #83, two distinct 120K conversations were alternated A→B→A→B. Every warm/return request reported 119,936 / 119,995 cached tokens (99.95%) with TTFT 0.567–0.590s.

Additional acceptance passed: forced tool call with required arguments, native vision smoke, both-rank model load, and no new NCCL/CUDA/EngineCore errors.

One combined-wiring hardening suggestion from the deployment: fail closed on both patch artifacts before stopping healthy containers, and pin the overlay order on both ranks as `hybrid -> per-group -> fine-grained`. We also added a static regression that both ranks receive the same fine-grained/per-group environment values.

Thank you — the 64-token alignment is observable and working as designed on the live fork/image.

nood-co1:
Thank you for the combined-wiring run and the 64-token wire evidence. The hardening you suggested is now on this branch as additive commits on top of the commit you tested — `df8942b` is preserved in history, nothing was force-pushed:

1. `cf53bb5` — **fail closed on both patch artifacts before `restart` stops anything.** `main()` runs `validate_numeric_config` *and* a new `validate_overlay_artifacts` on `start|restart` before `stop` is reached. Every mounted overlay — `patch_hybrid_prefix_hit.py` and `patch_apc_fine_grained_hits.py` included — must exist, be non-empty, parse as Python and carry its own distinct identity string (`[glm53-finegrained-apc]`, `[glm53-hybrid-apc]`, target-path / hook markers for the rest), and end with its exact last non-blank line (EOF sentinel), so a `*_PATCH_HOST` pointed at the *wrong* overlay, or a copy truncated anywhere before EOF even if it still parses, is caught, not just a missing one; the chat template and `ablit/LAYER_MAP.json` are gated too; `GLM53_FINEGRAINED_APC` keeps its exact-0/1 guard. This launcher validates what it forwards, so the `GLM53_APC_RETENTION_INTERVAL_SWA` rule (empty/0 or a positive multiple of 3584 ≤ 1,000,000) lives with the launcher that wires that knob, #83. README now states the head `python3` prerequisite.
2. `df9fc04` — **overlay order pinned once for both ranks as `hybrid -> per-group -> fine-grained`.** `GLM53_OVERLAY_ORDER` is the single list (the per-group slot `[ -f ]`-guarded until #83 lands) and `emit_overlay_block` writes it verbatim into the head and worker inner scripts.
3. `4014c5c` — **static + dry-run regression** `tests/test_launcher_rank_parity.py` (104 checks here): `restart` with a bad `GLM53_FINEGRAINED_APC`, or with *any* `*_PATCH_HOST` missing / pointed at a different overlay / empty / unparseable / truncated to its longest still-parseable prefix, or a missing or directory chat template / non-JSON `ablit/LAYER_MAP.json`, makes zero docker or ssh calls; the order pin asserted in the list and in both generated scripts; and a `launch_cluster` dry-run that captures both `docker run` lines (head argv, worker via the `nccl_common[]` replay in the ssh command string) plus every `scp`, asserting head-mount source = scp source = worker-mount for every `/opt/glm53` patch and identical, *present* effective `GLM53_FINEGRAINED_APC` on both ranks for 0 / 1 / default (knobs this branch does not wire are reported as absent on both rather than assumed). A synthetic one-rank mismatch is shown to be reported.

**Local test run (macOS, bash 5.3, pristine `kv_cache_coordinator.py` pulled from the ghcr image):**

```text
=== h84 @ 4014c5c 4014c5c df9fc04 cf53bb5  status: 0 dirty files
  FAIL tests/test_ablit.py
  PASS tests/test_apc_fine_grained_hits.py
  PASS tests/test_bringup_robustness.py
  FAIL tests/test_chat_template.py
  FAIL tests/test_exl3_overlay.py
  PASS tests/test_hybrid_prefix_hit.py
  PASS tests/test_kpool_tail_slotmap.py
  PASS tests/test_launcher_rank_parity.py
  FAIL tests/test_mixed_prefill_decode.py
  PASS tests/test_numeric_config.py
  FAIL tests/test_scheduler_decode_floor.py
  PASS tests/test_start_overrides.py
  PASS tests/test_suppress_stops.py
  PASS tests/test_warm_restart_stdout.py
  PASS tests/test_xgrammar_termination.py
  parity: rc=0 ok=104 fail=0
```
The five `FAIL` rows are environment-only on this Mac (no torch / jinja2 / vllm / scheduler.py; `test_mixed_prefill_decode.py` is a live bench) and fail identically at the tested commit `df8942b`. Every launcher/overlay test passes, including the two wiring tests (`test_kpool_tail_slotmap.py`, `test_xgrammar_termination.py`) that were taught about `GLM53_OVERLAY_ORDER`.

**Dry-run validator receipt on the Spark head (deb8, bash 5.2, Python 3.12).** Temp copy of this branch at `4014c5c` with the *deployed* `.env`, and `docker`/`ssh`/`scp`/`rsync` stubbed first on `PATH` so nothing real can be stopped:

```text
== real head container BEFORE: glm53-exl3-head Up About an hour; started=2026-09-01T01:02:34.588631621Z
==== h84 @ 4014c5c0f5f4fbe48aa8e41cd8814ea5da6361de (temp copy /home/blockbrain/harden-receipt/h84, deployed .env, docker/ssh/scp/rsync stubbed first on PATH)
docker on PATH -> /home/blockbrain/harden-receipt/bin/docker
-- missing-finehit: rc=2 stub docker/ssh calls=0
     overlay artifact missing, unreadable or empty: /nonexistent/patch_apc_fine_grained_hits.py
-- mispointed-finehit: rc=2 stub docker/ssh calls=0
     overlay artifact /home/blockbrain/harden-receipt/h84/overlay/patch_glm_video_placeholders.py does not carry its identity string '[glm53-finegrained-apc]' (wrong file?)
-- bad-finegrained: rc=2 stub docker/ssh calls=0
     GLM53_FINEGRAINED_APC must be exactly 0 or 1 (got: yes)
-- missing-hybrid: rc=2 stub docker/ssh calls=0
     overlay artifact missing, unreadable or empty: /nonexistent/patch_hybrid_prefix_hit.py
-- mispointed-video: rc=2 stub docker/ssh calls=0
     overlay artifact /home/blockbrain/harden-receipt/h84/overlay/patch_hybrid_prefix_hit.py does not carry its identity string 'vllm/model_executor/layers/' (wrong file?)
-- mispointed-drafter: rc=2 stub docker/ssh calls=0
     overlay artifact /home/blockbrain/harden-receipt/h84/overlay/patch_glm_video_placeholders.py does not carry its identity string 'vllm/v1/core/kv_cache_utils.py' (wrong file?)
     (truncated hybrid prefix parses as Python; only the EOF sentinel can refuse it)
-- truncated-hybrid: rc=2 stub docker/ssh calls=0
     overlay artifact /home/blockbrain/harden-receipt/log/h84-truncated_hybrid.py does not end with '    sys.exit(main())' (truncated copy? last line: '    return 0')
-- missing-chat-template: rc=2 stub docker/ssh calls=0
-- control-deployed-env: rc=1 stub docker/ssh calls=8
     [1;31m[glm53-exl3][0m ERROR: set NCCL_IB_GID_INDEX (same index both ranks) or HEAD_GID/WORKER_GID (per rank) in .env to populated indices
     control: first stub calls after the gate passed:
       docker rm -f glm53-exl3-head 
       ssh -T -o BatchMode=yes -o ConnectTimeout=15 blockbrain@10.0.22.2 docker rm -f 'glm53-exl3-worker' 
-- tests/test_launcher_rank_parity.py on deb8:
     rc=0 ok=104 fail=0 launcher rank-parity / order / pre-stop gate OK
== real head container AFTER: glm53-exl3-head Up About an hour; started=2026-09-01T01:02:34.588631621Z
```
Every bad input exits 2 with zero docker/ssh calls; the control run with the deployed `.env` passes the gate and reaches the (stubbed) `stop` on both ranks. The live head container's `StartedAt` is unchanged.

**No live reboot was performed.** The production pair had been restarted by the owner shortly before this run and is in active use, and the deployed launcher (`deploy/dogfood`) carries additional knobs beyond this PR, so a boot from a PR checkout would have changed the deployed configuration. Your boot and A/B receipts remain the live evidence for the tested commit; these commits change only the pre-stop gate, the order pin and the tests, and the generated in-container apply block is byte-identical to `df8942b` apart from the added `[ -f ]`-guarded slot.

Four adversarial review passes (gpt-5.6) were run on the diff before pushing and their findings folded in; two of its suggestions were deliberately not taken: a per-artifact SHA-256 manifest (this gate targets operator error — missing / mis-pointed / truncated / broken files — not tampering, and a digest manifest would couple every overlay edit to a manifest bump), and dropping the `[ -f ]`-guarded sibling slot from the order list (you asked for the combined pin on both PRs; the slot is a no-op unless that artifact is present), and adding EOF comment sentinels inside every overlay file (the exact-last-line check covers truncation without touching files this PR does not own; the stated residual is a line-boundary cut landing on an earlier identical last line while still parsing).

wdkr:
Thanks for PRs 83 and 84! Big improvement in for example OpenCode sessions. This is what my LLM had to say about it:

> Applied both PRs together on 2× DGX Spark (GB10/sm_121, TP=2, EXL3 4bpw,
> DFlash2 k=7, MNBT 2048, max_num_seqs 4): boot reports
> `retention_by_group=[None,None,None,None,None,None,0] (global=None swa_env=0
> eagle_min_exempt=[6])` and fine-grained hits ENABLED at alignment 64 with
> `scratch groups checked: {'KpoolTailManager': 4}`. On a 50K→63K conversation
> the token-denominated counters land at **554,048 hits against a theoretical
> maximum of 554,210 — 99.97% of the ceiling**, i.e. a 16.2 tok/turn shortfall
> against a 64-token block, with warm re-turn TTFT 1.11–1.20 s across two runs.

nood-co1:
Thank you for the independent run and the token-level counters — 554,048 / 554,210 (99.97%) on a 50K→63K conversation is excellent evidence, and yours is the second independent 2× Spark setup to exercise #83+#84 together, with `retention_by_group=[None×6,0]` and fine-grained ENABLED at alignment 64 in the boot lines. We appreciate the careful validation. If you hit anything odd in longer multi-session shapes (the regime #83 targets hardest), receipts like these are the most useful thing you can post.

nood-co1:
Updated to current main (e25aa70) by merging main into the branch — merge commit 8f103ca, plain push, no force-push; tested history preserved (df8942b and the three hardening commits remain ancestors). Conflicts: start.sh (7 hunks — knob capture/config/bool-flag helper/validator call/-e lines vs main's indexer-workspace/spinwait equivalents, both sides kept; both rank-script apply blocks re-derived onto `emit_overlay_block`, never keep-both-merged) and README.md (main's newer rows + this PR's fine-grained row). Interaction fixes inside the merge commit (same shape as #83): `GLM53_OVERLAY_ORDER` += `patch_spinwait.py` + `patch_indexer_workspace.py` in main's own position; artifact gate += the spinwait mount; main's two wiring tests adapted to the order-pin; rank-parity harness pre-sets the new knobs. Interaction (a) checked: this PR's fine-grained overlay and the merged #63 template touch different layers — no overlap found, none expected. Also adopted from the update review: (1) the GLM53_FINEGRAINED_APC caller capture is now setness-aware (`${VAR+1}`), matching the contract main's indexer/spinwait knobs introduced — an explicitly empty caller export now reaches the 0/1 guard and is rejected pre-stop instead of silently losing to `.env` (new C3/C4 regression checks in tests/test_apc_fine_grained_hits.py cover silent-caller, explicit-empty survival, and the rc=2 rejection); (2) the spinwait gate entry identifies the artifact by its own TARGET path (`device_communicators/shm_broadcast.py`) rather than the incidental temp-file name. Re-run green: `bash -n`, tests/test_launcher_rank_parity.py, part-C knob suite, full host suite 17/22 (fails = the 4 known env fails + test_apc_fine_grained_hits needing the in-image kv_cache_coordinator.py, identical at the tested SHA). Host suites re-run green post-merge; no new live-server measurements were taken for this update; the branch's live receipts describe the fine-grained overlay path, which the resolution leaves byte-untouched (the setness fix changes only the launcher's handling of an explicitly empty caller export — previously silently swallowed, now a pre-stop launcher error).

wdkr:
I thought I noticed a drop in per-position drafter acceptance in the logs when running PRs 83 and 84, so I tried to benchmark it. Below is what Fable thinks of it:

---

**A/B on 2x GB10 (TP=2), same image and same input rev, only the two overlay patches toggled: fine-grained hits (#84) reduce DFlash2 acceptance on warm turns; #83 alone does not.**

Setup: GLM-5.3-Flash EXL3 4bpw, DFlash2 drafter k=7, MNBT 2048, max-num-seqs 4, `VLLM_PREFIX_CACHE_RETENTION_INTERVAL_SWA=0` where #83 was on. Four arms, each a fresh boot: stock (neither patch), #83 only, #84 only, both. Per arm, 27 greedy requests with thinking off: 9 cold single-turn prompts, plus 3 six-turn conversations over a document of about 9.4k tokens with fixed assistant turns, so the request bytes are identical across arms. Turn 1 of each conversation is cold, turns 2 to 6 hit the prefix cache. Acceptance is taken from `/metrics` counter deltas per request (`spec_decode_num_accepted_tokens_total` / `spec_decode_num_drafts_total`, plus the per-position counters), so it is drafts-weighted and independent of the client.

| arm | warm hit fraction | cold accept len | warm accept len | warm pos-7 accept | paired warm delta vs stock |
|---|---|---|---|---|---|
| stock | 0.71 | 4.53 | 5.77 | 0.409 | reference |
| #83 only | 0.71 | 4.61 | 6.28 | 0.510 | +0.58 (higher in 13/15) |
| #84 only | 0.97 | 4.72 | 5.49 | 0.378 | -0.27 (lower in 10/15) |
| #83 + #84 | 0.97 | 4.69 | 5.42 | 0.344 | -0.35 (lower in 9/15) |

Accept len = 1 + accepted / drafts. Warm TTFT went from about 3.5 s (stock, #83) to about 1.1 s (#84, both), so the patch does what it says.

Reading:

- Cold acceptance is flat across all four arms, so neither patch changes drafting on a fresh prompt.
- The warm-turn loss appears in exactly the two arms with #84, with and without #83, at the same size (about 0.3 accepted tokens per step, worst at the tail positions). Dense drafter retention in the "#84 only" arm does not rescue it, so this is not the sparse-retention path in #83.
- In a full agentic sweep (OpenCode tool loops, about 50 turns per run) on the same seat I measured 4.82 before the pair and 4.10 after, so the effect is larger in the tool-loop regime than in this probe.
- Plausible mechanism: the drafter's sliding window is 2,048 tokens. On the 3,584 grid every warm turn re-prefilled a 2.5k to 3.5k tail, so the drafter's whole window was rebuilt from fresh hidden states. With 64-token hits only the new tokens are computed and the window comes from cache.
- One more observation I cannot yet explain: at temperature 0 the generated text differs between the #84 arms and stock for identical input. That suggests the target's computation on a 64-aligned hit is not bit-identical to cold. I plan to test cold-versus-warm output equivalence next. If the fine-grained hit is lossy, the acceptance drop would be a symptom rather than the problem.

Caveats: 3 repeats per cell, thinking off, single-stream only. Happy to share the JSON files and the probe script, or re-run with more repeats or specific settings.

nood-co1:
Thank you for the careful A/B — the per-position counter method and the fixed-assistant-turn construction are exactly the right instruments, and your measurement caught something our design doc had flagged as a risk (R4) but never measured. We re-ran your probe shape on our production pair (2× DGX Spark, TP=2, next3 image) with a Codex-reviewed kit: BAAB fresh boots (R_both → R_83 → R_83 → R_both), single-knob contrast — `R_83` = fine-grained hits killed via `GLM53_FINEGRAINED_APC=0`, `R_both` = enabled, per-group SWA retention pinned identical by value in both arms — with per-boot receipts (head rank: authoritative effective alignment/retention boot lines; worker rank: kill-switch environment parity), reps 2/boot (60 warm turns/arm), acceptance from drafts-weighted `/metrics` counter deltas like yours.

**On your acceptance claim: our result is weakly directional in the same direction as yours** (our pre-registered language cap: n = 2 boot pairs; both boot pairs agree in sign, cluster signs are 11/12, not unanimous).

| arm | cold acc-len (pooled) | warm acc-len (pooled) | warm TTFT med | warm hit frac |
|---|---|---|---|---|
| R_83 (fine hits off) | 2.43 / 2.50 | **5.693** | 2.24 s | 0.781 |
| R_both (fine hits on) | 2.47 / 2.51 | **5.159** | 0.50 s | 0.992 |

Boot-pair warm deltas (independent units): −0.361, −0.689; conversation-cluster deltas 11/12 negative (sign-p 0.006, descriptive — clusters are pseudoreplicated). Pooled delta −0.53, a bit larger than your −0.27/−0.35. Cold acceptance is flat across arms (spread 0.08), and — as in your table — each boot's pooled warm acceptance stayed above its pooled cold. Per-position, the arm-pooled rate gap is tail-heavy: ≈0.010 at position 0 growing to ≈0.115 at position 6 (small non-monotonicities at intermediate positions) — same tail-heavy shape as your pos-7 column.

Mechanism (candidate, consistent with source and both datasets, not yet instrumented): on the 3584 grid a warm turn re-prefills a 2.5–3.5k tail ≥ the drafter's 2048-token sliding window, so the drafter's whole window is rebuilt from fresh hidden states every warm turn — an acceptance bonus that costs the multi-second TTFT you measured. A 64-grid hit forgoes that re-prefill; with per-group SWA retention at 0 the drafter resumes on an empty window. Warm-above-cold in both arms is consistent with a *forgone bonus*, not drafter corruption. Your #83-only +0.58 remains unexplained by us too — open item. Definitive attribution needs drop_eagle_block/eagle_verified receipts from an instrumented boot; on our list, not claimed here.

Since alignment keeps coming up as the obvious knob: coarsening to 896 is not the fix. The gate's invariant does generalize (896 = 14×64), but the fine-grained lookup paths are positionally sound only at `alignment == hash_block_size`; 896 over a 64 hash-block overstates every fine hit 14× (silent wrong output), and the sound route (`--prefix-match-unit 896`) is rejected on this layout (896 ∤ 64 drafter SWA block). Even a hypothetical sound 896 grid leaves a grid-rounding remainder of ≤895 tokens (total turn-2 recompute additionally includes the reply and suffix on any grid), still < the 2048 window — so the drafter window still isn't rebuilt. Alignment is the wrong lever for the drafter effect.

**Net effect for the trade you surfaced:** at median completion length 35 tokens in both arms, median warm-turn wall fell 2.94 s → 1.22 s with fine hits on; decode medians were identical (0.66 vs 0.67 s). The acceptance drop costs ≈1.8 ms per generated token against a 1.7 s TTFT saving — breakeven ≈ 1000-token replies (estimate). In this probe regime #84 is clearly net-positive; broader production workloads remain unmeasured. Your agentic-sweep number (4.82 → 4.10) is the regime where the acceptance side weighs most, and a counterbalanced multi-boot window in that shape is the right next measurement.

**On your temp-0 divergence observation: honest status — our equivalence instrument has not yet cleared this deployment's own nondeterminism floor, and we won't claim either direction from it.** What we can report, with receipts: on this pair, two *cold* runs of the byte-identical request (fresh cache-salt namespaces, same boot, idle server) already produce different text at temperature 0. Separated by horizon: short answers are bimodal via single near-tie flips (max |Δlogprob| floor ≈ 1.12 nats); verbose 64-token answers gave cold-vs-cold floors of 8.04 and 11.96 nats in the same boot; independent 400-token cold pairs differed at 349–388 of 400 positions. Three probe designs failed to clear that floor (the first two — natural-EOS length-skew and `ignore_eos` post-EOS comparison — were themselves unsuitable instruments; the third measured the floor directly). Notably, in every window the warm-vs-cold divergence sat at or below the same boot's cold-vs-cold floor — nothing we measured distinguishes a 64-grid warm hit from an ordinary cold rerun — but identity-level equivalence is unresolvable there, so this is evidence about the floor, not a verdict on lossiness. Practical implication for your planned cold-vs-warm test: benchmark against a same-boot cold-vs-cold pair first; temp-0 text differences alone (your "cannot yet explain" observation) are expected under this floor and don't by themselves indicate a lossy hit. We will design and pre-register a dedicated lossiness protocol built for that statistics (warm-within-cold-distribution testing rather than token identity) and post it on this PR.

Probe script, driver, and the four per-boot JSON receipt files are available — happy to push them to this branch or share directly.

plotarmordev:
**Status:** rework — one default change and one trim, then this is the fine-grained candidate we stack on the replay guard.

**Verified:** read the full diff against current main; the core mechanism is the same veto-scoping as #125 (`KpoolTailManager` no longer vetoes 64-token hits), here packaged with a runtime-verified scratch invariant and a kill switch. On the acceptance question we are not asking you to re-litigate the thread data — we accept it as established: wdkr's 4-arm fresh-boot A/B (−0.27 / −0.35 warm accept-length, tail-heavy, cold flat, present with and without #83) and your own pre-registered BAAB re-run (boot-pair deltas −0.361 / −0.689, pooled −0.53, same tail-heavy shape) agree in sign and size, and the forgone drafter-window-replay mechanism is consistent with main's `MIN_NEW` behavior. We also accept your net-wall accounting for the probe regime; the deciding factor for us is that our primary workload is the agentic tool-loop regime, where wdkr measured 4.82 → 4.10 — the acceptance side dominates there.

**Change requests:**

1. **Default `GLM53_FINEGRAINED_APC` to `0` (opt-in) until a replay guard lands.** Two independent probes show a warm accept-length cost of −0.27…−0.53 in exactly the agentic regime; a change that is on-by-default must not ship a known acceptance regression. Keep the strict `0`/`1` validation as-is.
2. **Keep the runtime `hash % kpool == 0` verification and the kill switch.** The `_glm53_finegrained_hit_gate` cross-check (strict-int parse, contradiction = unverifiable, boot refusal only in the one load-bearing cell) is the genuinely load-bearing hardening — it is what makes a 25-line mechanism safe to ship as an overlay. This is the right kind of defense.
3. **Drop the connector receipt and the upstream-fixed detector.** The `truncate_computed_blocks` reachability assert is a log-only nicety, and the vLLM `e126687a` no-op detection is speculative future-proofing against a host image we control and pin. Removing both takes the overlay back toward the mechanism + the load-bearing gate.

**Waiting on:** our cache A/B queue on the 2x GB10 TP=2 E3 kit: #130 (per-group retention + safe DFlash replay, SWA=0, global unset) first; if it passes, the pair experiment — #130 + this PR at the 64 grid with `GLM53_FINEGRAINED_APC=1` — under the identical A/B/A2 fresh-boot protocol with the accept-length criterion tightened to ≥ A−0.05. #130's replay clamp is designed to restore exactly the fresh-window condition these two probes identified (recomputing ~2048 tokens instead of a full 3584 page at the 64 grid), so a pass there is the plausible best-of-both: ~1 s warm TTFT *with* the drafter window rebuilt. If the pair fails on acceptance, this stays opt-in and we revisit.

plotarmordev (changes requested):
Requesting changes — full rationale in https://github.com/MiaAI-Lab/GLM-5.3-Flash-EXL3-2x-DGX-Sparks/pull/84#issuecomment-5603758872. Edit list:
1. Default GLM53_FINEGRAINED_APC to 0 (opt-in): two independent probes show a warm accept-length regression; keep the strict 0/1 validation as-is.
2. Keep the runtime hash % kpool verification and the kill switch — that's the load-bearing hardening.
3. Drop the connector receipt and the upstream-fixed (e126687a) no-op detector.
4. Rebase onto current main dc6936c (same conflict class as #83; rank-parity test union when stacking).
Queued behind the #130 A/B as the hardened fine-grained candidate; no merge default-on in any form.

plotarmordev:
Closing as superseded by #171, not rejecting the fine-grained cache work. Direct comparison confirms that the overlay and design document are byte-identical in both PRs. #171 carries that mechanism on the post-#130 base, avoiding this branch's stale launcher/add-add conflicts while preserving the per-group retention work.

The outstanding change requests are now explicitly recorded on #171: default off at every layer, retain the runtime compatibility check and kill switch, canonical patch-owned-region validation, remove the connector receipt and silent upstream-fixed detector, preserve meaningful host/in-image coverage, and rebase after the independent fixes land. #171 remains held; this closure is not an approval to ship its current default-on version.

Your 64-token hit receipts and warm acceptance A/B measurements remain reference evidence for the successor, with their original attribution here. The unresolved acceptance-length and qualification concerns are carried forward rather than discarded. Thank you for establishing the mechanism and measurements.

PR #87, opened 2026-09-01
launcher: opt-in GLM53_DEFAULT_REASONING_EFFORT (server-side default reasoning effort; empty = unchanged, high recommended)

nood-co1:
### Problem
`files/chat_template.jinja` (line 7) maps `reasoning_effort` to itself only for `low`/`high` and to **`max`** for anything
else — including *unset*. The launcher passes no `--default-chat-template-kwargs`, so every client that omits
`chat_template_kwargs.reasoning_effort` silently runs at max, and nothing in the serve args says so.

Measured on one agentic coding task (OpenCode `build`, frozen microagent brief, 2 reps/arm, frozen 80-point grader, same
vLLM process, one run at a time). The grader was at or near its ceiling (median 79.5–80 of 80), so these small runs primarily measure
**cost**, not reasoning quality:

| effort | median wall | median completion tokens | grader (median) |
|---|---|---|---|
| unset (= max) | 2,160 s | 60,663 | 80/80 |
| high | 593 s (v1: 468 s) | 16,541 (v1: 12,739) | 80/80 |
| low | 252.5 s | 7,241 | 79.5/80 |

Decode rate was comparable in all arms (≈28 completion tok/s), consistent with most of the gap being extra generated
reasoning rather than slower serving (it does not isolate prompt-processing or tool latency). `max` also
compacted twice per run (0 compactions at `high`/`low`). Tool calling worked in every run; a dedicated tool-call canary
passed at `high` and at `low`. Caveats (n = 2 per arm, one task, saturated grader — the numbers are a direction, not two
significant figures) are recorded with the raw per-run tables in `docs/RECEIPTS-default-reasoning-effort.md`.

### Change
- `start.sh`: `GLM53_DEFAULT_REASONING_EFFORT` — **default empty = unchanged behaviour** (no flag is sent; the template
  keeps rendering `Max` for clients that omit `chat_template_kwargs`). Enum is exactly `low | high | max`, validated
  **fail-closed** in the numeric-config guard (literal match: `medium`, `High`, `" high"` are rejected at launch with a
  message naming the knob and the three legal values — `medium` deliberately, because the template has no such level and
  would silently render `Max`). Setness-aware caller capture (an explicit `GLM53_DEFAULT_REASONING_EFFORT= ./start.sh`
  wins over `.env`). When set, `--default-chat-template-kwargs '{"reasoning_effort":"<v>"}'` is appended at **both** rank
  serve-args sites (space-free JSON, one argv element, so it survives the worker's word-split env path); the scalar is
  also exported via `nccl_common`. Request-level `chat_template_kwargs` still override (vLLM merges server defaults first,
  request kwargs win).
- `.env.example`: knob present, **empty**, with `high` recommended for agentic coding. **Not silently enabled** — an
  operator has to set it.
- `README.md`: knob row + a note in the effort section that unset = max.
- `tests/test_default_reasoning_effort.sh` (27 checks, mutation-checked), `tests/test_start_overrides.py` (+ setness-aware
  override case), `docs/RECEIPTS-default-reasoning-effort.md` (A/B tables + the receipts below).

### What this PR does and does not do
- Merging it changes **nothing** for anyone who does not set the knob: launcher default empty → no flag → template
  default (`Max`), exactly as today.
- `high` is the **recommendation** for agentic coding, based on the measurements above. It is documented, not enabled.
- Un-proxied clients that need `max` (or `low`) keep it by sending `chat_template_kwargs.reasoning_effort` per request;
  the receipts below show the request override reaching the template on the live serve.

## Receipts

All collected read-only on the live 2× DGX Spark serve (image `ghcr.io/miaai-lab/glm-5.3-flash-2x-dgx-sparks:exl3`,
digest `sha256:ad0cdd86…`, vLLM `0.1.dev20051+g487ecf187`, model `brandonmusic/GLM-5.3-Flash-tr3-4bpw@792d065`), launched
with `GLM53_DEFAULT_REASONING_EFFORT=high` in `.env`. No restart and no chat completions were issued for these receipts.

### 1. The flag exists in the target image (`--help`)

`vllm serve --help` (the short page) does not list it; `--help=all` does:

```
$ docker exec glm53-exl3-head vllm serve --help=all 2>&1 | grep -n -A2 default-chat-template-kwargs
80:  --default-chat-template-kwargs DEFAULT_CHAT_TEMPLATE_KWARGS
81-                        Should either be a valid JSON string or JSON keys
82-                        passed individually. (default: None)
```

### 2. Both ranks carry the flag, once, as one argv element

The containers run `bash /start.sh` (so `docker inspect .Config.Cmd` is not the serve line); the `vllm serve` process is
pid 1 inside each container, so the receipt reads `/proc/1/cmdline` NUL-split — element index, the following element,
occurrence count:

```
== HEAD ==
host=spark-deb8 container=glm53-exl3-head image=ghcr.io/miaai-lab/glm-5.3-flash-2x-dgx-sparks:exl3
vllm serve pid=1
42:--default-chat-template-kwargs
43-{"reasoning_effort":"high"}
occurrences=1
argc=58
== WORKER ==
host=spark-d9d5 container=glm53-exl3-worker image=ghcr.io/miaai-lab/glm-5.3-flash-2x-dgx-sparks:exl3
vllm serve pid=1
43:--default-chat-template-kwargs
44-{"reasoning_effort":"high"}
occurrences=1
argc=59
```

(The worker's index is one higher because of `--headless`; `ps -eo args` in both containers shows the full line:
`… --kv-cache-dtype fp8 --default-chat-template-kwargs {"reasoning_effort":"high"} --speculative-config …`, with
`--node-rank 0` on the head and `--node-rank 1 … --headless` on the worker.)

Parsed config, from the head's API-server log (the headless rank runs no API server and prints no such line — its proof
is the argv above):

```
$ docker logs glm53-exl3-head 2>&1 | grep -i default_chat_template_kwargs
(APIServer pid=1) INFO 09-01 01:03:44 [api_utils.py:273] non-default args: {'model_tag': '…/GLM-5.3-Flash-tr3-4bpw/snapshots/792d065…', 'chat_template': '/opt/glm53/chat_template.jinja', 'default_chat_template_kwargs': {'reasoning_effort': 'high'}, 'enable_auto_tool_choice': True, 'tool_call_parser': 'glm47', … 'reasoning_parser': 'glm45', … 'nnodes': 2, 'tensor_parallel_size': 2, …}
```

### 3. Render boundary — the server default reaches Jinja, and request kwargs override it

`POST /tokenize` renders the chat template through the same path as chat completions. Because `token_strs` are
tokenizer pieces (`Ġ` markers), each result is also passed back through `POST /detokenize` to show the literal prompt.
Served template is byte-identical to `files/chat_template.jinja` (md5 `37639425…` in the container and in the checkout).

```
=== no chat_template_kwargs  -> server default (high) lands
REQ: {"model":"GLM-5.3-Flash-EXL3","messages":[{"role":"user","content":"hi"}],"return_token_strs":true}
RESP: {"count":13,"max_model_len":1000000,"tokens":[154822,154824,154826,25062,287,29905,371,25,5124,154827,6023,154828,154841],"token_strs":["[gMASK]","<sop>","<|system|>","Reason","ing","ĠEff","ort",":","ĠHigh","<|user|>","hi","<|assistant|>","<think>"]}
DETOK: {"prompt":"[gMASK]<sop><|system|>Reasoning Effort: High<|user|>hi<|assistant|><think>"}

=== "chat_template_kwargs":{"reasoning_effort":"low"}  -> request override wins
RESP: … "token_strs":["[gMASK]","<sop>","<|system|>","Reason","ing","ĠEff","ort",":","ĠLow","<|user|>","hi","<|assistant|>","<think>"]}
DETOK: {"prompt":"[gMASK]<sop><|system|>Reasoning Effort: Low<|user|>hi<|assistant|><think>"}

=== "chat_template_kwargs":{"reasoning_effort":"max"}  -> request override wins
RESP: … "token_strs":["[gMASK]","<sop>","<|system|>","Reason","ing","ĠEff","ort",":","ĠMax","<|user|>","hi","<|assistant|>","<think>"]}
DETOK: {"prompt":"[gMASK]<sop><|system|>Reasoning Effort: Max<|user|>hi<|assistant|><think>"}

=== "chat_template_kwargs":{"reasoning_effort":"medium"}  -> template has no such level, falls back to Max
      (this is why the launcher enum rejects `medium`)
token_strs: ['[gMASK]', '<sop>', '<|system|>', 'Reason', 'ing', 'ĠEff', 'ort', ':', 'ĠMax', '<|user|>', 'hi', '<|assistant|>', '<think>']
```

The token id at position 9 is the only difference between the three renders: `5124` (`ĠHigh`), `12035` (`ĠLow`),
`7487` (`ĠMax`).

**"Before" state (knob empty).** Capturing `/tokenize` on a serve launched *without* the knob would need a restart, which
was not done for these receipts (the box was in use). The template default is instead shown by rendering the served
template file offline inside the same container with jinja2 (`jinja2.ext.loopcontrols`, as `tests/test_chat_template.py`
does), which is exactly the branch a no-flag serve takes:

```
$ docker exec glm53-exl3-head python3 /tmp/render.py      # renders /opt/glm53/chat_template.jinja, messages=[user:"hi"]
 <no kwargs> -> '[gMASK]<sop><|system|>Reasoning Effort: Max<|user|>hi<|assistant|><think>'
         low -> '[gMASK]<sop><|system|>Reasoning Effort: Low<|user|>hi<|assistant|><think>'
        high -> '[gMASK]<sop><|system|>Reasoning Effort: High<|user|>hi<|assistant|><think>'
         max -> '[gMASK]<sop><|system|>Reasoning Effort: Max<|user|>hi<|assistant|><think>'
      medium -> '[gMASK]<sop><|system|>Reasoning Effort: Max<|user|>hi<|assistant|><think>'
```

It is also the state every prior measurement was taken in: the A/B v1 "direct" arm hit this serve before the knob
existed and behaved as `max` (60,663 median completion tokens, 2 compactions/run).

### 4. A/B measurements

Two runs, both on the same live serve, same task, same grader, one run at a time, server never reconfigured mid-run:

- **v1 — high vs unset(max)** (`AB-EFFORT-RESULTS-2026-09-01.md`): high 498/438 s vs max 1,955/2,365 s; all four 80/80.
  Medians: 468 s vs 2,160 s (4.61×), completion tokens 12,739 vs 60,663 (4.76×), compactions 0 vs 2.
- **v2 — high vs low** (`AB-EFFORT-V2-RESULTS-2026-09-01.md`): high 486/700 s (80/80, 80/80) vs low 297/208 s (79/80, 80/80).
  Medians: 593 s vs 252.5 s, completion tokens 16,541 vs 7,241, tool errors 0/71 vs 1/36 (self-recovered), compactions
  0/0. Tool-call canary (`opencode run` with a `read` tool) passed at both efforts. The single lost point at `low` was a
  41-line README against a 40-line limit.

Per-run tables and all caveats are in `docs/RECEIPTS-default-reasoning-effort.md`. What is **not** evidenced: a
`high`-vs-`low` quality difference on a task that is not at the grader ceiling — hence the recommendation is `high`, not
`low`.

### 5. Host tests (this Mac, no hardware)

```
$ bash tests/test_default_reasoning_effort.sh
ok   guard [] -> rc 0 … ok   guard [medium] -> rc 2 … ok   guard [<unset>] -> rc 0
ok   HEAD_SCRIPT [] -> no flag … ok   HEAD_SCRIPT [max] -> {"reasoning_effort":"max"}
ok   WORKER_SCRIPT [] -> no flag … ok   WORKER_SCRIPT [max] -> {"reasoning_effort":"max"}
ok   both ranks build the identical flag
ok   caller capture is setness-aware
ok   caller value wins over .env
ok   knob reaches both containers via nccl_common
ok   launcher default is empty (unchanged behaviour)
default reasoning effort tests: PASS            (27 checks)

$ python3 -m pytest tests -q
26 passed in 1.02s
```

Applies cleanly on `main` @ 493cb88 (`git merge-tree` clean).

### Audit log
Codex gpt-5.6-luna advisory (build-with-changes: keep launcher default empty, recommend `high`, enum `low|high|max` —
drop `medium`, receipts: target-image `--help`, both-rank argv/config, render-boundary, A/B metrics + tool canaries) →
applied. Final pre-deploy pass: deploy-for-live-test, with the note that `/tokenize` `token_strs` are tokenizer pieces
and the receipt should detokenize → done above. Pre-open pass on the final diff + body: OPEN-WITH-EDITS, no code blockers; four
wording fixes adopted (grader "at or near" its ceiling — v2-low scored 79/80 once; the wall-time gap is *consistent
with* extra reasoning, not isolated proof; guided-JSON was not tested; `start.sh` comment now reads "empty or unset =
send no flag").

---

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -140,6 +140,17 @@ GLM53_SUPPRESS_STOPS_IN_REASONING=1
 # When a decode seq is running, do not mix a peer prefill into that step
 # (issue #6). skip = decode-only steps; N>0 = cap mixed prefill tokens; 0 = off.
 GLM53_MIXED_PREFILL_CHUNK=skip
+# Server-side default reasoning effort ("" | low | high | max), sent as
+# --default-chat-template-kwargs. Empty (the default) sends no flag -- and
+# files/chat_template.jinja then renders "Reasoning Effort: Max" for every
+# client that omits chat_template_kwargs, because it maps the value to itself
+# only for low/high and to max otherwise. Recommended for agentic coding:
+#   GLM53_DEFAULT_REASONING_EFFORT=high
+# Measured 80/80 vs 80/80 against unset(max) at 4.6x less wall time and 4.8x
+# fewer completion tokens (docs/RECEIPTS-default-reasoning-effort.md).
+# Per-request chat_template_kwargs.reasoning_effort still overrides this.
+# 'medium' is rejected: the template does not recognize it (it renders max).
+GLM53_DEFAULT_REASONING_EFFORT=
 # EngineCore execute timeout (seconds). Stock 300 is short for a cold JIT.
 # VLLM_EXECUTE_MODEL_TIMEOUT_SECONDS=1800
 
--- a/README.md
+++ b/README.md
@@ -412,6 +412,28 @@ curl -s http://127.0.0.1:8888/v1/chat/completions \
 Thinking defaults on. Disable it with the **top-level** JSON field
 `"chat_template_kwargs": {"enable_thinking": false}`. This closes the empty
 thinking block in the generation prompt and omits the reasoning-effort hint.
+
+**Reasoning effort defaults to `max`, not to something cheap.** `files/chat_template.jinja`
+maps `reasoning_effort` to itself only for `low` and `high`, and to `max` for
+everything else — *including when the field is absent*:
+
+```jinja
+{%- set effective_reasoning_effort = reasoning_effort if reasoning_effort is defined and reasoning_effort in ['low', 'high'] else 'max' -%}
+```
+
+So any client that does not send `chat_template_kwargs` (OpenCode, Cline, plain
+`curl`, most SDK defaults) silently gets the most expensive setting. On one
+agentic build task, unset(max) and `high` scored an identical 80/80 while max
+took **4.6×** the wall time and generated **4.8×** the completion tokens
+(`docs/RECEIPTS-default-reasoning-effort.md`).
+
+Set a server-side default with `GLM53_DEFAULT_REASONING_EFFORT` in `.env`
+(recommended `high` for agentic coding); the launcher then passes
+`--default-chat-template-kwargs '{"reasoning_effort":"high"}'` on both ranks.
+The launcher default stays empty, so behaviour is unchanged unless you set it.
+A request that sends `"chat_template_kwargs": {"reasoning_effort": "low"}`
+still overrides the server default — vLLM merges the server defaults first and
+lets request kwargs win.
 Do not send a literal nested `extra_body` object over raw HTTP; `extra_body` is
 an OpenAI Python SDK option that merges its contents into the top-level request.
 The Hub `generation_config.json` stamps `temperature=1.0` / `top_p=0.95` unless
@@ -489,6 +511,7 @@ that are now documented/enforced:
 | `MAX_NUM_BATCHED_TOKENS` | `2048` | prefill chunk (P1 keep). 3584/4096 lost; 8192 oversubscribes GB10 indexer topk |
 | `GLM53_MIXED_PREFILL_CHUNK` | `skip` | do not mix a peer prefill into a decode step (issue #6). `N>0` = cap tokens; `0` = off. Solo prefill stays MNBT (2048) |
 | `GLM53_SUPPRESS_STOPS_IN_REASONING` | `1` | ignore client `stop` strings until `</think>` (thinking-on default) |
+| `GLM53_DEFAULT_REASONING_EFFORT` | *(empty)* | server-side default effort via `--default-chat-template-kwargs`. Empty sends no flag, and the template then renders **Max** for any client that omits `chat_template_kwargs`. `high` is the recommendation for agentic coding: same 80/80 grader score as unset(max) at **4.6×** less wall time and **4.8×** fewer completion tokens ([receipts](docs/RECEIPTS-default-reasoning-effort.md)). `low`/`high`/`max` only — `medium` is rejected at launch because the template does not recognize it |
 | `GLM53_BOOT_SHAPE_WARMUP` | `1` | after `/health`, burn DFlash2 BLOCK / sampler / kpool shapes (nonfatal) |
 | `TRITON_HOST_CACHE` / `TILELANG_HOST_CACHE` | `$CACHE_ROOT/triton` / `tilelang` | persist JIT caches across container recreate |
 | `LANGUAGE_MODEL_ONLY` | `0` | load vision tower (image + video) |
@@ -537,6 +560,7 @@ this Dockerfile instead. After CUDA compile, Python overlay edits
 | `overlay/patch_ablit.py` | install the load_weights hook; bind-mounted and run on both ranks |
 | `ablit/` | direction vectors + `LAYER_MAP.json` from drowzeys' published recipe; `fetch_transplant.py` + `transplant/` for the donor o_proj byte-copy |
 | `tests/test_ablit.py` | recipe integrity, orthogonalization math, TP-shard equivalence, transplant byte-copy + TP slice, hook gating |
+| `tests/test_default_reasoning_effort.sh` | `GLM53_DEFAULT_REASONING_EFFORT` enum guard (`""`/`low`/`high`/`max`; `medium` rejected) and the `--default-chat-template-kwargs` flag at both rank sites, sliced out of `start.sh` and evaluated |
 | `scripts/boot-shape-warmup.sh` | post-`/health` DFlash2 k=7 BLOCK ladder + sampler/kpool arms |
 
 Image-build runs `EXL3_SELFCHECK_GPU=0`. `./start.sh` runs the GPU self-check
--- a/docs/RECEIPTS-default-reasoning-effort.md
+++ b/docs/RECEIPTS-default-reasoning-effort.md
@@ -1,0 +1,133 @@
+# `GLM53_DEFAULT_REASONING_EFFORT` — receipts
+
+Server-side default reasoning effort for the **GLM-5.3-Flash EXL3** 2× DGX Spark serve, via
+vLLM's `--default-chat-template-kwargs`. The launcher default is **empty** — this PR changes
+nothing until an operator sets the knob.
+
+## Why the knob exists
+
+`files/chat_template.jinja` line 7:
+
+```jinja
+{%- set effective_reasoning_effort = reasoning_effort if reasoning_effort is defined and reasoning_effort in ['low', 'high'] else 'max' -%}
+```
+
+The value maps to itself only for `low` and `high`. Everything else — including **undefined** —
+becomes `max`. So every client that omits `chat_template_kwargs` (OpenCode, plain `curl`, most
+SDK defaults) silently gets the most expensive setting, and no serve flag says so.
+
+That also fixes the enum: **`low | high | max` only**. `medium` is rejected at launch, because
+the template does not recognize it and would render `Max` while the operator believed otherwise.
+
+## A/B — unset(max) vs `high` (2026-09-01)
+
+One frozen agentic build brief, run twice per arm on the live serve. Arm A reached the server
+through a proxy injecting `chat_template_kwargs.reasoning_effort=high`; arm B hit the server
+directly, i.e. **the server default, which on this template is `max`**. Server flags were not
+changed and the server was not restarted. Grader is a frozen 80-point rubric.
+
+| Run | Arm | Wall (s) | Grader | Turns | Tool calls | Compactions | Prompt tok | Completion tok |
+|---|---|---:|---:|---:|---:|---:|---:|---:|
+| A-1 | high | 498 | **80/80** | 29 | 30 | 0 | 562,519 | 13,856 |
+| A-2 | high | 438 | **80/80** | 26 | 32 | 0 | 447,190 | 11,621 |
+| B-1 | unset → max | 1,955 | **80/80** | 14 | 21 | 2 | 303,872 | 54,410 |
+| B-2 | unset → max | 2,365 | **80/80** | 20 | 31 | 2 | 399,109 | 66,915 |
+
+Medians (n = 2 per arm):
+
+| Metric | high | unset → max | max ÷ high |
+|---|---:|---:|---:|
+| Grader score | 80.0 / 80 | 80.0 / 80 | 1.00 |
+| Wall time | **468 s** | **2,160 s** | **4.61×** |
+| Completion tokens | **12,739** | **60,663** | **4.76×** |
+| Prompt tokens | 504,855 | 351,491 | 0.70× |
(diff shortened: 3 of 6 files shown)


nood-co1:
Updated to current main (e25aa70) by merging main into the branch — merge commit 7773343, plain push, no force-push; tested history preserved as ancestors. Conflicts: start.sh (7 hunks, all adjacent insertions vs main's indexer-workspace/spinwait knobs — both sides kept; the shared enum-guard comment merged without duplication), .env.example, README.md (main's newer rows + this PR's effort row), tests/test_start_overrides.py (adopted main's `_run_preamble`; this PR's setness-aware effort test rewritten onto it; helper's env scrub extended to GLM53_DEFAULT_REASONING_EFFORT). Interaction checked explicitly: #63 (merged) made the `Reasoning Effort:` line render unconditionally for prefix-cache stability — the line this PR's semantics rest on (`reasoning_effort` maps to itself only for low/high, else max) is byte-identical before/after #63, so the knob's documented behavior and recommendation stand. Also adopted from the update review: the pre-existing README sentence claiming `enable_thinking:false` "omits the reasoning-effort hint" was stale against #63 (the line now renders unconditionally) — corrected here since it contradicts this PR's subject. Re-run green: `bash -n`, test_start_overrides, test_chat_template, full host suite 16/20 (4 known env fails identical on upstream/main). Host suites re-run green post-merge; live receipts unchanged from c65f811 (no overlay-behavior change in the resolution). A read-only /tokenize re-receipt of the injected default against a server carrying the merged #63 template is pending a safe idle window (the only #63-template deployment we can reach is mid-experiment); will follow up on-thread.

nood-co1:
**Re-receipt: render-boundary behaviour of this knob against the #63-era template** (owed since the 2026-09-02 rebase pass, which corrected this PR's README wording against #63 and left the template re-receipt pending).

**Template provenance first**, because it changes how the original receipts should be read:

| ref | `files/chat_template.jinja` | note |
|---|---|---|
| `main@493cb88` (base this PR applied to) | sha256 `96ed8316…`, md5 `37639425…` | pre-#63 |
| `main@c190db1` → current `main` | sha256 `3050de3f…`, md5 `1f5c6f5f…` | #63-era |

The PR body's receipts (md5 `37639425…`) were therefore collected on the **pre-#63** template. The whole of #63 is one line — it drops the `thinking_enabled` guard from the `Reasoning Effort:` marker, so the marker renders regardless of thinking state whenever an effective effort is present (protecting the prefix cache across thinking toggles). The `reasoning_effort` mapping itself (`low`/`high` → themselves, anything else including *unset* → `Max`) is untouched.

**1. Live render-boundary receipts, pre-#63 template** (read-only `POST /tokenize` + `/detokenize` against a live serve launched with `GLM53_DEFAULT_REASONING_EFFORT=high`; its served template copy is md5-verified byte-identical to `main@493cb88`'s `files/chat_template.jinja`):

```
no chat_template_kwargs                              -> …<|system|>Reasoning Effort: High<|user|>hi…   (server default lands)
{"reasoning_effort":"low"}                           -> …Reasoning Effort: Low…                        (request override wins)
{"reasoning_effort":"max"}                           -> …Reasoning Effort: Max…
{"reasoning_effort":"medium"}                        -> …Reasoning Effort: Max…    (request succeeds, silently normalizes to Max — why the launcher enum refuses it)
{"enable_thinking":false}                            -> [gMASK]<sop><|user|>hi<|assistant|><think></think>   (marker dropped)
{"enable_thinking":false,"reasoning_effort":"high"}  -> [gMASK]<sop><|user|>hi<|assistant|><think></think>   (still dropped)
```

Same behaviour as the PR body's receipts — and the two `enable_thinking:false` rows document the pre-#63 marker drop that #63 removes.

**2. The #63-era template — offline render** (this leg is a template-level result, not a live re-receipt: jinja2 `Environment(extensions=["jinja2.ext.loopcontrols"])`, `messages=[user:"hi"]`, `add_generation_prompt=True`, over current `main`'s template bytes (sha256 `3050de3f…`) — the same substitution the PR body already uses for its knob-empty "before" state):

```
no kwargs                                  -> …Reasoning Effort: Max…      (unset -> Max, unchanged)
reasoning_effort=low/high/max              -> Low / High / Max              (unchanged)
reasoning_effort=medium                    -> …Reasoning Effort: Max…      (fallback unchanged)
enable_thinking=false                      -> …<|system|>Reasoning Effort: Max<|user|>hi<|assistant|><think></think>
enable_thinking=false, reasoning_effort=high -> …<|system|>Reasoning Effort: High<|user|>hi<|assistant|><think></think>
```

**Conclusion: the tested effort mappings render identically on the #63-era template; no change to this PR is needed.** The knob only appends `--default-chat-template-kwargs`; the template decides rendering. Unset still renders `Max` (the problem this PR documents), request kwargs still win, and `medium` still silently normalizes to `Max` (the launcher enum's rationale). The one intentional delta #63 introduces is the non-thinking case: the pre-#63 template drops the effort marker when thinking is disabled (last two live rows above), while the #63-era template renders it there too — by template-level inference, a server default set via this knob would then also reach non-thinking prompts, and the rendered prefix stays cache-stable across thinking toggles, per #63's intent. (That inference is from the offline render; it was not exercised on a live #63-template serve.)

---

plotarmordev:
**Status:** mechanism confirmed — queued for a boot A/B after #112's window. One rebase request and one receipt confirmation.

**Verified:** current main's `files/chat_template.jinja` still maps an unset `reasoning_effort` to `Max`, so clients that omit it burn the max-effort token budget by default — the problem statement is live. The knob is correctly minimal: it only appends `--default-chat-template-kwargs` and the template decides rendering; empty = unchanged. We also read your `7773343` merge and the re-receipt against the #63-era template — agreed that the tested effort mappings render identically there (unset → Max, request kwargs win, `medium` normalizes to Max), and that #63's always-render change only extends the marker to non-thinking prompts.

**Change requests:**

1. **Rebase onto current main** — the PR is CONFLICTING again; main moved past the `e25aa70` you merged (the adaptive-k + dense-fp8 merge touched the same start.sh / README / `.env.example` regions).
2. **Confirm the receipt still holds on the current template once rebased.** One interaction note to make that easy: #127 (the upstream chat-template port) is approved from our side pending its live smoke, and we verified its 6 hunks hunk-by-hunk against the upstream commit — all are in the tool-call/validation region; none touches the effort mapping line. So no interaction is expected whichever order these land in, but the re-receipt should be against whatever template main carries at merge time.

**Waiting on:** a boot A/B on our 2x GB10 TP=2 E3 kit, queued right after #112's fairness window: an agentic task battery with the knob unset (server default = Max) vs `high`, comparing wall time and token cost at equal task grade — replicating your n=2/arm cost table shape on the current image. Results posted here.

plotarmordev:
Merged via https://github.com/MiaAI-Lab/GLM-5.3-Flash-EXL3-2x-DGX-Sparks/pull/158 (rebased onto current main with maintainer fixes; your commits and authorship are preserved). Thank you for the contribution — closing this one in favour of the merged replacement.

PR #98, opened 2026-09-01
Add non-Blackwell KLD lane (4x CMP 170HX, sm80) + measured AWQ W4A16 row

Karl0007:
## What

The fidelity suite shipped with the KLD panel is SM120-only (`Dockerfile` builds SM120/SM121a cubins, `overlay/exl3.py` is the only capture path). This adds a **second engine** on a different architecture family — vLLM 0.11.2 fork, PP4 (14,12,12,7), Ampere sm80, 4×64 GB — plus one measured row, so a non-Blackwell box can put its own candidate on your yardstick instead of extrapolating from a Blackwell-only panel. Purely additive (21 files, no deletions): nothing touches the SM120 path, the SM120 scripts, or the suite bytes.

`bench/glm53-kld/README.md` documents the protocol and the traps; `bench/glm53-kld-runs/` ships the reports (anchor, self-canary, our row, one control).

## Why it's a measurement and not a vibe

Validated against the published rows before trusting any of it:

| check | result |
|---|---|
| reference lane vs itself | mean KLD **exactly 0.000000**, JSD 9e-9 (float32 floor), top-1 1.000000 |
| FP8 lane vs reference | reproduces `reports/report-fp8-vs-bf16.json`: scope-mean drift **1.5e-5** (48 contexts), median per-context delta 5.7e-5 |
| head equality (our side) | AWQ `lm_head.weight` bit-identical to `head/head.safetensors` |
| lane geometry | our lane's per-token RMS distribution (min .670 / median 1.369 / max 1.445) matches the reference lane (.660 / 1.370 / 1.445) — see trap #2 for why the *shape* is the check |
| `selftest.py` | 13 named CPU checks, all pass, incl. the cut-point refusal and the bf16/fp32 dtype rule |

## The row

```
cmp170hx-awq-w4a16   0.0770 nats   95% CI [0.0696, 0.0854]
108 contexts · 221,076 positions · 97 source clusters · top-1 0.9085 · p999 3.17
strata: literary 0.088 > code 0.085 > encyclopedic 0.080 > scientific 0.070 > multilingual 0.059
```

FP8 on the same 108 indices = 0.0269 → **2.86×**. Rank against your panel (full scope): K6 0.0137 < TR3-4bpw 0.0255 < K4 0.0273 < FP8 0.0281 < K3 0.0505 < **AWQ 0.077** < K2 0.155.

Two things worth flagging:

- **GSM8K-100 on these same weights is 99/100.** 9.2% of positions disagree with the teacher's argmax and the tail carries most of the distance — exactly what a pass@1 benchmark can't see, which is why I'd argue your panel is the right gate for MoE quants.
- **`d(AWQ, official FP8 lane) = 0.0836` > `d(AWQ, BF16) = 0.0770`**: our error is independent of FP8's, not an FP8-like perturbation.

**Scope, stated on the row.** 112 windows were captured (indices 0-114); 108 were scored. Four of them (63, 68, 90, 113) have no reference shard on this host because the reference-lane fetch stopped partway when the mirror went unreachable (384 shards present, index range 0-428, not contiguous; the FP8 lane has 243). The loss is scattered across indices rather than concentrated in a stratum, and finishing the fetch extends the row to 112 by re-scoring on CPU — no GPU hand-off needed.

**What the number cannot separate.** 0.077 includes *our engine* (PP4 + an mHC handoff patch, eager capture) measured against a teacher captured elsewhere. We hold no BF16 GLM-5.3-Flash weights, so there is no engine term to subtract; your own native-BF16 recapture floor was 0.0115 nats. Treat 0.077 as an **upper bound** on the AWQ quantization, not a decomposition. A TP4 recapture on the same windows is the control that would split the two, and it's queued on our side.

## Traps that cost hours, each with a loud check now

1. **A `__main__`-defined callable can't be pickled across the engine.** Workers resolve by qualified name, and `__main__` there is the worker entry script, so `apply_model` dies with *"it's not the same object as `__main__.install_tap`"*. Installer moved into an importable module on `PYTHONPATH` (pickle by reference).
2. **A pre-last-layer norm is not a valid cut point.** Our first tap hooked the mHC pre-norm and reported **0.0002 nats** on a lane whose per-token RMS peaked at **17.8** — the norm had swallowed the difference, so the number was quietly meaningless. `register_hidden_capture` now asserts the hook target is an `RMSNorm` instance and prints the class, and the lane's RMS distribution is checked against your published cut-point stats (re-normalising pins every row at ‖norm.weight‖ = 1.4315; capturing before it spreads widely with depth — the *shape* is the evidence, not the mean).
3. **The head replay must be bfloat16.** On the first 30 published contexts, fp32 gives 1.2e-2 median per-context delta vs **5.9e-5** for bf16. bf16 rounding noise (~0.08 nats) inflates KL by ~0.003 nats (~10% here), so an fp32 candidate would beat every published row by 1–4% for a reason unrelated to quantization. Default `--dtype bfloat16`; the fp32 run is kept in `runs/` as the negative control.
4. **`mm_processor_kwargs={}` breaks a text-only model** whose processor class is defaulted: vLLM probes `<model_dir>/processor_config.json` and rejects the model as multimodal — for every registry candidate at once. Omit it unless the processor class is explicit.
5. **A container capture lands root-owned**, so the host scorer cannot read its own lane.

## Test plan

- `python3 bench/glm53-kld/selftest.py` → 13 named checks pass (CPU only, no suite download needed)
- `MODE=self bash bench/glm53-kld/capture-glm-awq.sh` → KLD exactly 0
- `MODE=anchor bash bench/glm53-kld/capture-glm-awq.sh` → matches the published FP8 row to 1e-5
- `MODE=capture` / `MODE=score` need all four cards; the parity flags that differ from production (MTP off, eager, 2048-token context) are listed in the README and recorded in the lane's `capture-receipt.json`

Co-Authored-By: Claude <noreply@anthropic.com>


Karl0007:
Closing this. It was the wrong shape for this repo, and while checking the suite README for the reply I found that two of my claims here were overstated.

**Wrong venue.** The panel says out loud "This scores the **weights**, not this GB10 overlay" — the repo never claims to measure its own serving stack, so a capture path for a different engine doesn't fill a gap you acknowledge. The readers here own GB10; they can't run a 4×64 GB Ampere capture, and four of my five traps are vLLM-specific (`__main__` pickling, `mm_processor_kwargs`, container ownership, `apply_model`), so they don't apply to the exllamav3 path. Dropping 3.1k lines of bench code into a serving kit is maintenance you didn't ask for. The tooling lives in our own repo instead.

**Claim 1 overreach.** "The head replay must be bfloat16" — it must match the published protocol, which is not the same statement. Your README already cross-validates against a separate full-vocab **fp32** pipeline at 1.27e-2, which is the same order as the fp32-vs-published delta we measured (1.17e-2). fp32 replay is a legitimate yardstick, not a bug; bf16 is just what reproduces the rows in the table.

**Claim 2 wrong.** I wrote that we "cannot subtract an engine term" for our 0.077, and that isn't true — you publish the drift on this architecture: 8.7e-4 run-to-run floor, **1.49e-2** live-vs-replayed across launches, 1.27e-2 cross-pipeline. So ours decomposes to roughly **0.062 attributable to the weights** (still ~2.3× the FP8 row, ordering unchanged), and the useful rule I took from your receipts is that two candidates within ~0.015 nats of each other aren't separable by this protocol at all. Worth stating in the panel docs next to the table, since it bounds what the rows can decide.

The row itself stays in our repo (`bench/glm53-kld`), with the scope caveats (108 of 112 captured — our reference-lane fetch was partial; drift term subtracted from your published receipts, not ours).

Thanks for the repo — the suite is a good yardstick and the cut-point note in `capture-cut-point.json` is what caught a bug in our capture.

PR #77, opened 2026-08-31
perf(exl3): accelerate fat-expert prefill

plotarmordev:
## Summary

Accelerate EXL3 routed-expert prefill without changing the decode path:

- E1: sort routed rows by expert and process fat experts as contiguous batches.
- E2: add fused direct and scatter CUDA kernels for fat-expert GEMM.
- Keep the existing thin-expert path and legacy fallback.
- Expose tiered `EXL3_FAT_SORTED`, `EXL3_FAT_BATCHED`, and `EXL3_FAT_KERNEL` rollout switches; higher tiers imply their prerequisites.

## Implementation

- Patch the pinned exllamav3 extension during the image build and compile `exl3_fat_gemm` for SM121.
- Add direct-output and scatter-output bindings.
- Cache scratch tensors and staged expert counts to avoid repeated allocations on the hot path.
- Preserve CUDA-graph-safe decode behavior; the new paths apply to fat-expert prefill.
- Cover direct/scatter parity, isolated legacy/sorted/batched/kernel flag selection, mixed thin/fat composition, and CUDA graph capture.

## Verification

Executed from clean PR head `4b8d3c7` based on `origin/main` `493cb88`:

- Python compilation and launcher syntax checks.
- Full Docker build with pinned exllamav3 download, extension compilation, binding assertions, and image self-checks.
- GPU checks for finite EXL3 output, E2 direct/scatter parity, tier isolation, mixed thin/fat composition, and CUDA graph capture.

A local diagnostic follow-up (`86e09e0` + `f1d1d83`, not yet pushed) adds a stable 32-key receipt, both-rank tier/symbol/fallback/call/workspace counters, and fail-closed symbol handling. Its GPU self-check and live on/off/on campaign passed.

## Performance evidence — local promotion sequence complete

The original 64K/100K figures remain withdrawn because they included APC hits. The earlier corrected one-shot table was diagnostic only and is superseded by this repeated sequence.

Protocol:

- 2× DGX Spark GB10, TP=2.
- Same production checkout, weights, DFlash2 revision, runtime patches, ablation, and image lineage in every arm.
- Diagnostic image `glm53-pr77-diag:e7e1286-v3`, derived from proven `glm53-fat-e2:validated-20260901` by replacing only `vllm/.../quantization/exl3.py` and its narrow E2 self-check.
- MNBT 2048; fused MoE on; row tiles off.
- Sequence: E2 on → all three tier flags off → E2 on.
[...]

PR #99, opened 2026-09-01
Stage-1 store-only disk KV tier: eligible-group scoping + worker-local per-rank store (GLM53_KV_OFFLOAD, default off)

nood-co1:
# Stage-1 store-only disk KV tier: eligible-group scoping + worker-local store with chunk headers (`GLM53_KV_OFFLOAD`)

**Status:** knobs default OFF (control boot byte-identical). Host-side evidence below is complete; the live boot receipts (section ‘Receipts’) will be posted in this thread at the next server window — a measurement run currently owns the pair.
window after the #77 measurement pass. Codex gates: advisory
CODEX-OFFLOAD1-PLAN.md (DO-NOT-PROCEED v1 → all findings adopted/rebutted in
writing before build), adversarial diff review CODEX-OFFLOAD1-REVIEW.md
(DO-NOT-SHIP → all findings adopted/rebutted, fixes in 32f993e, confirm pass
appended there).**

## Problem

A session whose KV has left the ~455K-token resident GPU envelope reprefills from
zero (174 s measured at 120K; 15–25 min estimated at 450K). The research plan
(PLAN-KV-OFFLOAD.md v2.1 + stage-0 receipts) designs a disk tier: park-to-disk on
store-cascade, restore on next touch. This PR is **stage 1 only: store-only,
restore disabled by flag (the launcher refuses `GLM53_KV_OFFLOAD_RESTORE=1`),
no restore/read path** — the tier writes; nothing reads it back yet. knob=0 is
byte-identical serving; knob=1 carries one process-wide side effect (the
allocator change described in the Launcher section below).

Three things stand between the image's shipped `kv_offload` stack (present —
stage-0 receipt O4/R14) and a safe store on this deployment:

1. **The connector cannot boot on the hybrid layout.** `build_offloading_config`
   asserts `tokens_per_block % tokens_per_hash == 0` for every KV-cache group;
   the KpoolTailSpec scratch group (block 4, `prefix_cacheable=False`) fails
   (4 % 64). Upstream PR vllm-project/vllm#54743 (ours) scopes the group configs
   to prefix-cacheable groups; this overlay ports that scoping onto the image's
   older tree.
2. **The image wraps group specs.** The connector receives
   `UniformTypeKVCacheSpecs` wrappers (subclass of `KVCacheSpec` only): the
   scheduler's isinstance dispatch would crash on group 0's full-attention
   assert and silently strip the mamba groups' alignment/CoW semantics. The
   port adds a fail-closed single-type unwrap and reads cacheability through
   the image's `participates_in_prefix_caching` name.
3. **The stock fs tier is silently wrong on a 2-node TP pair (plan risk R0).**
   `TieringOffloadingSpec.get_manager()` constructs every secondary tier
   scheduler-side against the scheduler node's staging memoryview; on
   `--nnodes 2` rank 1's staging mmap lives on the other Spark, so the fs tier
   would persist rank-0 bytes plus never-written garbage as the rank-1 half of
   every file — corruption that only surfaces at restore time. The store path
   must be **worker-local**: each rank writes its own bytes from its own
   staging mmap to its own NVMe.

### `overlay/patch_kv_offload_scope.py` — eligible-group scoping (port of vllm#54743)

- `OffloadingGroupConfig` gains `group_idx` (original boot index, required);
  `build_offloading_config` filters to prefix-cacheable groups minus the fork
  policy exclusion of draft-model attention groups (`is_eagle_group`, or exact
  `SlidingWindowSpec` under an EAGLE speculative config) unless
  `GLM53_KV_OFFLOAD_DRAFTER=1`; the divisibility assert still guards eligible
  groups.
- Scheduler port: `group_states` sized by the full group count; every pairing
  through `group_idx`; `GPULoadStoreSpec.group_sizes`/`block_indices` stay
  full-length with zeros for ineligible groups (worker layout unchanged);
  `supports_partial_tail` conservatively requires every group eligible.
- Final eligible set on this deployment: **{0, 2, 3, 4, 5}** — g1 out by the
  upstream predicate, g6 out by fork policy (min-exempt drafter; #83
  retention-0). Logged at boot.

### `overlay/patch_kv_offload_store_local.py` — worker-local store + §4 chunk headers

- Store jobs carry a versioned metadata side-channel (`TransferJob.glm53_store_meta`):
  ordered keys aligned 1:1 with the job's block order, plus boundary-manifest
  candidates (cumulative chunk-hash chains).
- Each worker SNAPSHOTS the job's bytes out of its own staging mmap
  synchronously at DMA completion (capture-then-write — `reset_cache()`
  frees staging rows regardless of pending disk work, so async mmap reads
  could tear; captured at `get_finished` before this rank's ack, or right
  after `worker.wait()` on the flush/reset path), then writes the private
  buffers (512 MB budget) to
  `GLM53_KV_OFFLOAD_DIR/glm53kv_<model>_<ns>_r<rank>/<hhh>/<hh>_g<g>/<hash>.bin`
  on its local NVMe — **real (unpadded) bytes only** (the slot-share padding
  never hits disk), fsync + tmp-rename + dir-fsync; acks are never deferred.
  Rank = the initialized TP rank cross-checked against the connector config;
  any disagreement disables the writer (never cross-rank mixing).
- **Every chunk gets a real header** (format v1): magic, namespace hash
  (fences model/vllm/build/TP/KV-dtype/hash-algo/`num_speculative_tokens`/
  per-group real page sizes/KDA state-ABI/format version), original group
  index, per-layer segment table (conv/temporal split with shape+dtype+stride
  for KDA layers), boundary token index, payload CRC, header CRC. Any mismatch
  at read ⇒ reject.
- **Boundary manifests** publish per-rank only after all four mamba payloads +
  every cumulative full-attention chunk are durable AND header-verified
  (namespace+hash+group identity; per-chunk `[hash, len, crc]` recorded for
  full groups too); a key that failed to write this boot can never be
  manifested (failed-key ledger). Dedup likewise requires a verifying
  header, else the file is rewritten.
- **Inline keep-K retention** (`GLM53_KV_OFFLOAD_KEEP_BOUNDARIES`, default 2):
  a manifest is superseded once beyond the K most recent boundaries of EVERY
  chain containing it (mamba payloads unlinked; full-attention chunks stay
  dense — cumulative references keep shallower boundaries restorable).
- Failure semantics (plan §8): failed write ⇒ unlink + log (lost store = future
  reprefill; nothing can serve wrong bytes); ENOSPC ⇒ writer pauses; crash
  between group writes ⇒ orphans, no live manifest; `overlay/kv_offload_store_gc.py`
  (dry-run default) verifies headers, reports orphans, sweeps on request.
- Restore stays cold: `get_num_new_matched_tokens` short-circuits to 0 external
  hits under `GLM53_KV_OFFLOAD_RESTORE=0` — no loads, no promotions, no
  WAITING_FOR_REMOTE_KVS.

### Launcher (`start.sh`)

Knobs validated fail-closed inside the pre-stop gate (bad knob or broken overlay
artifact exits 2 before `restart` stops anything); knob=0 adds **no** serve
argv, no mounts, no store dir — byte-identical serving. Knob=1 appends identical
`--kv-transfer-config` JSON to both ranks (`TieringOffloadingSpec`,
`cpu_bytes_to_use`, `blocks_per_chunk=1`, `offload_prompt_only=false` so
decoded tokens store too, **no secondary tiers** — see deviations), mounts
the store dir rw on both nodes, ships both overlays to both ranks, pins
their order (scope → store_local) in `GLM53_OVERLAY_ORDER`, and self-checks
both patchers' injected sources (`--check-injected`) inside the pre-stop
gate. The store namespace is fenced by the recipe checkout SHA
(`GLM53_KV_OFFLOAD_BUILD_ID`).

When `GLM53_KV_OFFLOAD=1`, the launcher also omits
`PYTORCH_CUDA_ALLOC_CONF=expandable_segments:True` from both containers
(`087b5ea`, found live in the receipt window): vLLM rejects OffloadingConnector
under expandable segments unless the CuMem allocator is enabled, which this
deployment does not use. knob=1 therefore runs **both processes on PyTorch's
default native caching allocator — a process-wide change, not scoped to the
store writer**. Expandable segments exists to reduce fragmentation OOM under
changing allocation sizes, so long-context workloads on knob=1 can in principle
OOM despite apparent free capacity where knob=0 would not (`GPU_MEM_UTIL`
headroom is not a fragmentation guarantee). The knob=0 path keeps the stock
entry and remains byte-identical. The +8.3% cold-prefill overhead in the
receipts thread is entangled with this allocator change; an isolation
experiment (allocator-only vs store+allocator arms) is protocolized for the
next server window.

## Deviations (recorded, with reasons)

- **No fs secondary tier in the connector JSON** (Codex advisory finding 3/Q1,
  adopted): the scheduler-side `FileSystemTierManager` byte path is exactly the
  R0 hazard; configuring-then-neutering it adds fake cascade bookkeeping and a
  stock-namespace config.json for no benefit. The worker-local writer IS the
  disk tier; the in-image fs tier remains available untouched for stage-2 read
  work.
- `blocks_per_chunk=1` uniformly (plan §4 wanted 16 for MLA): the image has one
  global chunk size; all eligible groups share block 3584. g0 chunk files carry
  27,163,136 B payloads (11 MLA + 11 indexer pages — the stage-0 correction
  pricing the indexer KV rides in automatically).
- CRC is zlib crc32, not crc32c (stdlib; the header names its algorithm).
- The worker→scheduler ack protocol has no per-job failure bit in stage 1
  (Codex finding 2, partially rebutted): a failed key is un-retried this boot
  but can never be manifested; the failure bit lands with stage 2's T4 work,
  where it becomes load-bearing.

## Tests (host-side, all green; suite 33 passed)

| file | covers |
|---|---|
| `tests/test_kv_offload_scope.py` | patcher preflight/idempotence/tamper/half-patch on pinned image fixtures (sha256-recorded byte-exact captures); eligibility predicate matrix; patched-runtime drive under a stubbed vllm: eligible {0,2,3,4,5}, original-index keys, full-length zero group_sizes, mamba alignment 3584, store-only lookup short-circuit, no-scratch-model equivalence, divisibility regression |
| `tests/test_kv_offload_store_local.py` | header codec round-trip + truncation at every region + CRC flips; store-job meta 1:1 alignment + cumulative manifests + TransferJob pickle; writer: capture-then-write snapshot isolation (staging mutated after capture ⇒ disk bytes unchanged), flush-path capture, real-bytes payloads, KDA segment split, header-verified dedup + manifest ordering, failed-key ledger, keep-K retention, ENOSPC pause (no partials/temps left), TP-rank disagreement disable; GC dry-run/sweep/keep-K incl. zero-manifest orphans, stale temps, forked-chain shared-ancestor protection |
| `tests/test_launcher_kv_offload.py` | knob guard matrix incl. the RESTORE=1 refusal; pre-stop gate with zero docker/ssh calls; knob=0 no-mount parity; knob=1 both-rank env/mount/scp/JSON-shape parity + synthetic one-rank mismatch; overlay order pin |

## Receipts (pending: server window)

To be captured in the post-#77 window; the PR stays unopened until every line
has a receipt — and thereafter until the maintainer-response hold lifts:

1. **knob=0 byte-identical behaviour**: boot with `GLM53_KV_OFFLOAD=0`; the
   `vllm serve` argv is byte-identical to the previous boot (no
   `--kv-transfer-config`), no connector init in the log, no store mounts;
   serving probes green.
2. **knob=1 boot receipts**: `[glm53-kv-offload-scope] eligible groups:
   [0, 2, 3, 4, 5] of 7` on both ranks; TieringOffloadingManager created with
   0 secondary tiers; `[glm53-kv-offload-store] writer up: rank 0/2 …` on deb8
   and `rank 1/2` on d9d5 with per-rank base dirs; namespace JSON on both
   Sparks' local NVMe; driver version pinned in the capture.
3. **Store-cascade evidence**: during a normal session, headered files appear
   under `…_r0/**_g{0,2,3,4,5}/` on deb8 AND `…_r1/**` on d9d5 with payload
   sizes 27,163,136 (g0) / 21,086,208 (g2,g3) / 18,743,296 (g4,g5); boundary
   manifests for every complete 3584 boundary within keep-K; per-rank byte
   totals within ±10% of the stage-0 §C model; **R0 rank-routing receipt**: GC
   verifier reopen + checksum pass on BOTH ranks' stores across a container
   restart (verifier-level, not a GPU restore — stage 1 has no restore).
4. **Store overhead on prefill within noise**: prefill/decode rates knob=1 vs
   knob=0 on the standard probe within the kit's measured noise band.
5. **Fail-closed negatives**: bad knob values + `GLM53_KV_OFFLOAD_RESTORE=1`
   refused pre-stop with zero docker/ssh calls (live run); GC dry-run flags a
   deliberately truncated chunk and a manifest-less orphan; ENOSPC simulation
   (small loopback fs) pauses the writer with serving unaffected.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -140,6 +140,17 @@ GLM53_SUPPRESS_STOPS_IN_REASONING=1
 # When a decode seq is running, do not mix a peer prefill into that step
 # (issue #6). skip = decode-only steps; N>0 = cap mixed prefill tokens; 0 = off.
 GLM53_MIXED_PREFILL_CHUNK=skip
+# Disk KV-offload store tier, stage 1 (store-only; overlays
+# patch_kv_offload_scope.py + patch_kv_offload_store_local.py). 1 = enable the
+# OffloadingConnector: CPU staging bounce + worker-local per-rank headered
+# chunk files on each Spark's NVMe. 0 (default) = OFF, byte-identical serving.
+# Restore stays 0 in stage 1 (the launcher refuses 1). Exactly 0 or 1 each.
+GLM53_KV_OFFLOAD=0
+# GLM53_KV_OFFLOAD_DIR=$HOME/glm53-kv-offload
+# GLM53_KV_OFFLOAD_CPU_GB=4
+# GLM53_KV_OFFLOAD_RESTORE=0
+# GLM53_KV_OFFLOAD_DRAFTER=0
+# GLM53_KV_OFFLOAD_KEEP_BOUNDARIES=2
 # EngineCore execute timeout (seconds). Stock 300 is short for a cold JIT.
 # VLLM_EXECUTE_MODEL_TIMEOUT_SECONDS=1800
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -446,6 +446,11 @@ COPY overlay/patch_scheduler_decode_floor.py /opt/glm53/patch_scheduler_decode_f
 COPY tests/test_scheduler_decode_floor.py /opt/glm53/test_scheduler_decode_floor.py
 COPY overlay/patch_hybrid_prefix_hit.py /opt/glm53/patch_hybrid_prefix_hit.py
 COPY tests/test_hybrid_prefix_hit.py /opt/glm53/test_hybrid_prefix_hit.py
+COPY overlay/patch_kv_offload_scope.py /opt/glm53/patch_kv_offload_scope.py
+COPY overlay/patch_kv_offload_store_local.py /opt/glm53/patch_kv_offload_store_local.py
+COPY overlay/kv_offload_store_gc.py /opt/glm53/kv_offload_store_gc.py
+COPY tests/test_kv_offload_scope.py /opt/glm53/test_kv_offload_scope.py
+COPY tests/test_kv_offload_store_local.py /opt/glm53/test_kv_offload_store_local.py
 COPY overlay/patch_xgrammar_termination.py /opt/glm53/patch_xgrammar_termination.py
 COPY tests/test_xgrammar_termination.py /opt/glm53/test_xgrammar_termination.py
 COPY overlay/patch_kpool_tail_slotmap.py /opt/glm53/patch_kpool_tail_slotmap.py
@@ -461,6 +466,14 @@ RUN python3 /opt/glm53/patch_glm5_drafter_group.py
 RUN python3 /opt/glm53/patch_suppress_stops_in_reasoning.py
 RUN python3 /opt/glm53/patch_scheduler_decode_floor.py
 RUN python3 /opt/glm53/patch_hybrid_prefix_hit.py
+# Order-dependent pair: scope rewrites the offloading connector group
+# pairing; store_local anchors on scope's scheduler output. Host tests run
+# against the REAL in-image tree first (preflight + patched-runtime drive),
+# then the patchers bake the overlays in.
+RUN GLM53_REQUIRE_TARGET=1 python3 /opt/glm53/test_kv_offload_scope.py
+RUN python3 /opt/glm53/patch_kv_offload_scope.py
+RUN python3 /opt/glm53/patch_kv_offload_store_local.py
+RUN python3 /opt/glm53/test_kv_offload_store_local.py
 RUN python3 /opt/glm53/patch_xgrammar_termination.py
 RUN python3 /opt/glm53/patch_kpool_tail_slotmap.py
 RUN python3 /opt/glm53/patch_ablit.py
--- a/README.md
+++ b/README.md
@@ -533,6 +533,12 @@ this Dockerfile instead. After CUDA compile, Python overlay edits
 | `tests/test_xgrammar_termination.py` | exact two-file patch, idempotence, cross-file fail-closed drift, termination/rollback/reset and post-reasoning draft behavior, launcher wiring |
 | `overlay/patch_kpool_tail_slotmap.py` | clamp KpoolTail one-block circular slot mapping; identity for other KV groups |
 | `tests/test_kpool_tail_slotmap.py` | circular addressing math, exact kernel patch, idempotence, fail-closed drift, launcher wiring |
+| `overlay/patch_kv_offload_scope.py` | offloading-connector eligible-group scoping (port of vllm#54743 onto this image): prefix-cacheable groups minus the drafter policy exclusion, original indices preserved, full-length worker layout; unblocks the connector on the hybrid layout (KpoolTail scratch crashed it at boot) |
+| `overlay/patch_kv_offload_store_local.py` | stage-1 store-only disk KV tier (`GLM53_KV_OFFLOAD=1`): worker-local per-rank chunk files with versioned headers + segment tables, boundary manifests, inline keep-K retention; restore hard-off (`GLM53_KV_OFFLOAD_RESTORE` refused at 1); see `docs/DESIGN-kv-offload-store-only.md` |
+| `overlay/kv_offload_store_gc.py` | store verifier/GC: header+CRC checks, orphan/corrupt reporting (dry-run default), manifest-mediated keep-K sweep |
+| `tests/test_kv_offload_scope.py` | patcher preflight/idempotence/drift on pinned image fixtures; eligibility predicate; patched-runtime drive under a stubbed vllm (eligible {0,2,3,4,5}, original-index keys, full-length group_sizes, store-only lookup short-circuit) |
+| `tests/test_kv_offload_store_local.py` | header codec + torn-file rejection, store-job meta alignment + pickle, writer real-bytes/segment-table/manifest-ordering/retention/ENOSPC/rank-guard/delayed-ack, GC tool |
+| `tests/test_launcher_kv_offload.py` | knob guard matrix, fail-closed pre-stop gate, knob=0 no-mount parity, knob=1 both-rank env/mount/scp/JSON parity, overlay order pin |
 | `overlay/ablit_runtime.py` | load-time o_proj transplant / projection (`ABLIT=1`); no-op when off |
 | `overlay/patch_ablit.py` | install the load_weights hook; bind-mounted and run on both ranks |
 | `ablit/` | direction vectors + `LAYER_MAP.json` from drowzeys' published recipe; `fetch_transplant.py` + `transplant/` for the donor o_proj byte-copy |
--- a/docs/DESIGN-kv-offload-store-only.md
+++ b/docs/DESIGN-kv-offload-store-only.md
@@ -1,0 +1,135 @@
+# Stage-1 store-only disk KV tier (`GLM53_KV_OFFLOAD`)
+
+Design record for `overlay/patch_kv_offload_scope.py` +
+`overlay/patch_kv_offload_store_local.py` + the `start.sh` wiring. Stage 1 of
+the disk KV-offload research plan: the tier **writes**; nothing reads it back
+(`GLM53_KV_OFFLOAD_RESTORE` is refused at 1 by the launcher). Restore plumbing
+is stage 2.
+
+## Why
+
+A session whose KV has left the ~455K-token resident GPU envelope reprefills
+from zero (174 s measured at 120K). The plan's endgame is
+park-to-disk/restore-on-touch; stage 1 lands the write side with zero blast
+radius: knob off ⇒ no connector, byte-identical serving.
+
+## The two boot blockers this image has
+
+1. **Scratch-group divisibility crash.** `build_offloading_config` builds an
+   offload group config for every KV-cache group and asserts
+   `tokens_per_block % tokens_per_hash == 0` — the KpoolTailSpec scratch group
+   (block 4, hash grid 64) crashes the connector at boot. Our upstream PR
+   vllm-project/vllm#54743 scopes the group list to prefix-cacheable groups;
+   the scope overlay ports that onto this image's older tree.
+2. **Wrapped group specs.** This image hands the connector
+   `UniformTypeKVCacheSpecs` wrappers (subclass of `KVCacheSpec` only), which
+   the connector's isinstance dispatch cannot classify: the full-attention
+   assert would crash on group 0, and mamba groups would silently lose their
+   alignment/CoW semantics. The scope overlay adds a fail-closed single-type
+   unwrap (`_glm53_kvo_inner_spec`), and reads cacheability through the
+   image's `participates_in_prefix_caching` name.
+
+Eligible set on this deployment: **{0, 2, 3, 4, 5}** — g1 (kpool scratch) out
+by the upstream predicate, g6 (DFlash2 drafter) out by fork policy
+(min-exempt, retention 0: its window is rebuilt by the remainder's prefill;
+`GLM53_KV_OFFLOAD_DRAFTER=1` re-includes it for experiments). All eligible
+groups sit on the 3584 grid, so `blocks_per_chunk=1` gives one 3584-token
+chunk per group per boundary.
+
+## Why the store is worker-local (risk R0)
+
+`TieringOffloadingSpec.get_manager()` constructs the CPU primary tier AND all
+secondary tiers on the scheduler and hands the fs tier the *scheduler node's*
+staging memoryview. On this 2-node TP=2 kit, rank 1's staging mmap lives on
+the other Spark: the scheduler-side view's rank-1 slots are never written,
+and the stock fs tier would persist rank-0 bytes + uninitialised garbage as
+rank-1 halves — corruption that only surfaces at restore. So:
+
+- the connector JSON configures **CPU staging only** (no `secondary_tiers`);
+- store jobs carry a metadata side-channel (`TransferJob.glm53_store_meta`):
+  ordered `(hash, group_idx, chunk_idx)` aligned with the job's block order,
+  plus boundary-manifest candidates (cumulative chunk-hash chains);
+- each worker, after its GPU→CPU DMA completes, writes **its own bytes** from
+  **its own mmap** to **its own NVMe** under
+  `GLM53_KV_OFFLOAD_DIR/glm53kv_<model>_<ns>_r<rank>/…`, and only then acks
+  the store job (staging stays pinned until every worker acks, so bytes
+  cannot be evicted under the writer);
+- rank is the initialized TP rank cross-checked against the connector
+  config's rank — disagreement disables the writer rather than risking
+  cross-rank mixing.
+
+## On-disk format (v1)
+
+`GLM53KV1` magic | u32 header_len | canonical-JSON header | u32 header-crc32
+| payload. Header: format_version, namespace_hash, ORIGINAL group_idx,
+spec_kind, layers, block_size_tokens, n_tokens_valid, boundary_token_index,
+tp_rank/tp_world, per-layer segment table (KDA layers: conv/temporal split
+with shape/dtype/derived-contiguous stride), state-ABI tag +
+num_speculative_tokens for mamba groups, payload_len, payload_crc32
+(zlib crc32; the header names the algorithm). Payload = **real (unpadded)
+bytes only** — the CanonicalKVCacheRef layout the DMA uses already carries
+unpadded page sizes, so the mamba slot-share padding never reaches disk.
+
+Namespace hash: sha256[:12] over model, vllm version, build id, TP/world, KV
+dtype, prefix-cache hash algo, tokens_per_hash, blocks_per_chunk,
+num_speculative_tokens (baked into the KDA conv width), per-eligible-group
+(original idx, block size, layers, real page sizes), the KDA state-ABI
+descriptor, format_version. Any change forks the directory. Retention K is
+recorded in the namespace file but does NOT fence the hash.
+
+**Boundary manifests**: per (chain, boundary), keyed by the boundary hash,
+published only after all four mamba payloads AND the cumulative full-attn
+chunks are durable on that rank (fsync file+dir, tmp+rename). No manifest ⇒
+the boundary does not exist for stage-2 lookup. Cross-rank rule (stage 2): a
+boundary is restore-eligible iff BOTH ranks hold a valid manifest.
+
+**Retention** (`GLM53_KV_OFFLOAD_KEEP_BOUNDARIES`, default 2): inline, per
+publish — a manifest is superseded (manifest deleted, its mamba payloads
+unlinked) once it is beyond the K most recent boundaries of every chain
+containing it; full-attention chunks stay dense (cheap, and cumulative
+references from any live manifest keep shallower boundaries restorable).
+Cross-boot leftovers: `overlay/kv_offload_store_gc.py` (dry-run default).
+
+## Failure semantics (everything degrades to reprefill)
+
+| failure | behaviour |
+|---|---|
+| write fails | tmp unlinked, one log line, key enters the failed ledger (never manifested this boot), job acked — lost store only |
+| ENOSPC | writer pauses (all further writes dropped, one log); serving unaffected |
+| crash between group writes | orphan payloads, no live manifest; GC reports/sweeps |
+| torn/truncated file | header+CRC verification rejects at read (stage 2 / GC) |
+| rank mismatch, wrong layout, blocks_per_chunk≠1 | writer disabled at init with a named reason; jobs flow normally |
+
+Known stage-1 limit (recorded): the worker→scheduler ack protocol has no
+per-job failure bit, so the CPU tier believes a failed key stored and will
+not retry it this boot. Safe while restore is off (an incomplete boundary is
+simply invisible); the failure bit lands with stage 2's T4 invalidation work.
+
+## Knobs
+
+| knob | default | meaning |
+|---|---|---|
+| `GLM53_KV_OFFLOAD` | 0 | 1 = enable the connector + store tier. 0 = byte-identical serving (no argv, no mounts) |
+| `GLM53_KV_OFFLOAD_DIR` | `~/glm53-kv-offload` | host dir on EACH Spark's local NVMe; mounted rw at `/data/glm53-kv-offload` on both ranks |
+| `GLM53_KV_OFFLOAD_CPU_GB` | 4 | CPU staging bounce (`cpu_bytes_to_use`); never a capacity tier |
+| `GLM53_KV_OFFLOAD_RESTORE` | 0 | stage-1 hard rule: 1 is refused by the launcher |
+| `GLM53_KV_OFFLOAD_DRAFTER` | 0 | 1 = include the drafter group (stage-4 experiment) |
+| `GLM53_KV_OFFLOAD_KEEP_BOUNDARIES` | 2 | inline manifest retention (0 = dense) |
+
+## Deviations from the research plan (recorded)
+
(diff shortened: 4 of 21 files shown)


nood-co1:
# Live receipts — stage-1 store-only tier (server window 2026-09-01/02)

Captured on the production pair (deb8 head / d9d5 worker) in the post-#77 server window. All blocks below are **condensed/redacted excerpts** of the raw transcripts (kept in our receipts mirror), not full verbatim pastes. One launcher defect was found live and fixed during the window (receipt 2a) — provenance per receipt:

- **Receipt 1 (knob=0 control)** ran from staged deploy set `deploy/dogfood-next4` @ `7d98567` (= production `deploy/dogfood-next3` 972370a + `pr/kv-offload-mla-restore` 0a5c33e, which stacks this branch; launcher hunks re-derived). The staged `.env` was checked byte-identical to production before any boot (`sha256sum` of both files → the same `6a692047…`, 2×).
- **Receipts 2b–4** ran from `next4` @ `032ee21` (= the same set + the 2a fix `087b5ea`). Note the staged set includes the stacked stage-2 branch, so the running build carries #100's overlays too (inert; its receipts are in that thread).
- The build ID the writer fences into the store namespace is visible in the container env (`GLM53_KV_OFFLOAD_BUILD_ID=032ee216bb7a` on the knob=1 boots).

## 1. knob=0 byte-identical serving — CAPTURED

Control boot, knobs off. The `vllm serve` argv is **byte-identical** to the previous production boot (58 NUL-split argv entries, sha256 `50a55ab48f8f…343ac2` on both captures):

```
argv lines: 58; sha256: 50a55ab48f8f09b20c3c8c2425f567381f8f8f37f2015f224016a593e1343ac2
  ok   argv byte-identical to previous production boot
  ok   no connector init in head log
  ok   no store MOUNT on head container
  ok   no store MOUNT on worker container
health=200
```

(The `GLM53_KV_OFFLOAD*` values are replayed into the container env at knob=0 with `GLM53_KV_OFFLOAD=0` — the PR's documented contract is "no serve argv, no mounts, no store dir", which holds.)

## 2a. Launcher defect found live — fixed in `087b5ea`

The first `GLM53_KV_OFFLOAD=1 ./start.sh restart` failed fast at vLLM config validation (excerpt from the boot log, mirrored as `bootB-fail.txt`):

```
pydantic_core._pydantic_core.ValidationError: 1 validation error for VllmConfig
  Value error, KV connector OffloadingConnector is incompatible with
  PYTORCH_CUDA_ALLOC_CONF=expandable_segments:True unless enable_cumem_allocator
  is also enabled. …
```

The launcher exported `PYTORCH_CUDA_ALLOC_CONF=expandable_segments:True` unconditionally to both ranks; the check lives in `VllmConfig` and is unreachable from the host-side stubbed tests. Fix (`087b5ea`, on this branch): knob=1 boots drop the env on both ranks (logged); knob=0 boots keep the stock entry byte-identically. **Review status, disclosed: `087b5ea` was committed and pushed during this window; launcher stub tests C3b/D1b were added for both directions, and no fresh adversarial diff review has been performed on it yet.** Its visible consequence is real: every knob=1 boot now changes the allocator for both containers, not merely the store writer — which entangles receipt 4 below.

## 2b. knob=1 boot receipts — CAPTURED

```
(EngineCore …)  [glm53-kv-offload-scope] eligible groups: [0, 2, 3, 4, 5] of 7 (drafter_included=False)   [head]
(Worker_TP0 …)  [glm53-kv-offload-scope] eligible groups: [0, 2, 3, 4, 5] of 7 (drafter_included=False)
  ok   worker: eligible groups [0, 2, 3, 4, 5] of 7   [d9d5 container log]
(EngineCore …)  Created TieringOffloadingManager with primary tier (lru, 79 blocks) and 0 secondary tier(s)
(Worker_TP0 …)  [glm53-kv-offload-store] writer up: rank 0/2 base=/data/glm53-kv-offload/glm53kv__root_…_r0 eligible=[0, 2, 3, 4, 5]
  ok   worker: writer up rank 1/2   [d9d5 container log, same line shape with _r1]
```

- kv-transfer-config on the argv (both ranks): `{"kv_connector":"OffloadingConnector","kv_role":"kv_both","kv_connector_extra_config":{"spec_name":"TieringOffloadingSpec","cpu_bytes_to_use":4294967296,"blocks_per_chunk":1,"offload_prompt_only":false}}`
- Namespace-fenced per-rank base dirs on each Spark's local NVMe (`…_r0` on deb8, `…_r1` on d9d5); namespace JSON at each root.
- Driver version (both ranks): `580.173.02`.
- Harness note: the first receipts pass reported `FAILS=3` — a bug in our receipt script (`grep -q` under `pipefail` SIGPIPE-kills `docker logs`), not in the deploy; the corrected re-run reports `FAILS=0`. Both transcripts kept.

## 3. Store-cascade evidence — CAPTURED

Two cold prompts (fresh salts; the probe generator is labeled "20k" and the served `prompt_tokens` measured ≈25.9–26.1K — all "~26K" figures below are measured prompt_tokens), then the file survey. On-disk chunk sizes = §C payload sizes + per-chunk header, byte-exact:

```
deb8 (rank 0):                       d9d5 (rank 1):
      9 27167874   (g0)                    9 27167874
      6 21090519   (g2,g3)                 8 21090519
      6 18747175   (g4,g5)                 6 18747175
      4 27167873                           4 27167873
      2 21090518                           2 18747174
      2 18747174                           1 27167872
      1 27167872
30 chunks + 4 manifests + ns.json    30 chunks + 4 manifests + ns.json
```

Byte accounting against the stage-0 §C model, shown rather than asserted: a ~26K prompt crosses 7 aligned boundaries → per chain, 7 cumulative g0 chunks + (keep-K=2) 2 kept boundaries × 4 KDA groups. Two chains ⇒ predicted 14×27,163,136 + 8×21,086,208 + 8×18,743,296 = **698,919,936 payload B** + 131,852 header B = 699,051,788 B vs **observed 699,051,778 B** per rank (10 files carry a 1-byte-smaller header). Chunk counts match the model exactly (14/8/8). Per-probe store traffic ≈ 350 MB per rank. Equal sizes and per-file CRCs do **not** prove cross-rank payload byte-identity (the ranks hold different TP shards by design).

GC dry-run, both ranks:

```
r0: SUMMARY files=30 bytes=699051778 manifests=4 chains=2 corrupt=0 orphans=0 … mode=DRY-RUN
r1: SUMMARY files=30 bytes=699051780 manifests=4 chains=2 corrupt=0 orphans=0 … mode=DRY-RUN
```

**R0 rank-routing receipt (across a container restart):** after 8 probe chains (~2.8 GB per rank) and a full restart of the pair, the GC verifier re-opened both ranks' stores with **full payload-CRC verification**:

```
rank0 (deb8): SUMMARY files=120 bytes=2796207103 manifests=16 chains=8 corrupt=0 orphans=0 superseded=0 incomplete_manifests=0 stale_temps=0 broken_manifests=0 mode=DRY-RUN
rank1 (d9d5): SUMMARY files=120 bytes=2796207113 manifests=16 chains=8 corrupt=0 orphans=0 superseded=0 incomplete_manifests=0 stale_temps=0 broken_manifests=0 mode=DRY-RUN
```

(verifier-level, not a GPU restore — stage 1 has no restore.)

## 4. Store overhead — **FAILED as written**, numbers below

The body's criterion is "within the kit's measured noise band" (≤0.3% CV). Measured (cold ~26K prefills, temp 0, `max_tokens=4`, idle server, background transfer paused; knob=1 extended to 6 reps, reps 4–6 quoted as the settled band):

- knob=0 control: **25.50 / 25.26 / 24.71 s**
- knob=1 settled: **27.90 / 27.31 / 26.53 s**

**+8.3% mean wall time (paired +7.4–9.4%) — not within noise.** No decode-rate comparison was captured in this window. Two effects are entangled by necessity: (a) the store writes themselves (~350 MB per rank per probe at capture-then-write), and (b) the 2a allocator change — the connector cannot run under `expandable_segments:True`, so every knob=1 boot also swaps the allocator. We did not attribute the split. If this overhead matters for a workload, the knob stays 0 (the default), where serving is byte-identical.

## 5. Fail-closed negatives — CAPTURED (live, pre-stop; healthy pair untouched)

```
head StartedAt before: 2026-09-01T17:24:40.20220316Z
GLM53_KV_OFFLOAD=2 ./start.sh restart                     → rc=2  "GLM53_KV_OFFLOAD must be exactly 0 or 1 (got: 2)"
GLM53_KV_OFFLOAD_RESTORE=1 ./start.sh restart             → rc=2  "GLM53_KV_OFFLOAD_RESTORE=1 requires GLM53_KV_OFFLOAD=1 …"
GLM53_KV_OFFLOAD=1 GLM53_KV_OFFLOAD_DIR=relative/path …   → rc=2  "GLM53_KV_OFFLOAD_DIR must be an absolute path …"
head StartedAt after:  2026-09-01T17:24:40.20220316Z   →  NOTHING-STOPPED
```

(rc lines and messages re-associated from the interleaved transcript; the live run proves rc=2 with the serving container untouched — the zero-docker/ssh-call property is proven by the stubbed launcher tests, not by this transcript. Negative 2 exercised the stacked deploy set's stage-2 guard; the stage-1-only launcher refuses `RESTORE=1` with its own message.)

GC negative (deliberate corruption, in-container since the store is root-owned): one real g3 chunk truncated to 1,000,000 B, plus one 4 KiB garbage `.bin` planted. Detected: the truncated chunk (`CORRUPT … payload length 995689 != 21086208`), the garbage file (`CORRUPT … bad magic`), and the affected manifest (`MANIFEST-MISSING-PAYLOAD`, `incomplete_manifests=1`); nothing deleted in dry-run. **Note:** the garbage file is classified `corrupt`, not `orphan` — the body's "manifest-less orphan" cell (a valid-headered chunk no manifest references) was **not** planted and stays pending with ENOSPC.

**Pending, with reasons:**
- ENOSPC simulation (small loopback fs): needs root on the head; no passwordless sudo on this kit. The ENOSPC pause path is covered by the host writer test.
- GC orphan-classification cell (valid-header, manifest-less chunk): not planted this window (needs a header-writing helper on the host); the corrupt/incomplete-manifest cells above are captured.

Store dirs were removed at window close (stage-1 rollback rule); the production deploy was restored and verified — deployed state at window end: `deploy/dogfood-next3` 972370a, knobs absent, serve argv sha equal to the pre-window capture, health 200, cold-probe times back at the control band (24.87–26.57 s).

nood-co1:
**Follow-up: the owed adversarial review of `087b5ea` (alloc-conf drop) is done.**

The receipts comment above disclosed that `087b5ea` (knob=1 drops `PYTORCH_CUDA_ALLOC_CONF=expandable_segments:True` on both ranks) was committed mid-window with launcher stub tests only and no fresh adversarial diff review. That review has now run — an adversarial pass over the diff plus the full launcher env-assembly, precedence, and test-harness context. Verdict: **ship-with-changes — no blocker/major findings; the runtime fix is confirmed correct on both ranks** (empty-array expansion is safe under `set -euo pipefail` on bash 3.2/5; knob=0 keeps the stock entry at the same value and position on both the head array path and the worker string path; the caller-over-`.env` knob resolution completes before `launch_cluster` on every real path; the image bakes no `PYTORCH_CUDA_ALLOC_CONF` ENV that an omitted `-e` would unmask).

Three minor findings. Findings 1–2 were adopted in `0431c66` (additive on this branch); finding 3 was applied separately to the PR body:

1. **Test harness:** the worker-side env parser accepted any `KEY=value` token in the ssh command; it now only counts tokens immediately following `-e`, so a drifted worker env assignment fails C3b/D1b instead of passing on a stray token.
2. **Comment/log accuracy:** vLLM rejects OffloadingConnector under `expandable_segments:True` *unless the CuMem allocator is enabled* (not used on this deployment). The in-code comment, launcher log line, and test comment now say so; the pydantic error quoted in the receipts comment already carried the full condition.
3. **PR body:** the "zero blast radius on serving" claim understated knob=1 — the allocator change is process-wide on both ranks, not scoped to the store writer. The body now says knob=0 is byte-identical / knob=1 carries the allocator change, and a Launcher-section paragraph discloses the drop, the fragmentation/OOM caveat of leaving expandable segments (headroom is not a fragmentation guarantee), and CuMem as the unchosen alternative.

`0431c66` is merged forward to `pr/kv-offload-mla-restore` (#100's branch) as `c5dd69d`. Host suites re-run after the changes — store branch: 33 pytest passed + 3 standalone runners rc=0; offload branch: 40 pytest passed + 3 standalone runners rc=0.

On the entangled **+8.3% receipt-4 result**: an isolation protocol is now written and adversarially reviewed — a three-arm boot design (control / allocator-dropped-without-store via a test-only scratch-branch hook / knob=1) with pre-declared bars. It is not a full factorial: the stock-allocator+store cell is unbootable on this deployment's non-CuMem path, so the design attributes *conditional* effects only (allocator effect with the store off; on-path store overhead under the dropped allocator — the background write path is priced separately since it is off the request path). It runs in the next server window; the outcome lands in this thread under the pre-declared bars, which include a not-reproduced outcome and a mixed/no-attribution outcome as well as the allocator-accounts and store-path-cost verdicts.

nood-co1:
## Receipt 4 update: the +8.3% store overhead did not reproduce under the pre-registered isolation protocol

Receipt 4 in this thread reported knob=1 cold prefill at ~26K tokens +8.3% vs control, failing the within-noise bar, with a caveat that it was entangled with the mandatory allocator change. We ran the pre-registered three-arm isolation protocol on the same pair (boot-isolated arms, execution order A→C→B, 8 cold reps per arm, fresh `cache_salt` namespace per rep, all reps `hitsdelta=0`, paired-median tolerance T=0.80 s declared in advance):

| Arm | Config | Cold prefill median (~26K, n=8) |
|---|---|---|
| A | knob=0, stock allocator (production baseline) | 25.55 s |
| B | knob=0, `expandable_segments` dropped (allocator alone) | 25.32 s |
| C | knob=1 (store active; allocator dropped as required) | 25.91 s |
| A′ | re-baseline after restore (n=3) | 25.24 s |

Median paired deltas (per-index, bootstrap 95% CI, 10,000 percentile resamples): C−A **+0.36 s** [+0.21, +1.51]; C−B **+0.49 s** [+0.31, +1.81]; B−A **−0.19 s** [−0.30, +0.07]. Individual C−A deltas reached +3.55 s (arm C's first two reps carried boot-settle tails of 29.24/27.03 s); settled medians (reps 3–8): A 25.63 s, C 25.95 s. A′ drift check passes (|25.24 − 25.55| = 0.31 < T). The 3-rep decode probe was within its band in all arms.

**Verdict per the protocol's first-match ladder: R0 — the earlier +8.3% did not reproduce.** This run does not establish whether the earlier excess was environmental, allocator-related, or store-related; it establishes that under quiescence and cold-guard gates the knob=1 on-path overhead sits within the pre-declared tolerance.

Execution exceptions, disclosed: arm B's first boot was voided before any reps (warm-up never settled — the intermittent boot-settle failure noted earlier in this thread); the single contingency reboot supplied B's eight valid reps. Store activity during arm C wrote 136 files / ~2.80 GB per rank; the driver's post-run empty-root gate FAILED (root-owned files survived a non-sudo clear) — roots were cleared with sudo afterwards and now verify empty on both ranks (0 files head, 0 files worker). Production checkout and health verified restored (`972370a`, health 200).

Full receipts: `receipts-overhead-iso/` (per-rep rows, argv shas, boot timestamps, quiescence/cold-guard checks). The body's Receipts section will be updated to reference this correction.

---

nood-co1:
Updated to current main (e25aa70) by merging main into the branch — merge commit 26ca3cd, plain push, no force-push; tested history preserved (83e4bd5…0431c66 incl. the disclosed 087b5ea alloc-conf fix remain ancestors). Conflicts: start.sh (7 hunks — kv-offload knob capture/config/bool-flag helper/validator block/-e lines vs main's indexer-workspace/spinwait equivalents, both sides kept; both rank-script apply blocks re-derived onto `emit_overlay_block`, never keep-both-merged), .env.example, README.md (overlay table: both row sets kept). Dockerfile merged clean and was verified to carry BOTH chains: #77's fat-kernel copy/patch/compile-assert lines AND this PR's kv-offload test-then-patch chain. Interaction fixes inside the merge commit: `GLM53_OVERLAY_ORDER` += `patch_spinwait.py` + `patch_indexer_workspace.py` (main's own inline position); `patch_exl3_fat_kernel.py` is deliberately NOT added to the runtime order — it is a build-time patch (usage: EXLLAMAV3_EXT SOURCE_DIR, applied in the Dockerfile against the exllamav3 source tree before compilation; running it argument-less at container boot would abort the rank scripts) — recorded as a comment on the order list so the exclusion is pinned, with the compiled kernel asserted at image build; the launcher kv-offload guard harness pre-sets the two new knobs. Re-run green: `bash -n`, tests/test_launcher_kv_offload.py, tests/test_kv_offload_scope.py, tests/test_kv_offload_store_local.py, full host suite 19/23 (4 known env fails identical on upstream/main). Host suites re-run green post-merge. No new live receipts were taken for this update: the pre-merge receipts apply to the kv-offload path, which the resolution leaves byte-untouched; the rest of the deployment this branch now sits on (E2 fat kernels, MNBT default 7168, #63 template, indexer-workspace/spinwait knobs) is main's own merged, receipted work, not something this PR measures.

plotarmordev:
### Review follow-up — `26ca3cddd56a`

This is a maintainer scope deferral, not an author failure. Your explicit store-only boundary, allocator caveats, and correction of the earlier overhead measurement are acknowledged.

We have not selected an end-to-end hybrid-cache disk-tier design. A store-only stage with restore disabled does not yet deliver the restore benefit that would justify adopting this integration by itself. The prior #100 outcome also needs to be accounted for before choosing a staged approach.

Please hold as-is; no rebase or new measurement is requested. The merge hold means we own the architecture/prioritization decision, not that you owe an unspecified test.

plotarmordev:
### Maintainer architecture decision — close the store-only stage, preserve its work

We are declining to adopt a store-only stage without a usable end-to-end hybrid restore path and are closing this PR on **architecture/scope grounds, not as a duplicate or a defect-driven rejection**. #58 is the experimental offload direction we will pursue; it is not qualified or approved to merge, and it does not replace the unique format/GC work here.

We reviewed #100's [closed-unmerged disposition](https://github.com/MiaAI-Lab/GLM-5.3-Flash-EXL3-2x-DGX-Sparks/pull/100#issuecomment-5610602175): its group-0 restore is inert on this model's seven-group hybrid layout because the mamba groups have no restore source and its load-failure integration requires a single total group. The later single-group probe reached disk-job creation but did not establish successful worker restore. Thus #100 is not an available restore stage that makes adopting #99 alone deliver the intended benefit.

Thank you @nood-co1 and Blockbrain Labs for the worker-local disk format, namespace/rank fencing, chunk headers and CRCs, durable boundary manifests, GC/verifier design, eligible-group scoping port, and unusually explicit receipts. The earlier +8.3% overhead did not reproduce under your predeclared isolation protocol; we are not reviving that result as a rejection reason or asking for the pending receipt cells.

The source branch `pr/kv-offload-store-only` is being left intact. Preserve [the exact reviewed source at 26ca3cddd56ab5ab271ff26951db4a96eb6cc211](https://github.com/MiaAI-Lab/GLM-5.3-Flash-EXL3-2x-DGX-Sparks/tree/26ca3cddd56ab5ab271ff26951db4a96eb6cc211), especially `docs/DESIGN-kv-offload-store-only.md` and `overlay/kv_offload_store_gc.py`, as design references for any future hybrid restore proposal. No rebase, new measurements, or other author work is requested. A future adoption/reopening decision belongs to maintainer @plotarmordev when an end-to-end hybrid restore direction is actually selected; this closure does not create a standing request to build one.

Issue #85, opened 2026-08-31
Concurrency seems broken


gl319-oss:
Hi,

Thanks for the great work.
Just a quick feedback though, for my setup (2 x ASUS Ascent GX10), concurrency seems totally dead. Even lowering context windows and such; it seems to only work for 1 stream.

Default recipe used, with .env untouched (only the needed variables changed for the dual sparks to work).

linxichen:
second this

MiaAI-Lab:
This was the `GLM53_MIXED_PREFILL_CHUNK=skip` deferral gate: while any request is decoding, waiting prefills are held until that decode finishes. That is now opt-in isolation, not the default.

`fair` v5 is the default on `start.sh` / `start-tp3.sh` / `start-tp4.sh` (PRs #186 / #188). It time-shares a bounded mixed-prefill chunk with the incumbent decode instead of waiting for the whole generation. On the reporter recipe (thinking-on essay + cold unique newcomer) a ~2k prompt got its first token in ~12 s on TP=2 / 8.7 s on TP=3 while the essay kept ~80–87% of its decode rate. Receipts: https://github.com/MiaAI-Lab/GLM-5.3-Flash-EXL3-2x-DGX-Sparks/blob/main/docs/diditfix.md

If `.env` still sets `skip`, drop it or set `fair` and restart. Reopen if a newcomer still waits out the running generation under `fair`.

"Only one stream works" on the untouched default recipe was `skip` holding every newcomer behind the active decode.

PR #112, opened 2026-09-02
feat: LONG_PREFILL_TOKEN_THRESHOLD — stop a long prefill starving every other session (440 s freeze -> 20 s)

### LINKED ISSUE #110: Long prefill starves every other session for its full duration (440 s freeze); --long-prefill-token-threshold 1024 fixes it, mixed-prefill gate is not the cause
## Summary

A long chunked prefill takes the **entire per-step token budget for its whole duration**, so every other session on the engine is frozen until it finishes. On a 2× GB10 kit at MNBT=2048, one 325k cold prefill froze a second, fully-warm 325k session for **440 seconds**.

`--long-prefill-token-threshold 1024` fixes it: the same session stays at **17.9–24.9 s**, with zero evictions. Cost is ~10% on the long prefill, and **nothing at all** when it runs alone.

This is, I believe, the mechanism behind #85 ("concurrency seems totally dead… it seems to only work for 1 stream"), and part of what #43, #6 and #7 are each seeing from different angles.

## Reproduction

Session A is warmed to ~325k tokens, then polled with a short follow-up every 25 s. Sixty seconds in, session B starts a **cold ~325k prefill** with different text (no shared prefix). Same image, same kit, only the one knob changes between arms.

| arm | A's follow-ups during B's prefill | evictions | B's cold prefill |
|---|---|---:|---:|
| `GLM53_MIXED_PREFILL_CHUNK=skip` (recipe default) | **froze 440.2 s** | 1 | 447.0 s |
| `GLM53_MIXED_PREFILL_CHUNK=0` (policy off) | **froze 448.9 s** | 1 | 451.9 s |
| **`--long-prefill-token-threshold 1024`** | **17.9, 24.3, 19.8, 24.9, 20.6, 21.2, 18.5, 21.1, 23.3, 19.2, 19.3 s** (median 19.8) | **0** | 494.2 s |

Arm 2 is the important control: **the mixed-prefill gate is not the cause.** Turning it off entirely changes nothing. That surprised me, because the gate is the obvious suspect.

## Mechanism

`vllm/v1/core/sched/scheduler.py`, both the running path (~line 636) and the waiting path (~line 1020):

```python
if 0 < self.scheduler_config.long_prefill_token_threshold < num_new_tokens:
    num_new_tokens = self.scheduler_config.long_prefill_token_threshold
```

Unset (`0`), a chunked prefill claims `min(remaining, token_budget)` every step — i.e. all of it — for as long as it runs. Everything else waits. Set, each request is capped per step and the remainder of the budget is left for other requests.

There is a second-order effect worth naming: **a request that waits also loses its cached blocks.** In our traces the waiting session logged a 322,560-token prefix hit repeatedly while it waited, then dropped to `computed=0` and had to restore from the offload store. So the symptom reads as "prefix caching is broken" when it is really "this request never got scheduled".

## Cost

| | without | with 1024 |
|---|---:|---:|
| long prefill, **contended** | 447.0 s | 494.2 s (+10.6%) |
| long prefill, **solo** | 424.9 / 432.4 s | 435.9 s (within noise) |

The cost is only paid under contention, which is exactly when you want it paid. It is a fairness change: instead of one session starving another, they split the budget.

## Suggestion

Consider a first-class knob (`LONG_PREFILL_TOKEN_THRESHOLD`, default `1024`) rather than leaving it to `EXTRA_ARGS`, which is itself undocumented (#89). Happy to open that PR if you want it in this shape.

Two notes for anyone reproducing:

- Aggregate throughput and **median** latency both look fine while one session is frozen solid. The signature is only visible in p99/max and "turns over 60 s".
- The MiaAI-Lab DeepSeek-V4-Flash DSpark recipe has shipped `--long-prefill-token-threshold 1024` for months; that is where we got the idea, after our GLM port went to production without it and stalled a real 552k agent session.

## Environment

2× DGX Spark (GB10), TP=2 over CX7 RoCEv2, image `ghcr.io/miaai-lab/glm-5.3-flash-2x-dgx-sparks:exl3`, `Mia-AiLab/GLM-5.3-Flash-EXL3-TR3-4bpw` @ `25a44fdb`, DFlash2 k=7, `DFLASH_DRAFT_TP=2`, fp8 KV, `MAX_MODEL_LEN=1000000`, `MAX_NUM_SEQS=4`, MNBT=2048, `GLM53_INDEXER_WORKSPACE=rightsize`, prefix caching on, vision loaded.

jdecker76:
Fixes #110.

## Problem

Unset, one chunked prefill claims the **whole per-step token budget for its entire duration**, so every other session on the engine is frozen until it finishes. This is, I believe, the mechanism behind #85 ("concurrency seems totally dead… it seems to only work for 1 stream") and part of what #43, #6 and #7 each see from a different angle.

## Measurement

2× GB10, TP=2, MNBT=2048. Session A warmed to ~325k then polled with a short follow-up every 25 s; 60 s in, session B starts a cold ~325k prefill with unrelated text. Only the one knob changes between arms.

| arm | A's follow-ups during B's prefill | evictions | B's cold prefill |
|---|---|---:|---:|
| `GLM53_MIXED_PREFILL_CHUNK=skip` (default) | **froze 440.2 s** | 1 | 447.0 s |
| `GLM53_MIXED_PREFILL_CHUNK=0` (policy off) | **froze 448.9 s** | 1 | 451.9 s |
| **`--long-prefill-token-threshold 1024`** | **17.9–24.9 s** (median 19.8) | **0** | 494.2 s |

Arm 2 is the control worth noting: **the mixed-prefill gate is not the cause.** Disabling it entirely changes nothing, which is why this went unfound for a while — the gate is the obvious suspect.

Second-order effect: a request that waits also **loses its cached blocks**. Our traces show the waiting session logging a 322,560-token prefix hit repeatedly, then dropping to `computed=0`. So the symptom presents as "prefix caching is broken" when it is really "this request was never scheduled".

## Cost

| | without | with 1024 |
|---|---:|---:|
| long prefill, **contended** | 447.0 s | 494.2 s (+10.6%) |
| long prefill, **solo** | 424.9 / 432.4 s | 435.9 s (within noise) |

Paid only under contention. It is a fairness change: sessions split the budget instead of one starving another.

## Change

- `LONG_PREFILL_TOKEN_THRESHOLD`, default **1024**; `0` restores stock behaviour exactly (the flag is simply not passed).
- Validated as a non-negative base-10 integer in `validate_numeric_config`, and rejected when it exceeds `MAX_NUM_BATCHED_TOKENS`, where it could never bind.
- Threaded to **both ranks** (serve args, worker env passthrough, `docker -e`), following the `MAX_NUM_BATCHED_TOKENS` pattern.
- Documented in `.env.example` with the measurements above.

`bash -n start.sh` clean.

## On the default

I set it to 1024 because the solo-prefill cost is within noise and the contended benefit is a 440 s freeze becoming ~20 s. **Happy to flip it to `0` (opt-in) if you would rather not change default behaviour** — say the word and I will push that.

Credit: the MiaAI-Lab DeepSeek-V4-Flash DSpark recipe has shipped `--long-prefill-token-threshold 1024` for months. That is where the idea came from, after our GLM port went to production without it and stalled a real 552k agent session.

## Not tested here

TP=4 (`start-tp4.sh`) — the plumbing follows the same pattern but I have no 4-Spark kit to verify on.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -103,6 +103,17 @@ MAX_NUM_SEQS=4
 # Prefill chunk. E2 one-shot keep on this 2× GB10 kit (2026-09-01): 7168.
 # 2048/3548 were similar or slower on 100k–300k. Never 8192 (indexer smem).
 MAX_NUM_BATCHED_TOKENS=7168
+# Per-step prefill cap (vLLM --long-prefill-token-threshold). 0 = stock.
+# Unset/0, ONE chunked prefill claims the whole token budget for its entire
+# duration, so a long cold prefill freezes every other session until it is done.
+# Measured on 2x GB10 TP=2, a warm 325k peer polled every 25 s while a second
+# session cold-prefilled 325k:
+#     stock (0)                     peer froze 440 s, lost its cached blocks
+#     GLM53_MIXED_PREFILL_CHUNK=0   peer froze 449 s  (the gate is NOT the cause)
+#     1024                          peer stayed 17.9-24.9 s, zero evictions
+# Cost is only paid under contention: the long prefill goes 447 -> 494 s (+10.6%);
+# a SOLO prefill is unchanged (424.9 -> 435.9 s). Must be <= MAX_NUM_BATCHED_TOKENS.
+LONG_PREFILL_TOKEN_THRESHOLD=1024
 GPU_MEM_UTIL=0.87
 KV_CACHE_DTYPE=fp8
 # 0 = load vision tower (image + video). 1 = language-only.
--- a/start.sh
+++ b/start.sh
@@ -182,6 +182,13 @@ MAX_NUM_SEQS="${MAX_NUM_SEQS:-4}"
 # 8192 chunk × long history oversubscribes GB10 persistent_topk smem (300k crash).
 # E2 one-shot 2026-09-01: 7168 keep (100k ~1148 / 300k ~1107); 2048/3548 similar or slower.
 MAX_NUM_BATCHED_TOKENS="${MAX_NUM_BATCHED_TOKENS:-7168}"
+# Per-step prefill cap. Unset/0 = stock: one chunked prefill claims the whole
+# token budget for its entire duration, so a long cold prefill freezes every
+# other session until it finishes (measured 440 s on a warm 325k peer). 1024
+# leaves budget for other requests: the same peer stayed at 17.9-24.9 s with
+# zero evictions, costing the long prefill ~10% (447 -> 494 s) and a SOLO
+# prefill nothing (424.9 -> 435.9 s, within noise). Set 0 to restore stock.
+LONG_PREFILL_TOKEN_THRESHOLD="${LONG_PREFILL_TOKEN_THRESHOLD:-1024}"
 CHAT_TEMPLATE_HOST="${CHAT_TEMPLATE_HOST:-$SCRIPT_DIR/files/chat_template.jinja}"
 CHAT_TEMPLATE="${CHAT_TEMPLATE:-/opt/glm53/chat_template.jinja}"
 VIDEO_PATCH_HOST="${VIDEO_PATCH_HOST:-$SCRIPT_DIR/overlay/patch_glm_video_placeholders.py}"
@@ -361,6 +368,15 @@ validate_numeric_config() {
     _glm53_canonical_positive_int MAX_MODEL_LEN "$MAX_MODEL_LEN" 1000000 || return
     _glm53_canonical_positive_int MAX_NUM_SEQS "$MAX_NUM_SEQS" 4096 || return
     _glm53_canonical_positive_int MAX_NUM_BATCHED_TOKENS "$MAX_NUM_BATCHED_TOKENS" 8388608 || return
+    case "${LONG_PREFILL_TOKEN_THRESHOLD-0}" in
+        ''|*[!0-9]*)
+            echo "LONG_PREFILL_TOKEN_THRESHOLD must be a base-10 integer >= 0 (0 = stock) (got: ${LONG_PREFILL_TOKEN_THRESHOLD-})" >&2
+            return 2 ;;
+    esac
+    if [ "${LONG_PREFILL_TOKEN_THRESHOLD:-0}" -gt "${MAX_NUM_BATCHED_TOKENS}" ] 2>/dev/null; then
+        echo "LONG_PREFILL_TOKEN_THRESHOLD (${LONG_PREFILL_TOKEN_THRESHOLD}) exceeds MAX_NUM_BATCHED_TOKENS (${MAX_NUM_BATCHED_TOKENS}); it would never bind" >&2
+        return 2
+    fi
     _glm53_validate_enum GLM53_INDEXER_WORKSPACE "${GLM53_INDEXER_WORKSPACE-stock}" \
         stock rightsize || return
     _glm53_validate_spinwait_ms || return
@@ -953,6 +969,7 @@ ARGS=(
 [ -n "${GPU_MEM_UTIL:-}" ]  && ARGS+=(--gpu-memory-utilization "${GPU_MEM_UTIL}")
 [ -n "${MAX_NUM_SEQS:-}" ] && ARGS+=(--max-num-seqs "${MAX_NUM_SEQS}")
 [ -n "${MAX_NUM_BATCHED_TOKENS:-}" ] && ARGS+=(--max-num-batched-tokens "${MAX_NUM_BATCHED_TOKENS}")
+[ "${LONG_PREFILL_TOKEN_THRESHOLD:-0}" != "0" ] && ARGS+=(--long-prefill-token-threshold "${LONG_PREFILL_TOKEN_THRESHOLD}")
 [ -n "${KV_CACHE_DTYPE:-}" ] && ARGS+=(--kv-cache-dtype "${KV_CACHE_DTYPE}")
 if [ "${SPEC_METHOD:-mtp}" = "dflash" ]; then
     ARGS+=(--speculative-config "$(python3 -S -c 'import json,os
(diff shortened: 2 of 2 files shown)


plotarmordev:
thank you for the PR, will look into it asap, currently running tests and pr validations both on dsv4f and glm only on 2 sparks

danielbanu:
## Same-kit data: TP=2 / MAX_NUM_SEQS=4 / MNBT=7168 (fat kernel + indexer rightsize)

2x DGX Spark, DFlash2 k=7. Harness: session A warms a ~100K-token prefix, then polls it every ~25 s; 60 s in, session B starts a fully-uncached ~100K prefill with unique text (no shared prefix between sessions). Only the knob under test changes between boots. Four boots, one arm each.

| Boot | Knob | A max poll freeze | B cold prefill |
|---|---|---:|---:|
| 1 | unset (stock) | **150.2 s** | 102.9 s (973 tok/s) |
| 2 | `--long-prefill-token-threshold 1024` (PR default) | 6.9 s | 182.3 s (547 tok/s, **-44%**) |
| 3 | `MAX_NUM_BATCHED_TOKENS=2048`, threshold unset | 145.1 s | 98.5 s (1017 tok/s) |
| 4 | MNBT 7168 + `--long-prefill-token-threshold 3584` | **10.7 s** | **95.2 s (1046 tok/s)** |

Findings on this geometry:

1. **The PR default of 1024 over-taxes MNBT=7168 kits.** B keeps only 1024/7168 = 14% of each step, and our per-step fixed overhead (~0.95 s measured across arms) dominates the wasted budget -> -44% contended prefill. Your +10.6% contended cost was measured at MNBT 2048, where 1024 keeps 50% of the step - the default-to-threshold ratio is what matters.
2. **MNBT/2 is the sweet spot here.** Threshold 3584 (= MNBT/2): freeze collapses 150 s -> 11 s (14x) *with no prefill tax* - B actually measured fastest of all four arms. A half-step claim still guarantees interleaving every other step.
3. **Shrinking MNBT is not an alternative.** Boot 3 confirms the starvation is scheduling-rights, not chunk size: MNBT 2048 left A frozen 145 s while B prefill throughput stayed equal to MNBT 7168 (fat-kernel era).

Suggestion: consider documenting `threshold = MNBT/2` as a starting point for MNBT>4096 kits, or deriving the default as a fraction of MNBT instead of the fixed 1024. Happy to provide raw harness data. Thanks for the fix and the falsification of the mixed-prefill-gate theory - boot 2 vs boot 1 reproduces your arm-2 control exactly (disabling the gate changes nothing).

danielbanu:
Addendum: held threshold = MNBT/2 at MNBT 4096 (th 2048), same harness.

| Boot | Knob | A max poll freeze | B cold prefill |
|---|---|---:|---:|
| 4 | MNBT 7168 + th 3584 | 10.7 s | 95.2 s (1046 tok/s) |
| 5 | MNBT 4096 + th 2048 | **4.5 s** | 104.0 s (957 tok/s, −9%) |

The ratio scaling holds: smaller th → finer interleaving (max gap per step ~half), contended long prefill yields more of each step. Both configs collapse the 150 s stock freeze. We're standardizing on MNBT 7168 + th 3584 (maintainer-default MNBT, best contended throughput); 4096/2048 is the pick if peer-latency smoothing matters more than long-prefill completion time.

hushengkai:
Confirming on **TP=4 / nnodes=4** (4× DGX Spark GB10, `start-tp4.sh`, MNBT=2048, MAX_NUM_SEQS=16, DFlash2 k=7 / draft_tp=4, single switched 200G QSFP CX7 fabric).

Since the PR wasn't in my pulled image yet, I set the stock flag directly via `EXTRA_ARGS="--long-prefill-token-threshold 1024"` — the underlying vLLM flag works as expected on TP=4, so the plumbing in this PR (same pattern as MAX_NUM_BATCHED_TOKENS) should be safe.

4 concurrent **cold** ~60k-token prefill streams (unique salt per run so APC can't serve across runs), `max_tokens=256`:

| metric | value |
|---|---|
| TTFT (4 streams) | **165.4 / 165.9 / 169.5 / 178.8 s** — fair spread, no session starved |
| aggregate cold contended prefill | **~1,450 tok/s** (240k tokens / ~165 s) |
| aggregate decode during the 4-way phase | ~110 tok/s (filler prompts, high DFlash2 acceptance; prose will be lower) |
| cache-hit re-prefill (same prompts re-sent) | TTFT 31.6–34.8 s |

Pre-flag, two concurrent ~200k sessions on this kit collapsed to **7–17 tok/s aggregate** with sessions frozen for minutes at a time — the #110/#85 shape. With the flag, the dead window disappears; what remains is the documented MNBT=2048 trade-off (coarser decode interleave during another session's prefill bursts). Note 1024 = MNBT/2 here, matching @danielbanu's ratio finding in this thread.

One caveat so the numbers aren't misattributed: two runtime wedges inside `apply_exl3_fused_moe` (GPU busy-spin, counters frozen, self-recovery after minutes) occurred during this stress window — that's a separate engine bug I'm filing as its own issue with py-spy evidence; the scheduler-fairness numbers above are unaffected by it.

plotarmordev:
**Status:** mechanism confirmed against the image's scheduler — rework requested on the launcher default, plus a rebase. We are running a flag-only A/B/A tonight and will post numbers here.

**Verified:** read the long-prefill-token-threshold path in the published image's vLLM scheduler: the threshold clamps chunked-prefill budgeting in **both** the running loop and the admission loop, so the flag genuinely bounds how much of each step one long prefill can claim — the starvation fix is real, not a logging artifact. The launcher wiring (empty = stock, forwarded via `EXTRA_ARGS` to both ranks) is correct. We are also referencing rather than repeating the same-kit evidence already in this thread: @danielbanu's 4-boot table and MNBT/2 addendum (TP=2, seqs=4, MNBT 7168) and @hushengkai's TP=4 confirmation — both land on threshold = MNBT/2 as the sweet spot.

**Change requests:**

1. **The launcher default of 1024 is wrong for MNBT=7168 kits — per your own thread's addendum data.** The threshold-to-MNBT ratio is what matters: 1024/7168 leaves a contended long prefill only 14% of each step and measured **−44%** contended prefill throughput on this geometry, while 3584 (= MNBT/2) collapsed the 150 s freeze to ~11 s with *no* prefill tax. Please either derive the default as MNBT/2 or ship it empty (stock, flag absent) and document the `threshold = MNBT/2` starting point for MNBT>4096 kits. A fixed 1024 default actively hurts the deployment shape this repo targets.
2. **Rebase onto current main** — the PR is CONFLICTING in `start.sh` (main has since landed the E3 knobs, the adaptive-k / dense-fp8 overlays, and the 850K / 0.85 defaults in the same regions).

**Waiting on:** our flag-only A/B/A tonight on the 2x GB10 TP=2 E3 kit — no merge needed, `EXTRA_ARGS="--long-prefill-token-threshold 3584"` vs stock across fresh boots: a warmed ~100K session polled every ~25 s, with a fully-uncached ~100K prefill fired 60 s in; measuring max poll freeze and contended prefill wall against a stock control boot. Same-kit expectation per the thread data: freeze ~150 s → ~11 s, prefill ~0% tax. Numbers will be posted here; if they reproduce, this plus change (1) is a merge.

plotarmordev:
Merged via https://github.com/MiaAI-Lab/GLM-5.3-Flash-EXL3-2x-DGX-Sparks/pull/157 (rebased onto current main with maintainer fixes; your commits and authorship are preserved). Thank you for the contribution — closing this one in favour of the merged replacement.

PR #86, opened 2026-08-31
Indexer prefill-workspace right-sizing: reclaim ~4.5 GiB for KV (+27% capacity, opt-in)

nood-co1:
### Problem
`get_max_prefill_buffer_size` sizes the sparse-MLA indexer's prefill workspace at `max_model_len × 40` entries. At 1M context
that is 40,000,000 × 132 B + 1 MiB = **5,281,048,576 B = 5036.40 MiB**, requested during the memory profile and locked —
reconciled to the byte with the live boot receipt (`[WORKSPACE DEBUG] Resized workspace … 0.00 MB -> 5036.40 MB`,
`sparse_attn_indexer_kpool.py:284`). The legal per-step need is orders of magnitude smaller. (The related `deepseek_v4`
call site already divides by `compress_ratio`; the `glm5next` one does not — that asymmetry is the cleaner upstream fix.)

### Change
Overlay `patch_indexer_workspace.py` (fail-closed, transactional, opt-in): `rightsize` sizes the workspace at the **legal
per-step maximum** `min(max_num_seqs, MNBT) × cdiv(max_model_len + k_spec, index_kpool)`, clamped to stock so it can only
narrow; chunking behaviour proven byte-identical to stock over 400 randomized legal batches (+ a test-power test). An
unconditional cross-check of `hf_text_config.index_kpool` vs `kv_cache_spec.compress_ratio` raises at init on any mismatch,
both directions. Strict enum parsing parity-tested against the launcher's validator; setness-aware caller capture. Default
`stock` = byte-identical behaviour. 21-check host test; Dockerfile test-before-patch.

### Expected gain (estimates until the allocator receipt below)
Reclaims ~4.4–4.8 GiB of the 17 GiB KV budget → **~+26–28 % KV capacity** at `MAX_NUM_SEQS` 16/4. No latency claims.

### Live receipts (2026-09-01, both boots on the same deployed config)
| | stock | rightsize |
|---|---|---|
| workspace (boot log) | 40,000,000 entries / **5036.40 MB** | **4,000,032 entries** (`legal max 4000032`; ~4531.9 MiB reclaimed) |
| Available KV cache memory | 16.98 GiB | **21.5 GiB (+26.6 %)** |
| GPU KV cache size / max concurrency | 1,555,555 tokens / 1.56× | **1,929,951 tokens / 1.93×** |
| prefill smoke 8K / 50K / 100K | — | 9.7 s / 57.2 s / 115.1 s, clean outputs |
[...]

PR #115, opened 2026-09-03
docs(tp4): dated 4-Spark measurements and corrected stall attribution

simossss:
`start-tp4.sh` is marked "untested here (no 4-Spark kit)". We ran this recipe's image, overlay and weights on four DGX Sparks (two at 200G, two at 100G, CRS812 switch), DFlash2 draft TP=4, 1M context, and tuned the serve knobs with an autoresearch loop: one knob per relaunch, a ten-minute production-mix benchmark (coding + 24k tool-call mix, long generation, cold 64k prompt, 4/8-stream ladder, cancellation), a geometric-mean score against the baseline, and hard gates (contract checks, cancel drain, needle, zero errors, memory floor, no foreign traffic, GPU clocks). 24 experiments, replicates on the winner, then a 30-minute soak (195 requests, 0 errors).

This PR only adds documentation: the measured knobs and the reason for each to `.env.tp4.example` (sourced after `.env`, so nothing changes for the 2-node path), and one paragraph under "Experimental: 4× Spark". It does not touch `start-tp4.sh`.

What moved on TP=4 and why:

| knob | 2-node default | measured on 4 Sparks |
|---|---|---|
| `GPU_MEM_UTIL` | 0.87 | **0.75**: 0.85 left <1-2 GiB host memory per rank and preceded two engine deaths (NVRM `NV_ERR_NO_MEMORY`); 0.75 keeps ~9 GiB free, KV pool >3M tokens, no throughput cost |
| `MAX_NUM_SEQS` | 4 | **8**: 135 vs 84 tok/s aggregate at 8 streams, worst first token 1.0 s vs 27 s; 16 adds nothing (131) |
| `DFLASH_TOKENS` | 7 | **3**: 53 % accepted vs 30 %; prose decode 31-33 vs 26-28 tok/s; 2 and 5 worse; no speculation 22.6 |
| `MAX_NUM_BATCHED_TOKENS` | 7168 | **2048**: cold prefill on TP=4 is fabric-bound (1024-8192 all ~1.15k tok/s at 282k on 493cb88; +2-7 % at 4096/7168 with the E2 kernel), while 4096/7168/8192 push the first token of concurrent requests to 6-20 s |
| `GLM53_MIXED_PREFILL_CHUNK` | skip | **off**: with skip a short question waited 330 s behind a 12k answer; off answers it in 0.5 s and the running decode keeps ~90 % of its speed |

Also measured, in case useful for the 2-node docs: PR77's E2 kernel gives +16-18 % uncached prefill on TP=4 too (1156 → 1361 tok/s at 64k, 1162 → 1324 at 282k), replicated; `vm.swappiness=0` produced one engine death in two runs here and no gain, so we kept 60; TP=4 vs TP=2 on the same bench is 1.45x decode but only +29 % on 282k prefill.

Full recipe (node-local launcher, controller, watchdog, benchmark, autoresearch loop) and every receipt: https://github.com/punkjazz-labs/glm-5.3-flash-exl3-4x-dgx-spark

https://claude.ai/code/session_01Um6EfS81CU5BXMxYw94pkL

## Maintainer update 2026-09-15 — dated observations, corrected causality

Reviewed head `1ebad7129c77a08481d025fe7612d8ade2a3339a` updates the README and TP4-example caveat to the author’s [September 9 correction](https://github.com/MiaAI-Lab/GLM-5.3-Flash-EXL3-2x-DGX-Sparks/pull/115#issuecomment-5599489540): disabling DFlash did not eliminate the stalls, and the subsequent packet-loss-only explanation was withdrawn after stalls reproduced with flat monitored RoCE counters. On the reported kit, an H16 sparse-attention progress failure was localized; the internal race remains unidentified. The bounded final sparse-attention call (at most 64 query rows) is the workaround reported there, not a universal root-cause claim or a fix supplied by these env knobs.

The dated knob measurements and soak observations remain author-reported historical receipts. The five measured assignment lines and their dated explanatory block are byte-identical to the original PR head, including `GLM53_MIXED_PREFILL_CHUNK=off`; the correction adds no runtime/default change relative to that head. No new soak, benchmark, GPU run or runtime change was performed. These measurements are not fresh qualification of current main or another topology.

### DIFF
--- a/.env.tp4.example
+++ b/.env.tp4.example
@@ -55,3 +55,29 @@ CONTAINER_HEAD=glm53-exl3-tp4-head
 CONTAINER_WORKER=glm53-exl3-tp4-w1
 CONTAINER_WORKER2=glm53-exl3-tp4-w2
 CONTAINER_WORKER3=glm53-exl3-tp4-w3
+
+# --- serve knobs measured on a 4-Spark kit (2026-09-02) ----------------------
+# Validated on four DGX Sparks (2x 200G + 2x 100G to a CRS812 switch) with this
+# image/overlay (493cb88, then c190db1) and weights, DFlash2 draft TP=4, 1M context: 24
+# one-knob experiments, ten-minute production-mix bench after every relaunch,
+# then a 50-minute soak on the winner (188 requests, 0 errors). Receipts:
+# https://github.com/punkjazz-labs/glm-5.3-flash-exl3-4x-dgx-spark
+#   GPU_MEM_UTIL   0.85 left <1-2 GiB host memory per rank on TP=4 and preceded
+#                  two engine deaths (NVRM NV_ERR_NO_MEMORY); 0.75 keeps ~9 GiB
+#                  free, KV pool still >3M tokens, no throughput cost.
+#   MAX_NUM_SEQS   4 queues everything above four streams (84 tok/s aggregate at
+#                  8 streams, 27 s worst first token); 8 gives 135 tok/s and
+#                  1.0 s. 16 adds nothing on TP=4 (131 tok/s).
+#   DFLASH_TOKENS  3 accepts 53 % vs 30 % at 7: prose decode 31-33 vs 26-28 tok/s
+#                  single stream. 2 and 5 are worse than 3; no speculation 22.6.
+#   MAX_NUM_BATCHED_TOKENS  1024/2048/4096/8192: cold prefill on TP=4 is the same
+#                  (fabric-bound, ~1.15k tok/s at 282k), 8192 makes the first
+#                  token of a coding request 20 s under a 24k prefill. Keep 2048.
+#   GLM53_MIXED_PREFILL_CHUNK  skip makes a short question wait for the whole
+#                  running generation (330 s behind a 12k answer); off answers it
+#                  in 0.5 s and the running decode keeps ~90 % of its speed.
+GPU_MEM_UTIL=0.75
+MAX_NUM_SEQS=8
+DFLASH_TOKENS=3
+MAX_NUM_BATCHED_TOKENS=2048
+GLM53_MIXED_PREFILL_CHUNK=off
--- a/.env.tp4.example
+++ b/.env.tp4.example
@@ -76,6 +76,15 @@ CONTAINER_WORKER3=glm53-exl3-tp4-w3
 #   GLM53_MIXED_PREFILL_CHUNK  skip makes a short question wait for the whole
 #                  running generation (330 s behind a 12k answer); off answers it
 #                  in 0.5 s and the running decode keeps ~90 % of its speed.
+#   SPEC_METHOD / DFlash2 on TP=4 (2026-09-03): a 150-min soak at the 8-seq cap with
+#                  96k prompts in the mix hung the engine 3/3 times (31-78 min) with the
+#                  draft on: the 96k chunked prefill shares steps with 6-7 draft-verified
+#                  decode streams, all four ranks stop at the same forward, GPUs spin at
+#                  96 % / 21 W, vLLM dies 5 min later on "RPC call to sample_tokens timed
+#                  out". Fat-expert kernel on/off made no difference; the same soak with
+#                  the draft off passed 150 min (472 req, 0 err). Until fixed: run the
+#                  draft off on TP=4 (decode 22.6 tok/s), or keep prompts under ~64k, or
+#                  supervise the engine (a watchdog relaunch takes ~10 min).
 GPU_MEM_UTIL=0.75
 MAX_NUM_SEQS=8
 DFLASH_TOKENS=3
--- a/README.md
+++ b/README.md
@@ -439,7 +439,12 @@ at 100G, CRS812 switch), this image and overlay at 493cb88, DFlash2 draft TP=4,
 TP=4 vs the 2-node baseline on the same production-mix bench (temperature 0,
 30-min soak): decode 1.45x, mixed phases 20-35 % shorter, **cold prefill only
 +29 %** at 282k tokens (1162 vs 901 tok/s: a 4-node TP job is fabric-bound on
-prefill). The defaults above are tuned for 2 nodes; on 4 nodes an autoresearch
+prefill). One caveat found by a 150-minute soak on 2026-09-03: with the DFlash2
+draft on, a 96k chunked prefill sharing steps with 6-7 speculative decode
+streams hangs all four ranks (3/3 runs, 31-78 min; independent of the E2
+fat-expert kernel), while the same soak with the draft off passed clean, so the
+4-node production config runs without speculation until that is fixed (details
+and receipts in the linked repo). The defaults above are tuned for 2 nodes; on 4 nodes an autoresearch
 loop (one knob per relaunch, hard reliability gates) settled on the values now
 in `.env.tp4.example`: `GPU_MEM_UTIL=0.75` (0.85 left <2 GiB host memory per
 rank and preceded two engine deaths), `MAX_NUM_SEQS=8` (135 vs 84 tok/s
--- a/README.md
+++ b/README.md
@@ -433,6 +433,23 @@ and weights, does not change the supported 2× path. First run copies
 
 Do not pull `glm53-flash-sm121:v8` — that is the older NVFP4/Ray kernel.
 
+**Measured on a 4-Spark kit (2026-09-02).** Four DGX Sparks (two at 200G, two
+at 100G, CRS812 switch), this image and overlay at 493cb88, DFlash2 draft TP=4,
+1M context, launched per rank with the same `docker run` shape as `start.sh`.
+TP=4 vs the 2-node baseline on the same production-mix bench (temperature 0,
+30-min soak): decode 1.45x, mixed phases 20-35 % shorter, **cold prefill only
++29 %** at 282k tokens (1162 vs 901 tok/s: a 4-node TP job is fabric-bound on
+prefill). The defaults above are tuned for 2 nodes; on 4 nodes an autoresearch
+loop (one knob per relaunch, hard reliability gates) settled on the values now
+in `.env.tp4.example`: `GPU_MEM_UTIL=0.75` (0.85 left <2 GiB host memory per
+rank and preceded two engine deaths), `MAX_NUM_SEQS=8` (135 vs 84 tok/s
+aggregate at 8 streams, worst first token 1.0 s vs 27 s), `DFLASH_TOKENS=3`
+(53 % accepted vs 30 % at 7; prose decode 31-33 vs 26-28 tok/s),
+`MAX_NUM_BATCHED_TOKENS=2048` (larger chunks do not prefill faster on TP=4 and
+hurt first-token latency), `GLM53_MIXED_PREFILL_CHUNK=off`. Full recipe,
+launcher, watchdog, benchmark and every receipt:
+<https://github.com/punkjazz-labs/glm-5.3-flash-exl3-4x-dgx-spark>.
+
 API: `http://127.0.0.1:8888/v1` (LAN: `http://10.0.0.1:8888/v1`).
 `/v1` is unauthenticated unless you set `VLLM_API_KEY` in `.env` (opt-in;
 empty = no auth). vLLM reads the env var natively so the key never lands in


simossss:
Update after a longer soak on the 4-Spark kit (2026-09-03): a 150-minute run at the 8-sequence cap with 96k prompts in the mix hung the engine 3 out of 3 times (31-78 min). Each time a 96k chunked prefill was sharing steps with 6-7 DFlash2-verified decode streams; all four TP ranks stop at the same forward pass, GPUs spin at 96 % / ~21 W, and vLLM dies five minutes later on `RPC call to sample_tokens timed out`. No NVRM/Xid/dmesg lines. The E2 fat-expert kernel on or off made no difference; the identical soak with `SPEC_METHOD=none` passed 150 minutes (472 requests, 0 errors). I added the caveat to `.env.tp4.example` and the README paragraph in this PR, and the 4-node production config now runs without the draft until the cause is found. Full logs (scheduler dumps at the death, per-rank worker logs, watchdog) are in https://github.com/punkjazz-labs/glm-5.3-flash-exl3-4x-dgx-spark under `evidence/soak-crash-20260903T*` and `evidence/soak-spec0-20260903`. Happy to run a targeted repro if it helps narrow the DFlash2 verify path.

https://claude.ai/code/session_01Um6EfS81CU5BXMxYw94pkL

simossss:
Correction to my comment above: the draft is not the cause. An hour after the clean draft-off soak, the same draft-off quartet stalled during a 282k cold prefill that was running **alone** (scheduler dump at the death: `num_running_reqs=1`, 218624 of 282k tokens computed, 1792 scheduled, no spec decode). Same signature: all four TP ranks stop at the same forward, GPUs spin, `RPC call to sample_tokens timed out` five minutes later, nothing in dmesg. So on this build (c190db1 image, vLLM v0.1.dev20051+g487ecf187, FLASHINFER_MLA_SPARSE_SM120, TP=4 over 4 nodes) a long chunked prefill can stall by itself; DFlash2 and concurrency only raise the odds (3/3 varied soaks within 31-78 min with the draft on, 1 stall in ~3.5 h with it off). I am soaking the 493cb88 image with the draft on tonight to see whether it is a regression between the two; will report. Log: `evidence/soak-crash-20260903T1456Z/` in the linked repo.

https://claude.ai/code/session_01Um6EfS81CU5BXMxYw94pkL

simossss:
Final result of the A/B on the 4-Spark kit (all 150-minute soaks at the 8-sequence cap with 24k and 96k prompts, 4k generations, coding and short requests mixed, 8 workers):

| run | image | draft | fat-expert kernel | outcome |
|---|---|---|---|---|
| 1, 2 | c190db1 | on, 3 tokens | on | stall at 78 and 31 min |
| 3 | c190db1 | on, 3 tokens | off | stall at 35 min |
| 4 | c190db1 | off | on | 150 min clean, then a lone 282k cold prefill stalled an hour later |
| 5 | **493cb88** | on, 3 tokens | (not in that build) | **150 min clean**, 468 req, 0 err, 57 completed 96k prompts |
| 6 | c190db1 | on, 1 token | on | froze at 82 min, and this time `/health` stayed 200 for 35+ min (engine core stuck in the shm-broadcast wait, no RPC timeout) |

So the stall is a regression between 493cb88 and c190db1 on TP=4 across 4 nodes, and PR77 (E2 fat-expert kernel) is not it. Candidates from the window: the indexer workspace opt-in (#86), the spinwait change (#96), or the vLLM build itself. Our 4-node production is back on 493cb88 with the draft on. If you can point me at what else changed in the image between those two commits, I can bisect with the same soak (it reproduces in 30-80 minutes). Logs for every run, including the silent freeze of run 6, are under `evidence/` in https://github.com/punkjazz-labs/glm-5.3-flash-exl3-4x-dgx-spark.

https://claude.ai/code/session_01Um6EfS81CU5BXMxYw94pkL

simossss:
Bisect done, and it is #86 (the indexer-workspace patch). Two derived images from c190db1, each soaked 150 minutes on the 4-Spark kit with the same mix that stalls the stock c190db1 in 31-82 minutes:

- **c190db1 with the pristine vLLM `mla/indexer.py` copied back over the patched one** (fat kernel on, DFlash2 on, 3 tokens): 150 min clean, 488 requests, 0 errors, 121/121 needles, 96k prefill 1280 tok/s.
- **c190db1 with the 493cb88 `exl3.py` overlay but the indexer patch still applied**: stall at 64 min, same signature (all ranks stop on the same forward mid chunked prefill, `RPC call to sample_tokens timed out` 5 min later).

Neither #77 (E2 kernel; cleared twice now) nor #96 (spinwait; stock in our containers) is involved, and vLLM is the same build (`0.1.dev20051+g487ecf187`) in both images. We did not opt into `GLM53_INDEXER_WORKSPACE=rightsize`; the patched module's stock path is what hangs on TP=4 across 4 nodes, always inside a long chunked prefill at a chunk boundary. Our 4-node production now runs c190db1 minus that patch (`recipe/bisect/Dockerfile.noidx` in the linked repo, one COPY over the patched file). I have not tried to find the faulty line inside the wrapper; happy to run a candidate fix through the same soak if you have one. Evidence for both runs: `evidence/bisect-20260903/` in https://github.com/punkjazz-labs/glm-5.3-flash-exl3-4x-dgx-spark.

https://claude.ai/code/session_01Um6EfS81CU5BXMxYw94pkL

simossss:
**Correction to my previous comment: the PR86 verdict is withdrawn.** A second 150-minute soak on the same build with the pristine `indexer.py` (fresh boot, identical mix: short/coding/4k generations/96k and 282k prompts at `MAX_NUM_SEQS=8`) stalled after 24 minutes with the same signature: all four ranks at 96 % utilisation and ~21 W, `No available shared memory broadcast block found in 60 seconds`, `RPC call to sample_tokens timed out` after 300 s, engine dead. The first clean run without the patch was luck. By inspection the patched file is a runtime no-op in stock mode anyway (the knob is unset in our containers, `get_max_prefill_buffer_size` is called once at init from `glm5next/nvidia/attention.py`, no compile cache is persisted), so there was never a mechanism, and I should have required a second run before naming a cause. Apologies to @nood-co1.

What survives from the campaign, for anyone chasing this on 4 nodes:
- the stall reproduces on every c190db1-family build within 24-82 minutes of this mix (7 of 8 runs), with the fat-expert kernel on or off, the DFlash2 draft on, off or at k=1; the one 150-minute run on 493cb88 + draft was clean (n=1, so no conclusion);
- every engine dump (six so far) shows the fatal step scheduling a chunk of a long prompt's chunked prefill: `num_computed_tokens` 24256 / 53809 / 84864 / 89280 (96k prompts) and 218624 (a lone 282k prompt), chunk 1984 or 1792 tokens, usually with 3-6 decode requests carrying 3 draft tokens each in the same step;
- the ranks stay in the kernel (96 % util at idle power) and never return; `/health` sometimes stays 200 with the counters flat.

Next on our side: a reproduction with the freeze catcher armed (py-spy from the host on every rank's worker at the shm-broadcast warning, 4 minutes before the RPC timeout) to get the stuck native frames; I will post them here.

https://claude.ai/code/session_01Um6EfS81CU5BXMxYw94pkL

simossss:
**Stall caught with gdb on all four ranks (TP=4, 4x DGX Spark over a RoCE switch).** A freeze catcher (host-side `gdb -p` on every process of every rank, triggered by the engine's `No available shared memory broadcast block` line 60 s into the stall, 4 minutes before the RPC timeout kills it) fired on the third armed 150-minute soak, at 136 minutes, fatal step = a 1984-token chunk of a long prompt at offset 41152 beside three draft-carrying decodes.

- Every TP worker's main thread is inside `cudaStreamSynchronize`, reached from a Python torch custom op. On the c190db1 build that is the E1 prefill path of `overlay/exl3.py` (`count_stream.synchronize()` after `_stage_counts_to_host`): the first host-blocking point after the GPU stream stopped, not the cause (the 493cb88 `exl3.py` has no such sync and stalled the same way).
- The NCCL proxy-progress thread is the only other running thread on every rank: `ncclProxyProgress -> progressOps -> sendProxyProgress` (transport/net.cc:1362), yielding between polls. A TP collective was launched on all ranks and its network side never completes. GPUs at 96 % utilisation and ~21 W (flag-spinning kernels).
- `NCCL_DEBUG=WARN` has never logged anything at any of the 8 stalls: NCCL sees no transport error.
- The fabric is lossy: per-rank RoCE counters show `packet_seq_err` up to 3.1 M, `out_of_sequence` up to 2.9 M, `roce_adp_retrans` up to 1.3 M, some `local_ack_timeout_err`, global pause only (no PFC, no ECN), MTU 1500, and mixed links (two ranks at 200 Gb/s, two at 100 Gb/s).

So on this cluster the stall reads as a TP collective on a lossy RoCE path that never completes and is never reported, which is why it was independent of image, fat kernel, draft, PR86 and every engine knob. Your 2-node recipe has no switch between the Sparks, which would explain why you never see it. Mitigation soaks are running now (`NCCL_PROTO=Simple` first, then algorithm/buffer/timeout knobs, then the fabric itself); results and the raw dumps are in https://github.com/punkjazz-labs/glm-5.3-flash-exl3-4x-dgx-spark (`evidence/freeze-20260904/`, `recipe/freeze-catch.sh`).

https://claude.ai/code/session_01Um6EfS81CU5BXMxYw94pkL

simossss:
Follow-up with NCCL's own op state, read with gdb on the still-frozen worker ranks during the next stall (this one 6 minutes into a soak with `NCCL_PROTO=Simple`, so the protocol is not the variable):

| rank | proxy frame | op | channel | peer | posted | received | transmitted | done |
|---|---|---|---|---|---|---|---|---|
| 1 | `recvProxyProgress` polling `ibv_poll_cq` | AllReduce, opCount 305624, 24 x 1 MiB steps | 47 | 0 | 16 | 8 | 8 | 8 |
| 3 | `recvProxyProgress` | same op | 28 | 2 | 8 | 0 | 0 | 0 |
| 2 | `sendProxyProgress` (net.cc:1335) | same op | 43 | 3 | | | | |

One allreduce stuck on every rank: receivers have posted RDMA buffers for steps that never arrive (rank 1 waits for steps 9-16 from rank 0, rank 3 for steps 1-8 from rank 2), senders poll completion queues that never complete, nobody reports an error. In those six minutes the receiving NICs' counters moved by thousands (`packet_seq_err` +4,830 and `roce_adp_retrans` +498 on rank 1, `out_of_sequence` +5,057 on rank 2), and rank 2 carries `req_transport_retries_exceeded=8`. Global pause only, no PFC/ECN, MTU 1500, two of four links at 100 Gb/s.

So for this 4-node deployment the cause is packet loss on the RoCE fabric between the Sparks, occasionally left unrecovered and unreported below NCCL; nothing in the recipe is at fault beyond generating the traffic. The fix is lossless RoCE on the switch (PFC/ECN), jumbo MTU and equal link speeds, or a direct-cable topology like yours. Dumps and counters: `evidence/freeze-20260904/` in https://github.com/punkjazz-labs/glm-5.3-flash-exl3-4x-dgx-spark. Thanks for bearing with the detour through PR86.

https://claude.ai/code/session_01Um6EfS81CU5BXMxYw94pkL

hushengkai:
**Independent reproduction on a third 4-Spark kit — same stall signature, but ours self-recovers; py-spy Python-stack evidence and much cleaner RoCE counters**

We run this recipe on 4× DGX Spark (4 nodes, all-200G links through a QSFP switch, DFlash2 draft TP=4, MNBT=2048, `--long-prefill-token-threshold 1024` deployed from #112). Your stall description in this thread matches what our freeze watchdog (sampling /metrics every 10s into a timeseries DB) has been capturing all week:

- The fatal step is always a chunk of a long prompt's chunked prefill — including **fully prefix-cached ~230k re-prefills** (prompt counter static during the stall), sometimes with a lone request (num_running=1), sometimes with decode streams sharing the step.
- All four TP ranks stop at the same forward pass; GPUs sit at 95–96% util (54.9 W on our head); `/v1/metrics` keeps responding but the `vllm:*` counters are frozen (API layer healthy — "frozen, not down").
- py-spy dumps (host-side, sudo) of all head processes at stall time: Worker_TP0 MainThread active inside `apply_exl3_fused_moe` — sampled at `exl3.py:1099` (the argsort area) twice and at `exl3.py:1152` = `count_stream.synchronize()` after `_stage_counts_to_host` once — i.e. exactly the first host-blocking point you identified, not the cause. EngineCore MainThread spin-waits in `sched_yield` inside `shm_broadcast.wait` → `get_response` / `_wait_for_response`. MultiprocWorkerMonitor idle. 8 evidence packages so far (metrics samples, nvidia-smi, docker logs tail, py-spy dumps, metrics snapshot); happy to share.

Two differences from your kit worth reporting:

1. **Our stalls self-recover in 2–15 minutes** — counters resume, no RPC timeout, the engine survives. Our RoCE error counters are far cleaner than yours: cumulative across ~2.5 days of 4-node serving, `packet_seq_err` = 16/23/96/48 and `out_of_sequence` = 48/16/23/96 per node, `roce_adp_retrans` ≤ 5, `local_ack_timeout_err` = 0, `req_transport_retries_exceeded` = 0 everywhere (vs your millions and retries_exceeded=8). Same disease, much milder loss — plausibly why ours drains (adaptive retrans eventually completes) while yours exhausts retries and dies. We're wiring fabric-counter snapshots into the stall evidence package and will correlate the counters with the next stall window; if `packet_seq_err` jumps inside the stall window on our kit too, that closes the loop.

2. Our head engine has additionally died twice outside the stall pattern — once ~24 min into a frozen state (consistent with your RPC-timeout death mode), and once **completely idle** (running=0, 28 min after the last request, no stall signature). The idle death is unexplained so far; we lost the container logs to a `docker rm`+re-run recovery (lesson learned: `docker restart` preserves the evidence). Worth watching whether others see idle deaths after stall-heavy days.

Config notes: since deploying the threshold from #112, scheduler starvation is gone on our kit (4×60k cold prefills: TTFTs spread 165–179 s across four streams, decode interleaves with prefill at ~1.45k tok/s aggregate cold / ~110 tok/s aggregate during the 4-way decode phase), but the stalls continue — consistent with your finding that the two problems are independent. Fabric-side fix (PFC/ECN + jumbo MTU on our switch) is pending; we'll report whether the stalls disappear once lossless RoCE is on — that would confirm your root cause on a third topology.

Thanks for the `.env.tp4.example` knobs — adopting `DFLASH_TOKENS=3` and `MAX_NUM_SEQS=8` next restart window, and your swappiness=0 warning probably saved us an engine death.

hushengkai:
## Follow-up: PFC pipeline restored on our cluster — fabric-class stalls eliminated, but one non-fabric stall remains

Update to my comment above. We found the root cause on our side and fixed it; results below.

**Root cause (our cluster):** the April-era PFC setup had silently died:
- `mlnx_qos` settings are **runtime-only** — lost on every host reboot
- `NCCL_IB_TC=106` existed only in an old manual launch command, never in the current
  TP4 launcher (which builds its own `docker run -e` list — easy to miss even when the
  doc has it)
- Net effect: **100% of NCCL traffic rode priority 0** (proof: `rx_prio0_packets=11.6B`,
  prio 1–7 exactly zero, prio 3 zero) on a global-pause fabric — exactly your "global
  pause only" configuration

**Fix applied (all 4 nodes):**
- `mlnx_qos --trust dscp` + `--dscp2prio set,26,3` + `--pfc 0,0,0,1,0,0,0,0` + `ethtool -A rx off tx off`
- `NCCL_IB_TC=106` added to **both** launch scripts' docker -e arrays
- systemd oneshot unit per node to survive reboots (already proven: host rebooted once
  during the window and the unit restored everything)
- MTU was already 9000 on our side

**Verification:** ran our historically freeze-triggering load (4× concurrent 60k cold
prefills, the scenario that produced 6 of our 8 stalls): **~218M packets/node flowed
through the PFC-protected priority 3 queue with RoCE loss counters completely flat**
(packet_seq_err=1, out_of_sequence=20, adaptive retrans=0). No fabric-class stall.

**However — one important new data point:** during that same stress run a single
~100 s self-recovering stall occurred, and py-spy caught it live:

- Worker: `torch/cuda/streams.py:108 synchronize` ← `apply_exl3_fused_moe (exl3.py:1152)`
  (the `count_stream.synchronize()` after `_stage_counts_to_host`)
- EngineCore: `sched_yield` in `shm_broadcast.wait`
- **RoCE counters flat during the stall** — so this one was *not* fabric-driven
- Trigger fingerprint: two large requests completed and freed KV (running 6→4, KV
  13.7%→9.8%) while 4×60k prefills were in flight; startup log shows the fat-expert
  path at `fat_layers=42/42 (100.0%), effective_tier=kernel, max_rows=2048, fallback=0`

This refines "the exl3 frame is just the first host-blocking point, not the cause":
that holds for the fabric-loss stalls — but there appears to be a **second, non-fabric
stall class with the same Python-level signature**, where the RoCE path is provably
clean. Have you seen stalls with flat RoCE counters after your switch-side fix? If yes,
did you manage a native-side capture of one? Our watchdog now snapshots per-node RoCE
deltas + full py-spy stacks on every event, happy to share packages.

simossss:
Thanks for the detailed follow-up. Yes: after fixing our fabric, we also reproduced stalls with flat monitored RoCE error/drop/retry deltas. My September 4 comment attributing every stall to packet loss is outdated and too strong.

We subsequently collected native NCCL state followed by namespace-correct CUDA-GDB captures. Rank 2 retained one block of `sparse_mla_prefill_kernel<ModelType2,ComputeMode0,16,2048,64>` at the same selected PC across three captures, while ranks 0, 1 and 3 waited in NCCL. Receive-tail publication and outgoing peer credit were present, but rank 2 had not published outgoing GPU data. Exact-function disassembly places the PC at +0x7960 in the IO producer loop after a CTA barrier; this localizes our observed progress failure, but does not identify the internal race.

This is why `count_stream.synchronize()` at exl3.py:1152 remains a wait site rather than proof that E2 is faulty, even with clean fabric counters. GPU/native state is needed to identify the incomplete dependency. Our TP4 H16 specialization also differs from TP2's H32 path.

Our bounded workaround slices only the final sparse-attention call into at most 64 query rows, selecting the existing H16 decode/merge kernels. It preserves the outer 2048-token batch, EXL3, DFlash k=3 and fabric. Numerical/kernel-selection checks and subsequent mixed-load qualification passed. The independent scheduler improvement is the native long-prefill threshold of 1024. Current patch and scoped qualification are in the [public recipe](https://github.com/punkjazz-labs/glm-5.3-flash-exl3-4x-dgx-spark) and [evidence summary](https://github.com/punkjazz-labs/glm-5.3-flash-exl3-4x-dgx-spark/blob/main/EVIDENCE.md).

Your request-completion/KV-release timing is a useful additional discriminator. The ~100-second self-recovery does not establish that your event is the same H16 failure. The most useful comparison would be your pinned image/backend, exact request timeline and per-rank native/GPU state around that event. A healthy HTTP metrics endpoint alone is insufficient.

The proposed text in this PR also needs refreshing: disabling DFlash did not eliminate our original stalls, and the old switch-only explanation should not be treated as the final diagnosis. Current scoped recommendations depend on the attention workaround, not just the env knobs in this old diff.

plotarmordev:
### Review follow-up — `bba02ae7b0c6`

Your knob measurements and the later corrections in this thread are acknowledged. The documentation must not retain a causal explanation you subsequently withdrew.

Please align the PR body, README and TP4 configuration caveat with the latest findings, or remove the stale causal claims while preserving the clearly dated, kit-specific tuning table. Do not turn the current workaround into a universal root-cause claim. Mixed-prefill “off” remains a valid value; that spelling does not need changing.

Next action: documentation correction, then our review of the revised guidance. No additional soak is requested merely to fix the text.

plotarmordev (approved):
Exact-head approval: 1ebad7129c77a08481d025fe7612d8ade2a3339a. Read the final carried documentation and live September9 comment5599489540: withdrawn draft and packet-loss-only causality corrected; H16 progress localization, unknown internal race and <=64-query-row workaround scoped to the reported kit. Original knob assignment SHA2569040421857e38af17c45f123fa59bc0c6f1c7811d5a7b98e3c3fee9165241b7b and dated explanation SHA25615d15cadd59dab36a6c6ebdd0e8219639f58425430c390406b00ed375caeb043 preserved. Scope, secret and deployment-content scans cover clean worktree and entire pushed range; no issues identified. Documentation verification only; no new soak or GPU qualification.

cc17486:
## Repro data points for the TP=4 stall (4-Spark kit, 1.6.0 + InstantTensor image)

Adding our reproductions to the stall attribution here — several are new trigger shapes not covered by the current note. Kit: 4× DGX Spark (GB10), TP=4, repo @ `ca855766` (1.6.0) + local launcher port (E3 default-on, `--ulimit nofile=65536` — required by InstantTensor, see bottom), image `:exl3-instanttensor` rebuilt on-stamp locally. fair v5 scheduler, `GLM53_APC_RETENTION_INTERVAL_SWA=0`, E3 grouped active (`effective_tier=grouped`), indexer-workspace `rightsize` active (confirmed in boot log), 1M ctx, KV fp8, `GPU_MEM_UTIL=0.75`.

### Signature (identical every time)

Worker step wedges: all 4 ranks spin ~96% GPU with `Running=1–2` requests, `prompt_tokens_total` / `generation_tokens_total` frozen, zero scheduler/loggers output. NCCL `HeartbeatMonitor` TCPStore check failures on the worker ranks; head dies ~600 s later on the shm mq dequeue `TimeoutError` → `EngineDeadError`. Matches the PR's H16 sparse-attention step description.

### Reproductions (all on fair v5 unless noted)

| # | Config | Trigger | Result |
|---|---|---|---|
| 1 | k=7, seqs=4, MNBT=7168 | lone 100k cold prefill | **wedge** (client 600 s timeout) |
| 2 | k=7, seqs=4, MNBT=2048 | lone 100k cold prefill | **wedge** |
| 3 | k=3, seqs=8, MNBT=7168 | mixed soak (2× ~25k-token prompts + 2 short chats, conc 4) | **wedge @ ~9 min** |
| 4 | k=3, seqs=4, MNBT=7168 | same mixed soak | **wedge @ ~25 min** |
| 5 | k=3, seqs=8, MNBT=7168, `GLM53_MIXED_PREFILL_CHUNK=off` | lone 300k cold prefill | **wedge** |
| 6 | k=3, seqs=4, MNBT=7168 | lone 300k cold prefill (repeat ladder) | **wedge** |
| 7 | k=5, seqs=4, MNBT=7168 | lone 300k cold prefill | **wedge** |
| 8 | k=5, seqs=4, MNBT=2048 | lone 300k cold prefill, first try | **wedge** |

### Passes (contrast)

- k=3 and k=5: lone 100k / 256k cold prefill — multiple clean passes at both MNBT 2048 and 7168 (full 8k–300k ladders).
- k=5, seqs=4, MNBT=7168: 30-min mixed soak, 532 requests, 0 errors (the only soak config to finish clean today).

### New information vs the current note

1. **Drafter depth modulates the trigger threshold**: k=7 makes even a *lone 100k* cold prefill fatal 2/2; k=3/k=5 pass that rung reliably.
2. **The ~300k-class lone cold prefill is the dominant, config-independent trigger** — wedged on every config tried (k=3, k=5, off-mode; MNBT 2048 and 7168 alike; #8 wedged on the very first attempt at 2048, so the smaller chunk does not protect). 100k/256k are reliably safe on k≤5.
3. **fair v5 does not prevent the mixed-load wedge**; seqs 8→4 only delays onset (~3×: 9→25 min).
4. All repros ran with the bounded indexer-workspace patch active (`rightsize`), so this path escapes the current mitigation.
5. InstantTensor loading required `--ulimit nofile=65536` in the container run — at the Docker default 1024, the loader's `_determine_io_params` all-reduce dies with `socket: Too many open files` (NCCL host-network sockets + O_DIRECT handles). Unrelated to the stall, but every TP4 InstantTensor user will hit it.

### Local mitigation (working)

Token-freeze watchdog: poll `/metrics` every 15 s; `running>0` with prompt+gen counters frozen ≥120 s → launcher restart. 3 hands-off recoveries today (~9–10 min back to healthy). Happy to share the script.

Full per-run logs (bench JSON, cold-ladder raw, soak logs, watchdog log) retained on our repo branch — can attach specific excerpts or run targeted experiments (e.g. `ENFORCE_EAGER=1` to exclude cudagraphs, draft-off 300k hammer, `NCCL_DEBUG=INFO` capture) if useful.

PR #125, opened 2026-09-04
Fix fine-grained prefix-cache hits for hybrid GLM KV

rwl4:
Restore 64-token prefix-cache lookup for the GLM hybrid KV layout.

`KpoolTailSpec` explicitly opts out of prefix caching, but its `KpoolTailManager` was still included when vLLM decided whether fine-grained hash lookup was safe. Because that transient manager does not support fine-grained lookup and has a wider block, it forced every reusable cache group back to the 3,584-token hybrid-page boundary.

This change excludes only KV groups whose specs explicitly set `participates_in_prefix_caching=False`.

MLA and Mamba still participate in and constrain cache matching. KpoolTail remains transient and is rebuilt normally. Physical KV allocation remains 3,584-token pages; only prefix lookup granularity changes to 64 tokens.

The patcher is also additive: it can install this fix when the published image already contains the earlier `glm53-hybrid-apc` overlay.

## Measured impact

Tested directly against a 2× DGX Spark TP=2 deployment:

- GLM-5.3 Flash EXL3
- DFlash2 `k=7`, draft TP=2
- FP8 KV
- `max_model_len=1,000,000`
- `max_num_batched_tokens=7,168`
- 44 unchanged tool definitions
- Thinking disabled
- Stable per-run cache salt
- Exact rendered-token LCP measured before every request

| Append workload | Cached before | Cached after | TTFT before | TTFT after |
|---|---:|---:|---:|---:|
| Extend a 12K prefix to 24.2K | 7,168 | 11,968 | 14.38s | 10.18s |
| Short append to a 24,208-token prefix | 17,920 | 24,192 | 5.95s | 0.53s |
| Next short append at ~24.3K | 21,504 | 24,256 | 2.88s | 0.71s |

For the closest matched short-append case:

- Cache-hit ratio: 73.7% → 99.5%
- Reusable-prefix tokens unnecessarily recomputed: 6,288 → 16
- Redundant repeat-prefill reduced by 99.75%
- TTFT reduced by 91.1%
- Time to first token improved by 11.2×

The next append showed a 75.4% TTFT reduction, or 4.1× faster.

A separate semantic long-document test retained the correct unique checksum
across the cached follow-up:

- Cold: 8.31s
- Warm: 1.15s
- Prefix-cache hit ratio: 99.44%
- Speedup: 7.24×

This PR does not claim a steady-state decode tok/s improvement. It does not touch model execution, DFlash2 sampling, decode scheduling, or KV allocation.  The improvement comes from eliminating redundant prefill before the first token.

## Correctness and safety

- Only managers whose KV specs explicitly opt out of prefix caching are ignored
  by the fine-grained eligibility check.
- MLA and Mamba remain part of the hybrid cache minimum.
- KpoolTail is not made persistent or shareable.
- Physical 3,584-token attention-page geometry is unchanged.
- Cold requests follow the same compute path.
- If upstream source anchors change, the overlay fails closed.

## Validation

- Focused patch test passed against the published Mia image.
- Patched source compiles successfully.
- Reapplying the overlay is idempotent.
- TP=2 startup and CUDA graph capture completed on both ranks.
- Built-in shape/sampler warmup: 20/20 requests passed.
- Four-request exact-LCP append-only canary passed.
- Long-document semantic cold/warm test passed.
- Both ranks remained healthy with zero restarts and no engine errors.

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,7 +111,7 @@ same path as the compact-64 fp8 serve (not NVFP4 KV).
 | Worker | `WORKER_USER@WORKER_IP` (this kit: `zurih@10.0.0.2`), `--headless`, `glm53-exl3-worker` |
 | Fabric | CX7 QSFP: `enp1s0f1np1`/`rocep1s0f1` ↔ `enp1s0f0np0`/`rocep1s0f0`. Image NCCL (`USE_HOST_NCCL=0`) |
 | Attention | `FLASHINFER_MLA_SPARSE_SM120` (NoPE MLA padded into GLM_NSA 576-wide) |
-| KV | `--kv-cache-dtype fp8` → packed **`fp8_ds_mla`** (target). Draft DFlash2 KV is `auto`/bf16. Latest validated 7168/rightsize pool: **1,243,902** tokens / **1.24×** at 1M. `--enable-prefix-caching` (block-aligned hits; see Prefix caching) |
+| KV | `--kv-cache-dtype fp8` → packed **`fp8_ds_mla`** (target). Draft DFlash2 KV is `auto`/bf16. Latest validated 7168/rightsize pool: **1,243,902** tokens / **1.24×** at 1M. `--enable-prefix-caching` (64-token hash lookup; 3584-token physical attention pages; see Prefix caching) |
 | Experts | packed trellis + suh + svh + mcg, codebook MCG, **one fused `exllamav3_ext.exl3_moe` launch per layer** |
 | Dense / shared / attn / embed / lm_head | native (unquantized) |
 | Spec | **DFlash2 k=7** (`incoai/GLM-5.3-Flash-DFlash2`); draft KV `auto`/bf16, draft TP=2, FLASH_ATTN. Rollback `SPEC_METHOD=mtp` |
@@ -305,14 +305,15 @@ Keep **`SKIP_MM_PROFILING=1`** — a max-size image+video dummy profile OOMs thi
 **NVFP4 KV is not available here.** FlashInfer’s SM12x NVFP4 kernels are dense MHA,
 not sparse MLA. Do not confuse that with NVFP4 **weights** (`--moe-backend marlin`).
 
-## Prefix caching (this kit, 2026-08-30)
+## Prefix caching (this kit, 2026-09-04)
 
 `--enable-prefix-caching` is on. The OpenAI API is **stateless**: the client
 resends the full history each turn; vLLM hashes that prefix. Concurrent chats
 do **not** mix activations. `--max-num-seqs 4` is four **in-flight** generations,
-not four parked sessions. MLA `KpoolTailManager` disables **fine-grained**
-hits — only **block-aligned** tokens count (3584-token hybrid align).
-`KpoolTail` already opts out of the hybrid min (1-block circular scratch).
+not four parked sessions. Physical sparse-MLA attention pages remain 3584
+tokens, but reusable prefix hashes now match at the underlying **64-token**
+cache-block granularity. `KpoolTail` is 1-block circular scratch and already
+opts out of prefix caching.
 
 `dflash` is `use_eagle()`. GLM never sets `is_eagle_group` (that annotator is
 DeepseekV4-only), so stock HybridKVCacheCoordinator flagged **every** group.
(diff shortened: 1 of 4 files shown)


plotarmordev:
**Status:** keep as-is — no change requests. This stays in the queue and will be A/B'd stacked after #130 passes.

**Verified:** read the full diff against current main. The single runtime hunk — rewriting the `unsupported_partial_hit_managers` comprehension to filter on `group.kv_cache_spec.participates_in_prefix_caching` so `KpoolTailManager` (block-size-4 transient scratch) no longer vetoes 64-token hits while MLA and mamba remain in the hybrid min — is the same core mechanism as #84 in its leanest possible form (~61 lines, additive, idempotent onto the deployed `glm53-hybrid-apc` overlay). The measured receipts on the exact owner config (TTFT 5.95 → 0.53 s, hit rate 73.7 → 99.5%) are consistent with the mechanism. The conflict against current main is README-only doc churn — trivial to resolve when the time comes; please don't bother rebasing yet.

**Why stacked after #130:** the one thing this PR's thread doesn't carry is acceptance data, and #84's thread now has two independent probes showing a warm-turn spec-decode accept-length drop of −0.27…−0.53 with 64-token hits (mechanism: on the 3584 grid a warm re-prefill rebuilds the drafter's full 2048-token window from fresh hidden states — an acceptance bonus that a 64-grid hit forgoes). #130's safe-replay path exists precisely to restore that fresh-window condition, so fine-grained hits get their fair trial only stacked on a replay guard. If that stacked A/B passes, this minimal variant is our preferred form of the fine-grained change (with #84's runtime `hash % kpool` verification and kill switch as the alternative if we want the hardening).

**Waiting on:** our #130 A/B first (per-group retention, SWA=0, global interval unset, fine-grained off), then the stacked run with fine-grained on under the identical protocol, on our 2x GB10 TP=2 E3 kit. Nothing needed from you in the meantime.

plotarmordev (dismissed):
Requesting changes — full rationale in https://github.com/MiaAI-Lab/GLM-5.3-Flash-EXL3-2x-DGX-Sparks/pull/125#issuecomment-5603757416. Edit list:
1. README rebase only: re-apply the PR's two README paragraphs onto current main dc6936c (E3/adaptive-k doc churn is the sole conflict). Please wait until the stacked A/B is scheduled — no need to rebase yet.
2. Nothing else: the runtime hunk and re-baselined bench_prefix_cache.py are accepted as-is.
Note: no standalone merge — 64-token hits ship only stacked on #130's replay guard.

plotarmordev:
### Review follow-up — `318795b15a0c`

Updating our earlier “nothing needed; wait for #130” note: #130 is merged, and #171 is now the active fine-grained candidate. The old queue description is no longer current.

Your lean mechanism and published measurements remain acknowledged, but this branch now touches the replay-guard file rewritten by #130; the earlier “README-only conflict” statement should not be relied on for today's integration.

Please do not rebase or rerun benchmarks for us now. We own deciding whether this remains a useful alternative after the #171 work. The #130 merge does not itself prove that the previously promised A/B passed; we are not making that claim. No current author action is required.

plotarmordev:
### Maintainer direction — retain this lean alternative

We have decided to keep #125 open as a **specifically retained lean alternative**, not a duplicate of #171. Thank you @rwl4 for the minimal participating-group lookup mechanism, measured append-workload results, and `bench_prefix_cache.py` rebaseline; those remain distinct useful work even while #171 is the active candidate.

The revisit trigger is **#171's final qualification/disposition decision**, whether that results in adoption, a narrowed supported scope, or a decision not to proceed. Maintainer @plotarmordev owns revisiting #125 at that point and explicitly deciding whether the lean form should be integrated or retired. #171 is still draft and unqualified; neither its existence nor #130's merge establishes that the earlier promised stacked A/B passed.

No rebase, rerun, or repeated author request is pending. If we choose to revive this implementation, maintainers will first re-derive it against the replay-guard file rewritten on main rather than relying on the obsolete README-only-conflict assessment. Until that trigger, preserve this branch and its evidence as-is.

plotarmordev (commented):
## Hold update

Public head checked: `318795b15a0c42ed37469516ce6401b6b1aae6df`.

**Why it is held:** This remains a useful lean alternative, not an automatic duplicate of #171. We now also have the rebase in #195 to consider.

**What we tested:** In an integrated version of the cache mechanism, long and medium append requests gained reuse, but the short append reused 7,168 tokens instead of 10,752. Reference answers passed. This does not qualify the public branch by itself or prove internal numerical/state safety.

**Next action — me, @plotarmordev:** Decide which implementation to retain after the #171 qualification/disposition decision, then integrate and check that exact version with the replay guard preserved.

**Author action now:** None. Please preserve the branch and measurements. No rebase or new run is owed while this direction is undecided. If selected, the agreed checks must keep short/no-benefit cases as well as successful long-prefix cases.

PR #127, opened 2026-09-05
Porting chat template update from upstream

RustyAiLab:
## Description and Motivation

Porting chat template update from upstream commit - https://huggingface.co/zai-org/GLM-5.3-Flash/commit/690b705278a3a58e538fcb37c2ca8b5f9511213c

## Testing

Boots up fine and works with ZCode harness.

## Checklist:

### DIFF
--- a/files/chat_template.jinja
+++ b/files/chat_template.jinja
@@ -73,8 +73,8 @@ For each function call, output the function name and arguments within the follow
                 {{- emit_audio() -}}
             {%- endif -%}
         {%- endfor -%}
-    {%- else -%}
-        {{- content }}
+    {%- elif content is not none -%}
+        {{- content -}}
     {%- endif -%}
 {%- endmacro -%}
 {%- macro tool_response(text) -%}
@@ -164,7 +164,7 @@ For each function call, output the function name and arguments within the follow
 {%- if tc.function %}
     {%- set tc = tc.function %}
 {%- endif %}
-{{- '<tool_call>' + tc.name -}}
+{{- '<tool_call>' ~ tc.name -}}
 {% set _args = tc.arguments %}{% for k, v in _args.items() %}<arg_key>{{ k }}</arg_key><arg_value>{{ v | tojson(ensure_ascii=False) if v is not string else v }}</arg_value>{% endfor %}</tool_call>{% endfor %}
 {% endif %}
 {%- elif m.role == 'tool' -%}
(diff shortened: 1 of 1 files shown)


plotarmordev:
**Status:** approved pending one live tool-call smoke, which we will run in the next maintenance window. No change requests — merge-ready from our side once that smoke passes.

**Verified:** merged your head onto current main locally (clean, no conflicts; 10 insertions / 6 deletions, `files/chat_template.jinja` only).

- **Provenance:** all 6 changed hunk bodies, including context, are byte-identical to the independently fetched upstream diff of zai-org/GLM-5.3-Flash commit `690b705278a3a58e538fcb37c2ca8b5f9511213c`. The fetched upstream pre/post blobs hash to the old/new blob IDs in the published upstream patch. There are no extra hunks, and the PR preserves our local deltas (the `thinking`/`enable_thinking` compatibility block, conditional generation suffix, terminal newline). Note for the record: the upstream commit message only describes the early-termination change; the null-content and `+` → `~` hunks are in the upstream diff but not its title — provenance rests on the diff, which we verified hunk-by-hunk.
- **CPU rendering:** a sandboxed Jinja harness (matching the HF rendering convention) ran the pre/post templates over the relevant message shapes: plain user turn with thinking on and off (byte-identical), assistant tool calls with reversed-arrival tool results (byte-identical, results correctly reordered into call order), tool result immediately followed by user (byte-identical), and system prompt + tools list (byte-identical). All marker sequences consumed by the `glm47` tool parser, the `glm45` reasoning parser, and our reasoning-stop overlay patch are unchanged; the existing `tests/test_chat_template.py` (7 tests) passes.
- **The actual defect is reproduced and fixed:** assistant `content: null` with tool calls renders a literal `None` into history pre-patch (we captured the exact 4-byte span); post-patch those bytes are gone with no other change.
- **8 fallback/ordering cases pass byte-identical:** missing/unknown/duplicate result IDs, missing/duplicate call IDs, and list-of-outputs entries with missing/unknown/duplicate IDs — each still falls back to arrival order, so the early-break hunks do not alter the rejection paths.
- One scope note: the `+` → `~` change adds Jinja string coercion for non-string tool names, but valid string names render identically — this PR does not change (and does not need to change) anything about separators between normal adjacent tool calls.

**Change requests:** none.

**Waiting on:** a live smoke against a boot serving this template, scheduled for the next maintenance window on our 2x GB10 TP=2 E3 kit. Seven sequential requests, all `temperature 0`, non-streaming: (1) plain reply, thinking on; (2) plain reply, thinking off; (3) parallel tool generation — system prompt + tools list, expecting exactly two `lookup` calls with distinct IDs; (4) tool continuation with assistant `content: ""` and reversed-arrival results, expecting the exact composed answer; (5) tool → immediate user turn with empty content; (6) same continuation with `content: null`; (7) same tool → user with null content. Any raw XML in content, malformed/blank arguments, missing/duplicate tool IDs, a length finish, or a template error fails the smoke. We will post the results here.

plotarmordev:
Merged via https://github.com/MiaAI-Lab/GLM-5.3-Flash-EXL3-2x-DGX-Sparks/pull/156 (rebased onto current main over #150; your commit and authorship are preserved). Thank you — closing this one in favour of the merged replacement.

PR #129, opened 2026-09-05
docs: historical image4 rejection and current LIMIT_MM admission semantics

Capicua25x:
## Description and Motivation
Docs only. Adds `docs/images-per-request.md` and one sentence on the README's Vision row.

What it documents: with the kit's then-default (2026-09-05) `LIMIT_MM='{"image":4,"video":1}'`, any OpenAI-compatible chat client that resends the conversation on every turn (OpenCode, Open WebUI, LibreChat, agent harnesses) works for four images and then fails every later turn of the session with `At most 4 image(s) may be provided in one prompt`, because vLLM checks the limit per request and the resent history now carries five. Not a client bug and not obvious from the error text; it cost me a morning on a stock deployment of this kit (2× DGX Spark, `.env` defaults, vision on).

The page explains the mechanism, gives the recipe-level fix (raise `LIMIT_MM`; with `SKIP_MM_PROFILING=1` it is not a boot-time reservation, so the KV pool should be unchanged, to be verified after the restart) and a client/gateway-level fix (keep the newest N image parts, replace older ones with a note the model understands; reference implementation linked), plus a six-image curl protocol to reproduce and verify either fix. No knob, script or default changes; nothing affects measured numbers.

## Related Issues
None found for this.

---
## Testing
Docs only, no script change. The claims were measured on the running serve:
* six-image conversation straight to the head with the then-default `.env` → the error above;
* the same conversation through a trimming proxy (N=4) → the model answers that it sees four images (3–6) and reads the last one;
* the protocol in §5 of the doc is what was run, `max_tokens` 600 so the thinking tokens leave room for the answer.

## Current-main reconciliation (2026-09-15)

Reviewed head `f18e1a43db785b14990803f9b74657d39643a388`: current main uses `LIMIT_MM='{"image":48,"video":1}'`. This is an image-item count admission threshold: up to 48 passes that check, 49 exceeds it; memory, context and other limits may prevent completion below the ceiling. The dated image4/image16 measurements and proxy observations remain historical author-reported results, not a measured concurrent image budget or a capacity/performance guarantee. No live image request or new memory measurement was run for this review.

### DIFF
--- a/README.md
+++ b/README.md
@@ -118,7 +118,7 @@ same path as the compact-64 fp8 serve (not NVFP4 KV).
 | Context | **1M** (`MAX_MODEL_LEN=1000000`). Latest validated 7168/rightsize pool: **1,243,902** tokens / **1.24×**. Pool size varies with MNBT, activation/graph reservations and hybrid block geometry |
 | Tools / reasoning | `--tool-call-parser glm47 --enable-auto-tool-choice --reasoning-parser glm45` |
 | Graphs | on (`ENFORCE_EAGER=0`) — MTP capture `1 2 3 4 6 8 12`; DFlash2 capture `1 2 4 8 16 24 32` |
-| Vision | on (`LANGUAGE_MODEL_ONLY=0`) — image + video, `--limit-mm-per-prompt {image:4,video:1}`, `--skip-mm-profiling` |
+| Vision | on (`LANGUAGE_MODEL_ONLY=0`) — image + video, `--limit-mm-per-prompt {image:4,video:1}`, `--skip-mm-profiling` | Stateless chat clients resend every image each turn, so the 5th image of a session trips this limit — see `docs/images-per-request.md`.
 | Ablit | **off** (`ABLIT=0`). Stock `o_proj`. Set `ABLIT=1` to enable; see [Abliteration](#abliteration-ablit1) |
 
 Kernels: `TORCH_CUDA_ARCH_LIST=12.1a`. ExLlamaV3 pin `c5d9c657` (0.0.43) exposes
--- a/docs/images-per-request.md
+++ b/docs/images-per-request.md
@@ -27,17 +27,33 @@ The client shows it as an error bubble; the session never recovers on its own.
 Checked straight against the head (`gx10a:8888`) with a synthetic six-image conversation: the request is rejected
 with the message above. It is not a client bug.
 
-## 3. Fix A — raise the limit in the recipe
+## 3. Fix A — raise the limit in the recipe (measured)
 
 `LIMIT_MM` is the knob. `.env`:
 
     LIMIT_MM='{"image":16,"video":1}'
 
 then `./start.sh restart`. With the README's `SKIP_MM_PROFILING=1` (the default: a max-size image+video dummy profile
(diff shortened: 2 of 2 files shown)


Capicua25x:
Verified Fix A on the cluster (2× DGX Spark, this kit's `.env` with `LANGUAGE_MODEL_ONLY=0`, `GPU_MEM_UTIL=0.875`, `MAX_NUM_BATCHED_TOKENS=2048`, 1M):

- `LIMIT_MM='{"image":16,"video":1}'` boots with the KV pool unchanged: `Available KV cache memory: 12.89 GiB · GPU KV cache size: 1,229,468 tokens` vs 12.93 GiB / 1,234,299 with the default 4 (noise-level difference, no reservation as expected with `SKIP_MM_PROFILING=1`).
- One request with 16 distinct 1080×1920 screenshots: HTTP 200 in 63 s, all 16 labels read in order, `prompt_tokens 43136` (≈2.7k tokens per screenshot), head container alive, no OOM. Host-side free memory on the head sat at ~1 GB during the request (it is ~1–2 GB at rest with vision on at util 0.875), so on this UMA the headroom for concurrent image-heavy requests is thin; a client-side cap remains the safer default for busy multi-user serves.

I'll fold these numbers into the doc if you'd like them in the page itself.

Capicua25x:
Folded the measurements into the page (§3, §3b, §4, §6): the `LIMIT_MM=16` boot table, the 16-screenshot request, and the point that the default 4 is really `MAX_NUM_SEQS 4 × 4 = 16 images in flight` — the quantity this UMA is sized for — plus the in-flight budget we run on the client side and the ~1 GB host headroom note.

plotarmordev:
### Review follow-up — `eb9bd4805605`

The stateless-history/per-request image-limit explanation and your measured workarounds are useful. Two documentation changes remain.

Current main uses an image limit of 48 (main b3bd1f69), not 4. Please distinguish the historical fifth-image reproduction from the current default rather than presenting it as today's behavior. Also remove the inference that MAX_NUM_SEQS multiplied by the per-request image cap defines the machine's safe concurrent image-memory budget; that requires more than the reported single-request measurements.

Keep the dated measured tables and the mechanism explanation. This is a documentation request, not a request to rerun the vision benchmark.

Capicua25x:
Reconciled in `7a263e9` (branch updated with main, `d17bb21`).

1. Historical framing: the 4-image default and the fifth-image repro are dated and marked as such; the doc now opens with current main's `{"image":48,"video":1}` (`b3bd1f6`) and notes the same session now trips at the 49th distinct image. README's vision row carries the current default and a limit-agnostic version of the note.
2. Budget inference removed: §3b is now "observed, not a measured envelope" — single-request data only, no derived in-flight budget, and the "64 on paper" line is gone. The proxy cap in §4 is labelled as our client-side guard, not a machine limit.

Dated tables and the mechanism section are unchanged.

plotarmordev (approved):
Exact-head review: f18e1a43db785b14990803f9b74657d39643a388. Current launcher defaults and both argument paths agree with image48 count admission; corrected wording explicitly separates that threshold from memory/context capacity and successful completion. Historical measured tables retained and concurrency envelope left unquantified. Complete proposed docs and carried diff reviewed; publication tree and commit-range scope/secret/deployment-content scans found no issues. Docs-only validation; no fresh live-image or memory/performance qualification.

PR #131, opened 2026-09-05
Improve GB10 EXL3 prefill kernels and reduce indexer memory peak

RodriMora:
## Description and Motivation

I focused on fat-expert GEMM time and peak temporary indexer memory during GB10 prefill.

I added M64 tiling with launch bounds, paired gate/up projections, fused clamped SwiGLU, and double-buffered `cp.async` staging. I kept both K-loop barriers. For the indexer, I added `del logits` after the final prefill consumer so the next allocation doesn't overlap the previous tensor.

I haven't changed weights, quantization, context limits, KV precision, sampling or allocator defaults.

I compared the direct/scatter kernels against upstream `3021f24`, interleaving three panels per image on each GPU:

| Device | Upstream median | My branch median | Kernel-time reduction |
|---|---:|---:|---:|
| Head GB10 | 9.453 ms | 7.158 ms | 24.3% |
| Worker GB10 | 9.205 ms | 7.019 ms | 23.7% |

That measures the M64/async path, not overall inference. I also reproduced a 507.01 MiB reduction in peak temporary indexer allocation on both GPUs, from 1045.26 to 538.25 MiB. Indexer loop time was effectively unchanged. I am not claiming an equivalent reduction in host RSS or an end-to-end speedup.

I've included the scripts, configuration, every measured panel and the failed attempts in the [validation report](https://github.com/RodriMora/GLM-5.3-Flash-EXL3-2x-DGX-Sparks/blob/5584f739f4b962a390d181795007fec95a3f6c12/docs/exl3-prefill-validation.md).

## Related Issues

None.

---

## Testing

- The CPU suite passed 108 checks and skipped ten integration checks.
- I ran the full image GPU self-checks on both GB10s. All nine built production/test file hashes match this branch.
- Numerical checks, nondefault-stream execution and graph replay passed memcheck, racecheck and synccheck on both GPUs. Racecheck reported zero errors and warnings. The initial synccheck runs failed after a barrier-tracking overflow warning. I reran them with `--num-cuda-barriers 64`; those passed. I've kept the original failures in the report.
- I booted my branch without an arena cap using the existing `rightsize` workspace and 16 ms spinwait opt-ins. Automatic profiling provided 15.67 GiB KV memory and 1,057,491 KV tokens for the 1M context. The cold 8k/64k and 100k/300k requests, prefix-cache check and median-five decode tests passed. Every cold request had zero cache hits. All fifteen measured decode requests produced 400 tokens. The report includes TTFT and acceptance at each draft position.

I couldn't complete the full-model comparison safely. With stock workspace/spinwait, both images failed startup. Upstream had 10.40 GiB available KV memory and my branch had 10.47 GiB, below the required 14.52 GiB. The successful opt-in boot doesn't count as a stock-default pass.

For the matched-capacity comparison, I set the upstream control to 16.25 GiB KV per rank. It reported 1,118,466 KV tokens, but its first cold 8k warmup hit the head memory guard at 112.04 MiB available. The guard and controller stopped both model containers, and both hosts recovered. No measured baseline samples completed. I did not run my branch or the return control at that capacity.

I've paused further model runs rather than shrink KV capacity or add an arena cap to get a pass. The [memory audit](https://github.com/RodriMora/GLM-5.3-Flash-EXL3-2x-DGX-Sparks/blob/5584f739f4b962a390d181795007fec95a3f6c12/docs/gb10-memory-budget-audit.md) rules out first-time fat scratch allocation as the explanation, but doesn't identify the allocation responsible for the drop.

I'm keeping this in draft. The isolated kernel and memory results are reproducible, but I haven't established a safe matched live comparison, absence of a decode regression, long-run stability or concurrent-service safety.

---

## Checklist:

- [ ] Defaults still work out of the box; a new knob has a sane fallback consistent with the existing ones.

I've checked reproducibility for the isolated measurements and the documented opt-in boot. I've left the defaults box open because both images failed that configuration on this kit.


RodriMora:
Could the existing stock-startup capacity failure stay outside this PR's scope, with the defaults checkbox left open and the limitation documented?

At the tested 1M context and .87 utilization, stock workspace/spinwait startup failed on both upstream `3021f24` and this candidate. They reported 10.40 and 10.47 GiB available KV memory, respectively, versus 14.52 GiB required. The successful candidate boot used the existing `rightsize` workspace and 16 ms spinwait opt-ins, without an arena cap. It is not a stock-default pass.

The PR's claims are limited to the interleaved isolated kernel measurements and the reproduced 507.01 MiB temporary indexer peak saving. Full-model throughput and absence of a decode regression remain unverified against a matched live control. The fixed 16.25 GiB/rank upstream attempt hit the head memory guard during its first cold 8k warmup; no measured baseline completed.

Model runs are paused while I audit the existing memory records. I will not retry that configuration unchanged, reduce KV capacity, or silently add an allocator default to obtain a pass.

Would you prefer to review this narrower scope and track default startup separately, or require the default-deployment issue to be resolved before this PR leaves draft? [Commands and complete receipts](https://github.com/RodriMora/GLM-5.3-Flash-EXL3-2x-DGX-Sparks/blob/f5a784d01816e9decbc11662841b8d420dff77d5/docs/exl3-prefill-validation.md).

Issue #128, opened 2026-09-05
Runtime wedge inside `apply_exl3_fused_moe` on TP=4 / nnodes=4: GPU busy-spin ~95%, all counters frozen for minutes, self-recovers — py-spy evidence, 2 distinct freeze lines


hushengkai:
Under TP=4 / nnodes=4 the engine intermittently wedges inside the EXL3 fused-MoE path: one scheduling step never returns, `prompt_tokens_total` and `generation_tokens_total` freeze for 90 s to ~15 min while the GPU busy-spins at ~95% util, then the engine recovers by itself and serves normally. Captured 8 times over 2 days with full py-spy sets; **two distinct in-function freeze lines** (`exl3.py:1099` sort region, `exl3.py:1152` stream synchronize). We believe no existing tracker report matches this runtime signature (checked starvation/boot-deadlock/parser-hang families — see "Related but distinct").

## Environment

- Hardware / nodes: 4× DGX Spark (GB10, 128 GB unified memory), TP=4, nnodes=4, via `start-tp4.sh` (1 head + 3 worker containers)
- Interconnect: all four nodes wired **star-topology into a single QSFP 200 Gb switch** (no direct-attach pairs); ConnectX-7 ports (`enp1s0f1np1` / `rocep1s0f1`), RoCEv2; `NCCL_CROSS_NIC=1`; `NCCL_IB_GID_INDEX=3` with `HEAD_GID`/`WORKER_GID` pinned `=3` on every rank (single-subnet fabric, line rate verified)
- Image: `ghcr.io/miaai-lab/glm-5.3-flash-2x-dgx-sparks:exl3` — image ID `09004103c61f`, vLLM `0.1.dev20051+g487ecf187`; pulled 2026-09-04, not rebuilt
- Model (`MODEL` + `MODEL_REVISION`): `brandonmusic/GLM-5.3-Flash-tr3-4bpw` @ `b20c49ba` (4bpw EXL3/TR3 K4; 288 routed experts, top-8, 1 shared; `glm5_next`, 45 layers = 34 KDA + 11 sparse-MLA)
- Served name (`SERVED_MODEL_NAME`): `GLM-5.3-Flash-EXL3`
- How you started the serve: `./start-tp4.sh`
[...]

Issue #121, opened 2026-09-03
Long-form generation (>1K tokens) corrupts mid-stream on rebuilt main image (mojibake, apologizing loops, essay restarts) — clean on pinned GHCR image + 1df71c1-era repo; k/drafter/eager/E2/MNBT all ruled out


Juliuscply:
## Summary

Single-completion **long-form generation goes corrupt mid-stream** on the current `main` (image rebuilt per the recipe stamp, `eb0469f` era): literal mojibake characters, the string `乱码` ("garbled") emitted *as generated text*, English fragments (`meta-commentary`) embedded inside Chinese prose, apologizing loops (`apologizing again... apologizing again...`), the model restarting the essay mid-completion (up to 5 `# title` starts in one output), and 2–4× length overshoot. **Short generations (<~700 tokens) are always clean**; the corruption is position-dependent and appears from roughly 1K+ generated tokens onward. The model itself reacts to it in-context (translated: *"the previous two generation attempts failed (garbled), this time it must be clean"*).

Rolling back to the **pinned GHCR image (`9bb1557a…`) + the pre-E2 repo (`1df71c1` era)** makes it disappear completely (8/8 clean runs today, same prompts, same kit).

## Reproduction matrix (same 2× GB10 kit, same 4000-char Chinese essay prompt ×3 retries per cell, temp 0.7, thinking off)

| image / repo | k | DFlash2 ckpt | extra flags | result |
|---|---|---|---|---|
| rebuilt from `main` (`eb0469f` era) | 3 | `bf582e4` (new) | default | corrupt ×3 (apologizing loops, double starts, outline degeneration) |
| rebuilt from `main` | 7 | `bf582e4` | default | corrupt ×2 (embedded `meta-commentary`, up to 3 starts) |
| rebuilt from `main` | 7 | `7d74cdd8` (old) | `ENFORCE_EAGER=1`, `EXL3_FAT_KERNEL=0`, `MNBT=1024`, no threshold | corrupt ×1, worst (5 starts, hit max_tokens) |
| **GHCR `9bb1557a` + `1df71c1`-era repo** | 7 | `7d74cdd8` | default | **clean ×2** |
| **GHCR `9bb1557a` + `1df71c1`-era repo** | 3 | `7d74cdd8` | + `--long-prefill-token-threshold 1024` | **clean ×6** |

[...]

PR #132, opened 2026-09-07
E3 grouped MoE default, 850k context, client defaults + clock-lock harness

MiaAI-Lab:
Publishes the E3 work that had been sitting on an unpushed local `main`, plus the 850k
context default and this session's doc corrections.

## Runtime

- **E3 grouped fat-expert MoE tier** (`1a0feb0`, `2c0ebe5`): three GPU-driven kernels per MoE
  layer (gather / gate+up+SwiGLU / down+scatter) built from device-side segment tables,
  replacing E2's per-expert host loop. Additive - no upstream exllamav3 kernel is modified.
  Measured **+37-45% cold prefill, TTFT -27-31%**; decode unchanged by construction (E3 never
  runs on decode-sized steps). Numerically indistinguishable from E2 against the `LinearEXL3`
  reference, including real checkpoint experts.
- **`start.sh` default context 900000 -> 850000** (`f4ef41e`): `.env.example` and the README
  said 850k while the launcher fallback still emitted 900k, so an unset var produced a context
  the docs did not describe. `start-tp4.sh` stays at 1M (different kit geometry). `start.sh` is
  outside `overlay_recipe_hash()`'s scope, so the published image is unaffected.

## Docs

- README context default is now 850k across the config table, Context row, headroom paragraph
  and the "do not drop MAX_MODEL_LEN" note. Figures **measured at** 900k (pool receipts, the
  256k/0.87 driver-retry note, concurrency rows) are deliberately left alone - they are
  receipts, not defaults.
- New cold-prefill table measured with [sparkDash](https://github.com/MiaAI-Lab/sparkDash) on
  `glm53-flash-sm121:e3-20260907` at 900k / util 0.86: **1,428-1,587 tok/s across 8k-256k**.
  The 2026-09-06 A/B receipt vs E2 at 500k / 0.84 is kept below it, since it is the source of
  the +37-45% claim.
- KV pool documented as ~1M tokens, with both utils given (989,010 at 0.85; 1,023,626-1,084,615
  at 0.86) and an explicit note that the 850k default is not yet measured.
- Client request defaults documented (`f8eb75f`).

## Tests

- A/B/A harness for the GB10 SM clock lock (`4bac6e2`).
- `tests/test_numeric_config.py` passes with the new launcher defaults, which resolve to
  `MAX_MODEL_LEN=850000 GPU_MEM_UTIL=0.85`.

## Verified before pushing

Image tags consistent: `glm53-flash-sm121:e3-20260907`, `ghcr.io/...:exl3` and
`ghcr.io/...:exl3-e3-20260907` all resolve to `9581c4c74257`; both GHCR manifests identical;
worker carries the same image ID. `overlay/` and `Dockerfile` are clean against HEAD, so the
published image still matches the runtime recipe.

## Known follow-ups (not in this PR)

- `overlay_recipe_hash()` pipes absolute paths through `sha256sum`, whose output includes the
  filename, so the stamp is checkout-path dependent and a fresh clone can never match the
  published image's stamp - meaning cloners rebuild locally instead of using GHCR. It also
  hashes untracked files under `tests/`.
- The 850k default is not yet reflected in a live boot; KV pool at 850k remains unmeasured.

https://claude.ai/code/session_01T5BNDFakn4Pex2NdC3HCq8


hushengkai:
**Field report on 4× DGX Spark TP=4 / nnodes=4: E3 installs and runs cleanly on all ranks, wedges persist (cross-ref #128 — and one new capture lands outside exl3 code), plus one observability regression worth reconsidering.**

Environment: image `09004103c61f` (not rebuilt, per the #121 caution) with E3 files from `main`@`6599585` mounted as overlay: `exl3.py` replaced in-container + `exl3_fat_moe_ext.so` (nvcc 13.0 build) bind-mounted into `dist-packages` on head and all 3 worker containers, `EXL3_FAT_GROUPED=1`, `EXL3_TEMP_ROWS_FUSED=64` (sized to `MAX_NUM_SEQS×(DFLASH_TOKENS+1) = 8×8`). Boot verification on every rank: `exl3 e2 diag schema=2 configured_tier=grouped effective_tier=grouped tier_reason=grouped_ok grouped_eligible=1 cap=12.1 cap_ok=1`; `#125` fine-grained prefix cache canary unchanged (6.6× speedup, 99.05% hit); spec-decode acceptance unchanged (3.8 tok/step); KV pool delta −0.4%.

Two findings from ~3 h of soak incl. repeated 8×60k cold-prefill runs:

1. **The #128 wedge survives the E2→E3 transition — and a full-E3 capture froze inside an attention-layer GEMM dispatch (`__torch_function__`, parameter.py:126), no exl3 frames in the stack.** Same trigger (large prefill + request-completion release rhythm), same signature (GPU busy-spin ~95%, counters frozen, 90 s–6 min self-recovery, RoCE counters flat). Full evidence table and the stack in #128. We no longer read the wedge as an exl3-path bug — but since #132 makes grouped the default tier, flagging that TP4/4-node kits on GB10 will keep hitting it.

2. **Observability regression: E3 logs the `exl3 e2 diag` line once at load time only.** E2 re-emitted it periodically with live counters (`prefill_layer_calls`, `effective_tier`, fallback counts), which made it possible to (a) confirm per-rank tier/code generation at runtime — which is exactly how we caught a half-applied install where head had E3 but workers silently kept E2 — and (b) use the diag heartbeat as an in-band stall probe. With load-time-only logging, neither is possible anymore, and runtime `grouped_calls` cannot be confirmed from logs at all. Request: restore a periodic diag emission (even every N seconds or N forward calls, rate-limited) or expose `exl3_fat_diag()` counters via `/metrics`. Happy to test a patch.

No other functional regressions observed so far; prefill speedup assessment on our kit is pending until the stall pollution is resolved (all 8×60k datapoints are wedge-contaminated).

PR #140, opened 2026-09-08
perf(exl3): accelerate E3 activation and fix default startup budget

RodriMora:
## Description and Motivation

Retarget #140 from the unused E2 helper to the active **E3 grouped kernel**, and correct the launcher memory budget so the advertised 850K configuration starts with graph estimation enabled.

1. Replace the fused E3 epilogue's FP64 exponential with accurate FP32 libdevice `__nv_expf`, explicit rounded arithmetic, and unchanged FP16 boundaries. This is not approximate `__expf` or an E2 fallback. A native image rebuild is required for the kernel.
2. Change the general launcher/template prefill-batch default **7168 → 5120**. Keep **850000 context, utilization 0.85, CG_ESTIMATE=1, CUDA graphs, E3/cap 32, maxseq 4, DFlash2k 7, vision/image 100**. Explicit user overrides still win; existing .env files are not rewritten.

The old 7168 and tested 6144 settings failed cold-start KV-budget initialization. Instrumented repeat profiling did not support discounting the graph reservation. 5120 reduces actual batch-related memory demand; it is an explicit performance/headroom tradeoff, not a claim that the original defaults passed.

## Measurements and reproduction

- **Updated defaults:** four native baseline/candidate A/B/A/B boots passed, first per variant with fresh compilation caches and second warm, using the actual updated start.sh/.env.example and no inference overrides. Warm-pair prefill size-geomean gain **+2.70%**. The larger cold-pair gain is reported separately with baseline-drift caveats; one cold short-document holdout regressed 0.86%.
- **Earlier 7168/CG0 configuration:** native E3 A/B/A/B measured **+3.32% overall**, **+5.72% at 64K**. These remain historical configuration-specific results, not the revised default's measurements.
- The initial 5120 sizing control was 0.73% slower on the size ladder and 10.99% slower on docs-short than the previous 7168/CG0 candidate configuration. This cost is disclosed rather than hidden.

See [default-startup qualification and raw receipts](docs/default-startup-validation.md), [kernel validation and historical results](docs/e3-activation-validation.md), and the checked-in frozen fixtures/portable replay script.

## Related Issues

Addresses #147's E2-only applicability concern for #140. Does not close that broader triage issue.

## Testing

- Both GB10 nodes:582 strict actual-source activation checks each, with/without fast math, including FP16 bits, NaN masks, signed zeros, nondefault stream and changed-input graph replay.
- Both native images on both nodes: existing GPU integration, real-checkpoint parity and graph checks passed with frozen tolerances; no real-checkpoint skips or standalone-module substitution.
- Updated defaults: all 100 fixed fixture requests completed, usage matched across arms, zero prefix-hit deltas, no restarts/OOMs. Minimum sampled MemAvailable **3.31GiB head / 6.78GiB worker**. Memory guard did not trip.
- Vision passed all arms. Cold arms passed 22/25 canaries; warm arms 23/25. Failures were only the existing original-code arithmetic prompt returning 10 instead of 15, also observed upstream and without speculation. Expectations unchanged; runner deliberately exited 2. This remains an unresolved semantic caveat, not a fully-green claim.
- Image-recipe/default consistency and caller-override regressions, shell/Python/Node syntax, portable replay mock tests and git diff --check passed.
- Test ranks stopped; worker staging restored; production settings, weights, fabric, clocks and observability preserved.

## Checklist

- [x] README reviewed and changes consistent with the active E3 path.
- [x] Specific title/description, explicit default change and native-rebuild requirement supplied.
- [x] Reproducible native boots and paired performance measurements supplied.
- [x] **Updated defaults start out of the box on the tested 2x GB10 lane:** four cold/warm boots at 5120/850K/.85/CG1, no inference overrides. The original 7168 setting was not certified; it remains an explicit tuning option requiring separate memory qualification.
- [x] README, validation documents, raw rows and frozen fixtures updated.

Startup qualification is bounded to the tested cluster/configuration. It is not a populated 850K-context, arbitrary-concurrency, hundred-image-workload, or universal-memory guarantee. Existing skip-MM-profiling behavior and the semantic caveat remain documented.


plotarmordev:
Community triage note (full sweep: #147): this optimizes the E1/E2 batched helper `apply_exl3_batched_fat`, but on main (9c0794b) the E3 grouped path uses its own fused `exl3_fat_moe_gather` and never calls that helper — so on a 2x GB10 TP=2 E3 deployment this PR is a no-op. Clean, honestly-measured kernel work; just the wrong tier for E3. Tagged triage/skip for this deployment.

RodriMora:
Retargeted this PR to the active **E3 grouped kernel** in 7dca9b1, replacing the original E2-only proposal. E3 already fuses activation; this revision replaces the FP64 exponential with accurate FP32 libdevice math while preserving the FP16 boundaries.

Native E3 A/B/A/B on two GB10 nodes: **+3.32%** size-ladder throughput geometric mean (**+4.58% / +2.08%** in the two pairs), **+5.72% at 64K**, and positive holdout changes. Both nodes passed actual-source activation tests plus native-image real-checkpoint/graph validation. Raw rows, fixed fixtures, reproduction steps, and limitations are checked in.

The body explicitly retains the CG_ESTIMATE=0 compatibility override, limited headroom, and the identical baseline/candidate arithmetic-canary failure; this is not an untouched-default or fully-green semantic claim.

Please re-evaluate `triage/skip`: the E2-only applicability rationale from #147 no longer describes this revision. The existing label has been left for maintainers/community triagers to reassess.

RodriMora:
Resolved the remaining default-startup checklist item in **40d7169** by explicitly changing the launcher/template prefill batch from **7168 to 5120**. Context remains850000, utilization0.85, and **CG_ESTIMATE=1** with CUDA graphs enabled.

The actual updated defaults passed four native TP2 baseline/candidate boots (fresh and warm compilation caches), all100 fixed fixture requests, zero prefix-hit deltas, identical token usage, and no OOMs/restarts. Minimum sampled host headroom was3.31GiB on the head. Warm-pair E3 throughput gain was+2.70%; cold-pair variability and the batch-size performance tradeoff are documented in docs/default-startup-validation.md with raw receipts.

The box now refers to the **revised, tested defaults**, not the original failing7168 setting. The original-code arithmetic canary remains an explicitly disclosed baseline issue; this is not a fully-green semantic or arbitrary-workload certification. Both test ranks are stopped and staging restored.

plotarmordev:
Thanks for the fast retarget — this is now on the E3 grouped path we actually run, so our earlier "E2-only" note no longer applies; re-tagging P1. Before merge we need to run it ourselves: (1) your `test_exl3_e3_activation.py` plus the existing E3-vs-loop parity check inside our SM121 image (needs a rebuild), (2) a cold-prefill A/B/A at 8K/32K/128K on a 2×GB10 TP=2 kit. Could you trim the committed receipts/fixtures (`docs/*.json`, `tests/fixtures/*.gz`) out of the diff and link them from the PR body instead? We'll post numbers after our next image-build window.

plotarmordev:
**Maintainer verification (GPU, 2×GB10 SM121, image built from 7dca9b1 merged onto current main):**

- `tests/test_exl3_e3_activation.py`: **pass, 0 failed checks** (all cartesian/rounding/clamp diagnostics within tolerance).
- `EXL3_SELFCHECK_GPU=1 test_exl3_overlay.py` with `EXL3_SELFCHECK_REQUIRE_GROUPED=1`: E3 grouped **OK** — synthetic `e3_vs_loop nrmse 0.000229` (identical to the unpatched production image), real-checkpoint parity **ok** (`e3_vs_loop nrmse 1.05e-4`, maxabs 0.00195, `e3_repeat` 0), `glm53 EXL3 overlay verify OK`.

So the numerics change is safe on real expert tensors. Remaining before merge: (1) the prefill A/B/A on this kit to confirm the claimed +3–6% (scheduled for our next maintenance window); (2) please trim the committed receipts/fixtures (`docs/*.json`, `docs/*-validation.md`, `tests/fixtures/e3-prefill.json.gz`, `tests/bench_e3_serving.mjs`) out of the diff and link them from the PR body — the overlay change + the activation test are what we want in tree. Keep main's launcher defaults (7168 MNBT) rather than the PR's 5120.

plotarmordev:
**Prefill A/B/A on our 2×GB10 TP=2 E3 kit** (fresh boot per arm, same `.env`, only the image differs; cold prefill with unique salts, 3 reps each, median prefill tok/s = prompt_tokens / TTFT; 400-token structured decode ×3):

| | A1 main | B (#140) | A2 main | B vs A | A1↔A2 drift |
|---|---:|---:|---:|---:|---:|
| 8K | 1540 | 1547 | 1531 | +0.7% | −0.6% |
| 32K | 1622 | 1643 | 1621 | +1.3% | −0.1% |
| 128K | 1595 | 1611 | 1572 | +1.7% | −1.4% |
| decode | 63.7 | 63.4 | 65.0 | −1.5% | +2.0% |

Result: a small, consistent prefill gain (+0.7…+1.7%) — real in sign at every rung but well below the +3–6% in the PR receipts and within our A/A drift at 128K. Decode unchanged within noise. Numerics were verified earlier today (activation test pass; real-checkpoint E3 parity nrmse 1e-4).

Decision: **accept** once the diff is trimmed to `overlay/exl3_fat_moe.cu` + `tests/test_exl3_e3_activation.py` + the recipe test update (receipts/fixtures out of tree, linked from the body), keeping main's launcher defaults. Please push that and we merge.

plotarmordev:
**Update to our earlier conditional acceptance: defer merge pending performance evidence.**

We repeated the 64K prefill measurement using [TheGrill](https://github.com/plotarmordev/thegrill), our independent measurement/evidence tool. By “standardized” we mean a frozen workload and collector with the same rules across every arm—not an industry standard or a benchmark chosen to reproduce a particular result.

Protocol: collector [34251ca](https://github.com/plotarmordev/thegrill/tree/34251ca0f98f0937c2bf71e5bd8380864ba3660b), extending its [shipped prefill workload](https://github.com/plotarmordev/thegrill/blob/34251ca0f98f0937c2bf71e5bd8380864ba3660b/crates/grill-perf/examples/sparkdash-prefill-v1.json) to a nominal 65,536 tokens (`fill.unit: " the"`, `fill.repeat: 65510`). Concurrency 1; thinking off; exactly 8 output tokens; temperature 0 and top-p 1; `reported-prefix-zero`; fresh boot per stock → patched → stock arm; one discarded warmup plus five measured requests per arm. All 18 requests completed, reported zero cached prompt tokens, and began with answer text. Actual prompt counts were 65,552–65,555 tokens.

| Arm | Median prefill tokens/s | Measured range |
|---|---:|---:|
| Stock A1 | 1,647.63 | 1,643.16–1,653.76 |
| Patched B | 1,652.78 | 1,637.90–1,655.78 |
| Stock A2 | 1,637.25 | 1,631.50–1,647.38 |

The descriptive difference was **+0.52% versus pooled stock**, while stock-to-stock median drift was **−0.63%**. TheGrill's reference comparison withheld the prefill percentage-change claim because the patched range overlaps the combined stock range. This small experiment **does not establish a performance improvement**; it also does not prove the patch cannot help another workload. We did not reproduce the reported +5.72% at 64K under this protocol.

Scope: these results cover the archived verification image built from the earlier `7dca9b1` retarget integrated with our baseline. The PR now points to `40d7169`; this is **not** validation of that newer head. The earlier GPU activation and real-checkpoint parity checks passed for the tested candidate, but those checks alone do not establish a speed benefit or universal numerical equivalence.

We therefore withdraw the earlier “accept once trimmed” recommendation as the current merge decision. Please help identify which workload or conditions explain the gain, ideally with a reproducible TheGrill workload, pinned revisions/settings, and receipts. Author-specific benchmarks remain useful supporting evidence; our acceptance measurement should remain independently defined. We are leaving the PR open pending that evidence, with no further local testing scheduled for now. Thank you for continuing to work through this with us.

RodriMora:
Thanks for the thorough independent verification and for working through the retarget with me.

After reviewing the TheGrill results and re-examining my original receipts, I don’t think this PR currently demonstrates a meaningful end-to-end performance improvement. In particular, the first stock arm in my campaign was still warming, which inflated the reported 64K result. The remaining apparent improvement was not reproduced independently: the standardized run measured +0.52% versus pooled stock with overlapping ranges, comparable to stock-to-stock drift.

The numerical and real-checkpoint checks passed, but that alone does not justify changing arithmetic in a hot production kernel when the performance benefit is unproven. Rather than continue carrying an optimization claim that the evidence does not support, I’m withdrawing and closing this PR.

Thank you again for the careful GPU validation and benchmark work. If a clearly reproducible workload or kernel-level result emerges later, I’ll treat it as a new investigation with fully pinned build provenance and independent measurements.

PR #144, opened 2026-09-09
perf: fuse EXL3 input gather and Hadamard

RodriMora:
Depends on #140. This branch adds one gather/Hadamard commit on top of the activation commit in that PR. It does not replace #140. Until #140 merges, GitHub's diff against `main` includes both commits. Merge #140 first, then rebase this PR if needed to drop the activation commit from the diff.

## What changes

Replace `torch.index_select` followed by `ext.had_r_128` with one Triton kernel in the fat-expert path. The kernel reads the indexed FP16 hidden-state rows directly into the pre-scaled Hadamard-128 transform.

It preserves the FP16 scale-product rounding, native FP32 butterfly order and final FP16 conversion. It also removes the separate `h` scratch buffer, which uses 64 MiB at 8192 rows and hidden width 4096. Routing, packed GEMM, down Hadamard and DFlash stay unchanged.

The new commit changes four files: the helper, its GPU tests, the call site in `exl3.py`, and the Dockerfile copy step. It leaves #140's activation helper unchanged.

## Measurements

I compared this against the activation-only code in #140 on two DGX Spark GB10 nodes. Each value of `EXL3_TEMP_ROWS_FUSED` used an A/B/A/B sequence with fresh deployments. These are incremental gains from input gather/Hadamard fusion, not total gains over the original unfused model.

| `EXL3_TEMP_ROWS_FUSED` | A1 tok/s | B1 tok/s | A2 tok/s | B2 tok/s | Gain across candidate/baseline pairs |
| --- | ---: | ---: | ---: | ---: | --- |
| 128 | 1333.18 | 1383.90 | 1324.52 | 1383.96 | 3.80% to 4.49% |
| 256 | 1339.86 | 1363.56 | 1330.15 | 1369.69 | 1.77% to 2.97% |

The metric is the geometric mean of the 4K, 16K and 64K prefill medians, with three measured repetitions after warmups. I used sparkDash's unchanged prompt builder and streaming request code, identical 25-request fixtures and fresh prefix caches. Every arm reported zero prefix hits.

At cap128, the real-document, code and mixed-text holdout medians improved by 1.58% to 4.40%. At cap256, code and mixed text improved modestly, while short documents ranged from -0.53% to +0.40%. This is not a speedup for every workload.

The isolated gather plus Hadamard operation fell from 3968.55 to 1861.12 microseconds, a 53.10% reduction in the sum of 16 case medians. That is an operation-level result, not a serving-throughput claim.

Both arms ran E2 with TP2, a 131,072-token context and GPU memory utilization set to 0.82. Serving was text-only, with DFlash k7 at draft TP2 and FP8 KV. Both used 7,168-token chunks and at most four sequences. NCCL settings stayed unchanged. The runtime was vLLM `0.1.dev20051+g487ecf187` with the installed M64 E2 kernel. I did not rebuild the native extension. These results do not establish a latest-default E3 or decode speedup.

## Checks

- 64 standalone GPU cases passed exact FP16 bit comparisons against the native gather plus Hadamard sequence, except NaN payloads. The tests check NaN semantics separately. They cover strides, padding, tails, empty inputs, repeated indices, exceptional values, current streams and graph replay with changed inputs.
- Full registered E2/E1, mixed-expert, row-tile, expert-map and CUDA graph checks passed on the tested runtime.
- All 25 token-usage records matched across paired deployments. Generated previews were not identical in every case, including between baseline runs. No bitwise generation-equivalence claim.

The branch merges cleanly with `main` at `9c0794b`, and the merged Python files pass syntax checks. The GPU and serving results above come from the pinned runtime, not that newer build.

Run the standalone test only with inference stopped:

```bash
python3 tests/test_exl3_gather_hadamard.py
```

It requires CUDA, Triton and the installed ExLlamaV3 extension. No NCCL tuning or other experimental kernels are included in this PR.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -437,6 +437,7 @@ RUN set -eux; \
 # rebuild exllamav3_ext. Exl3Config.override_quantization_method requires
 # "exl3" in ModelConfig's ordered overrides list.
 COPY overlay/exl3.py /usr/local/lib/python3.12/dist-packages/vllm/model_executor/layers/quantization/exl3.py
+COPY overlay/exl3_swiglu.py /usr/local/lib/python3.12/dist-packages/vllm/model_executor/layers/quantization/exl3_swiglu.py
 COPY overlay/patch_model_overrides.py /opt/glm53/patch_model_overrides.py
 COPY overlay/qwen3_dflash2.py /opt/glm53/qwen3_dflash2.py
 COPY overlay/dflash2_speculator.py /opt/glm53/dflash2_speculator.py
--- a/Dockerfile
+++ b/Dockerfile
@@ -438,6 +438,7 @@ RUN set -eux; \
 # "exl3" in ModelConfig's ordered overrides list.
 COPY overlay/exl3.py /usr/local/lib/python3.12/dist-packages/vllm/model_executor/layers/quantization/exl3.py
 COPY overlay/exl3_swiglu.py /usr/local/lib/python3.12/dist-packages/vllm/model_executor/layers/quantization/exl3_swiglu.py
+COPY overlay/exl3_gather_hadamard.py /usr/local/lib/python3.12/dist-packages/vllm/model_executor/layers/quantization/exl3_gather_hadamard.py
 COPY overlay/patch_model_overrides.py /opt/glm53/patch_model_overrides.py
 COPY overlay/qwen3_dflash2.py /opt/glm53/qwen3_dflash2.py
 COPY overlay/dflash2_speculator.py /opt/glm53/dflash2_speculator.py
--- a/overlay/exl3.py
+++ b/overlay/exl3.py
@@ -931,9 +931,6 @@ def _fat_scratch(
         "gate_up": torch.empty(
             (capacity, 2 * intermediate), dtype=torch.float32, device=device
         ),
-        "act": torch.empty(
-            (capacity, intermediate), dtype=torch.float32, device=device
-        ),
         "act_h": torch.empty(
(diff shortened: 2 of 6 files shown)


plotarmordev:
Community triage note (full sweep: #147): same call-path caveat as #140 — this stacks on the E2-only `apply_exl3_batched_fat` path (`index_select` + `had_r_128`), which main's E3 grouped path never calls. No effect on a 2x GB10 TP=2 E3 deployment. Tagged triage/skip for this deployment.

plotarmordev:
Thank you — this is clean, honestly-measured kernel work (A/B/A/B fresh-boot numbers, incremental over #140, no overclaiming), and on the tier it targets the fusion is sensible. We're closing it for this deployment for one structural reason: on main, the E3 grouped path runs its own fused `exl3_fat_moe_gather`/`exl3_fat_moe_gateup` kernels and never enters the batched `apply_exl3_batched_fat` helper (`index_select` + `had_r_128`) that this PR optimizes — with `EXL3_FAT_GROUPED=1` (the default here) the change is inert. Nothing wrong with the patch; it's the wrong tier for an E3 kit.

Two fair ways back: (a) retarget to the E3 grouped kernel itself, the way #140 just did — a native E3 A/B with actual-source validation would be reviewable on its own merits; or (b) if the E1/E2 tier is the intent, reopen scoped explicitly to the `EXL3_FAT_GROUPED=0` path with a receipt measured on that tier, so the applicability boundary is stated up front.

PR #80, opened 2026-08-31
scheduler: opt-in warm bypass and deadline eligibility

nood-co1:
## Reviewed source publication — HOLD, not merge-ready

Exact head `ff30124ed456c1ac847e5dacff1b8c0b0ab3dc11` was fast-forward published to the original author branch, preserving history. Integrated base: `a9bbbadc4c7253870d0d519c3d445e4ee7b97540`. This is **older-base source publication**, not current-main integration qualification. Later public main changes, including the TP=3/TP=4 default decisions, must be preserved in a subsequent normal integration; this publication does not overwrite main or establish current default policy.

### Current source scope

Retain the unique warm-bypass and deadline policy as **operator opt-in**, with `GLM53_MIXED_PREFILL_WARM_TOKENS=0` and `GLM53_MIXED_PREFILL_MAX_WAIT_MS=0` by default. Existing skip behavior is unchanged unless explicitly enabled. The deadline makes a request eligible for a late-capped step; it does **not** bound allocation, admission, service time, TTFT or completion. Fair policy is unaffected; the deadline does not override an explicit cap. Caller overrides, including explicit-empty values, use the existing generic mechanism.

`tests/validate_warm_gate.py` is exploratory observation, not runtime admission: capture/peer errors are void, all A/B/C conclusions wait for final peer validity, and void takes precedence over an unmet expectation in the exit status (the unmet evidence remains in the summary). Admission counters do not measure compute service. Exit 0 is not qualification.

### Strict installer migration and recovery requirement

One shared installer authenticates published v1/v2/v5 helper bytes, validates exact helper/gate regions before replacement, requires round-trip consistency and verifies current v6 idempotence. Main independently matched the three registry SHA-256/length entries against public producer source, not only the test fixtures. Drifted, partial, duplicated, decorated and marker-only inputs are refused without a write.

**v3/v4 migration is unsupported and fails closed without rewriting the source.** No authenticated producer for those helper bodies was recovered; method names or invented fixtures do not establish canonical provenance. An operator encountering these markers must restore/reinstall the pinned pristine or an authenticated supported scheduler through an approved maintenance procedure before applying v6; do not erase markers or bypass validation. This is a deliberate narrowed support contract, not a claim that every historical variant migrates.

### CPU evidence and remaining hold

[...]

PR #149, opened 2026-09-09
serve: derive long-prefill threshold and retain opt-in FP8 screening

MiaAI-Lab:
## Conservative source scope — publication under HOLD

Source head `1c552f4d6666cf4f3f5dc83c1b0a522e8e8c4c7c` preserves author history and integrates main `a9bbbadc4c7253870d0d519c3d445e4ee7b97540`. This is **older-base source publication**, not current-main integration or runtime qualification.

- Retain automatic long-prefill threshold derivation on TP=2/TP=4: unset selects half the validated token budget, floored at 1 and capped at 3584; explicit empty keeps stock behavior; explicit values are validated. Preserve caller precedence and rank forwarding.
- **Do not change mixed-prefill defaults:** TP=2 stays `fair`, TP=3 stays `0`, TP=4 stays `skip`. The shared TP=2 `CHUNK=0` proposal remains gated on #180 contention qualification, not landed through this PR.
- **Keep the scoped FP8 fallbacks off** in `start.sh`, `overlay/exl3.py`, and `overlay/patch_dense_fp8.py`. TP=3 is the existing exception: `start-tp3.sh` still defaults to `dense,kda`; this PR does not change that path or claim TP=3 has the measured TP=2 cost.
- Retain projection screening and quality tooling. Shared-top-K conditional KL is a screening proxy with support coverage, not full-distribution quality certification. Historical aggregates below were not rerun or rescored and are not current-head qualification.
- Correct the two Edit-case rubrics: exact normalized relative target, literal replacement result matching the stated reference outcome; candidate code is not executed by that scorer. Reject absent/empty/non-object argument payloads rather than converting them to an empty object; keep explicit JSON-object controls and invalid rescore failures. Other cases remain narrower tool-selection/syntax probes, not proof of arbitrary command/task correctness.

### Verification and remaining ownership

Main supplied passing scoring fixtures and before/after results through actual `run_case`: previously accepted wrong basename, comment-only replacement, missing arguments and empty arguments now fail; valid edit and explicit-object controls remain accepted. Prior scoped panel/threshold/recipe/numeric CPU receipts apply to unchanged paths. This reviewer read the exact scoring receipts and source; no tests or live operations were run in this final wave. Historical scores are not comparable across changed rubrics without a separately declared rescore; stored argument truncation can make old records invalid.

**Maintainer plotarmordev owns the next actions:** reviewed current-main integration and exact-head cumulative CPU gates, then any separately authorized numerical/server-side FP8 prefill-cost admission and the prospective #180 long-prefill/active-decode contention protocol. Production FP8 default-on and TP=2 `CHUNK=0` remain unqualified and excluded from this source candidate. No author rebase or new benchmark is requested now. `merge-hold` remains; no approval or merge is claimed.

Quality scripts are operator-invoked tooling, not an autonomous qualification system: Python `-I` in the code evaluator is not an OS sandbox, and corpus/runtime/receipt identity must be controlled by the maintainer before running candidate-generated code or comparing arms.

<details>
<summary>Original author proposal and historical evidence — preserved, superseded defaults and unrescored results</summary>

## What
- `GLM53_MIXED_PREFILL_CHUNK` default `skip` → `0`. `skip` deferred every new request's prefill until no running request was decoding; under continuous decode arrivals newcomers waited for other users' generations to finish, which serialised coding agents' tool round-trips (each hop is a fresh prefill). `skip` / `N` stay available per deployment.
- `GLM53_DENSE_FP8` default `off` → `dense,kda`. The 2026-09-08 ship left it PROVISIONAL pending a KLD panel; the panel now exists (`scripts/quality/kl_panel.py`, receipts `logs/quality-20260909`): BF16 boot vs FP8 boot, same weights — KL 0.005–0.017 nats/token on 57k tokens of code/prose, flat across 0–32k depth, argmax agreement 95–98 %, tool-call span KL 0.003–0.007 vs a 0.002 same-boot floor, tool-calling eval identical (47/51 both arms). ~2× the noise floor for ~−11 ms/step.
- Adds `scripts/quality/` (KL-by-depth panel, tool-calling + long-context recall eval, HumanEval/MBPP pass@1, arm runner/compare) used for the above.

## Tests
`tests/test_dense_fp8_patch.py tests/test_scheduler_decode_floor.py tests/test_start_overrides.py tests/test_exl3_overlay.py` pass. Live: restarted with both defaults on 2026-09-09.

https://claude.ai/code/session_01UcpeC2Stu6CEVADAj12BD1

</details>

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -106,7 +106,7 @@ DFLASH_TOKENS=7
 DFLASH_DRAFT_TP=2
 MTP_TOKENS=2
 
-# --- decode speed-ups from the 2026-09-08 overnight run (both default OFF) ---------------
+# --- decode speed-ups from the 2026-09-08 overnight run (adaptive-k opt-in; FP8 default ON since 2026-09-09) ---
 # Adaptive verification length (overlay/patch_adaptive_k.py): the drafter keeps its 7-token
 # block, the scheduler verifies only a per-step prefix (2, 4 or 7) chosen from a per-request
 # EMA of accepted draft tokens, batch-uniform so FULL CUDA graphs still run. Lossless at temp 0.
@@ -116,8 +116,10 @@ MTP_TOKENS=2
 # GLM53_ADAPTIVE_K_SET=2,4,7
 # Marlin FP8 weight-only for the KDA + dense-MLP projections (overlay/patch_dense_fp8.py):
 # ~-11 ms per decode step (+10 % on everything, prose +12-19 % alone, +37 % on hard prose
-# stacked with adaptive-k). PROVISIONAL: changes target numerics by FP8 rounding (KL proxy vs
-# stock 0.002-0.013 nats/position; no full KLD panel yet). Groups: shared,dense,kda,mla.
+# stacked with adaptive-k). Default dense,kda since 2026-09-09. Measured cost vs BF16 (KL panel,
+# logs/quality-20260909): 0.005-0.017 nats/token on 57k tokens of code/prose, flat across 0-32k
+# depth, argmax agreement 95-98 %, tool-calling eval unchanged. Set off for BF16.
+# Groups: shared,dense,kda,mla.
 # GLM53_DENSE_FP8=dense,kda
 # The graph list adaptive-k needs, plus a KV-pool cap so FP8's freed GPU memory becomes host
 # headroom instead of a bigger pool (uncapped, the head dropped to ~1.5 GiB MemAvailable at 850k).
@@ -189,8 +191,9 @@ GLM53_BOOT_SHAPE_WARMUP=1
 # Suppress client stop strings until </think> (thinking-on default). 0 = off.
 GLM53_SUPPRESS_STOPS_IN_REASONING=1
 # When a decode seq is running, do not mix a peer prefill into that step
-# (issue #6). skip = decode-only steps; N>0 = cap mixed prefill tokens; 0 = off.
-GLM53_MIXED_PREFILL_CHUNK=skip
+# (issue #6). 0 = off (default since 2026-09-09, skip starved new requests behind
+# running decodes); skip = decode-only steps; N>0 = cap mixed prefill tokens.
+GLM53_MIXED_PREFILL_CHUNK=0
 # Sparse-indexer prefill gather workspace sizing.
 #   stock     = max_model_len * 40 entries (5036.40 MB locked at 1M, measured)
 #   rightsize = the legal per-step maximum, ~+26% KV cache. Opt-in.
--- a/README.md
+++ b/README.md
@@ -621,7 +621,7 @@ that are now documented/enforced:
 | `ABLIT` | `0` (off) | opt-in. `1` = apply o_proj edit at load (both ranks). Unset = stock weights |
 | `GLM53_ADAPTIVE_K` | `off` | `ema` = adaptive verification length (prose +13–21 %); needs the capture-size list in `EXTRA_ARGS`. See *Faster prose decode* |
 | `GLM53_ADAPTIVE_K_SET` | `2,4,7` | candidate draft lengths; graphs are captured for each length + 1 |
-| `GLM53_DENSE_FP8` | `off` | `dense,kda` = FP8 weight-only (Marlin) dense projections, ~-11 ms/step; PROVISIONAL numerics. Groups: `shared,dense,kda,mla` |
+| `GLM53_DENSE_FP8` | `dense,kda` | FP8 weight-only (Marlin) dense + KDA projections, ~-11 ms/step (default since 2026-09-09). Measured cost vs BF16: KL 0.005-0.017 nats/token on 57k tokens of code/prose, flat across 0-32k depth, argmax agreement 95-98 %, tool-calling eval unchanged (`logs/quality-20260909`). `off` = BF16. Groups: `shared,dense,kda,mla` |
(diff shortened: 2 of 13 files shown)


plotarmordev:
Tested the current PR head with FP8 off, FP8 on, a repeat off run, and FP8 on with a smaller token budget. All 36 measured benchmark runs completed.

Median time in seconds, from three measured trials per scenario. Lower is better.

| Measurement | FP8 off | FP8 off, repeat | FP8 on | FP8 on, smaller budget |
| --- | ---: | ---: | ---: | ---: |
| Generate 400 tokens, single request | 6.487 | 6.451 | 5.743 | 5.740 |
| First output from a cold ~8K-token prompt | 5.233 | 5.252 | 6.071 | 6.438 |
| Finish decoding with an overlapping prompt request | 11.752 | 11.553 | 11.814 | 11.899 |

At the regular budget, FP8 reduced solo-decode time by about 11%, but increased cold-prompt first-output latency by about 16%, compared with the average of the two off-run medians. Mixed-request decode completion stayed roughly the same. The smaller-budget setup (2048, with threshold 1024) also started and completed the tests.

These are small local tests—not a full quality check of the standard model. The small numerical check also showed more change with FP8 than between the two FP8-off runs, so this is a speed/precision tradeoff, not identical output.

My recommendation: merge with FP8 still opt-in for now, while keeping the scheduler and automatic threshold changes. The faster decode is useful, but the slower cold-prompt handling and extra numerical drift make me cautious about enabling it for everyone. If your larger quality results support default-on and you are happy with the latency tradeoff, that is also a reasonable choice.

Would you prefer the conservative merge with FP8 opt-in, or do you want to keep FP8 enabled by default based on your broader results?

plotarmordev:
Following up after our discussion: since avoiding a prefill slowdown is important, I recommend holding the FP8 default-on change until this is checked.

In our small test, a cold ~8K-token prompt took about 5.24 seconds to first output with FP8 off, versus 6.07 seconds with it on—about 16% longer. The repeated off run was consistent with the first. This does not establish the cause or mean every workload will see the same slowdown.

If the slowdown is confirmed, my preference is to keep FP8 opt-in and retain the scheduler and automatic threshold changes. If it can be fixed or shown to be specific to our setup, we can reconsider default-on.

Leaving this with you for the prefill check; we will move on to the other PRs.

plotarmordev:
### Maintainer split decision — `7a5ca48e8efe66c6245157b4ff5e10e0829ae7a7`

We select the **conservative, maintainer-owned split**: retain the real FP8 opt-in implementation, automatic long-prefill threshold derivation/validation and quality tooling; keep the FP8 fallbacks **`off` at all three sites** (`start.sh`, `overlay/exl3.py`, `overlay/patch_dense_fp8.py`) and make the documentation describe that scope. This supersedes leaving the split decision or an immediate new prefill run with the author.

`CHUNK=0` is the selected scheduler direction, **not an independently cleared production default**. Carry it through a single integration with #180's validator/canary/docs and #80's explicitly opt-in warm/deadline policy; it must not land via this split before #180's still-unmet contention qualification. Retain the unique threshold and quality-tool work without duplicating scheduler implementation. If a smaller independent landing is prepared, it must retain current scheduler defaults until that gate clears.

Do **not** restore `test_documented_defaults` or `test_recipe_stamp_wiring`: the deleted source-text/default-pinning checks are not an admission condition. Preserve useful behavior checks and real opt-in code, not those implementation assertions.

TP3 already has different defaults (`dense,kda` and `CHUNK=0`); the TP2 hold does not imply FP8 is off across every topology. The reported roughly 16% cold-prefill cost is evidence from the measured TP2 setup, **not a measured or necessarily identical TP3 cost**. No topology-wide default change or new qualification is made here.

**Next-action owner: maintainer (plotarmordev)** for the split, shared scheduler integration, and validation schedule. Unblock only after exact-head review/verification of the conservative scope and, for any production `CHUNK=0` landing, the declared #180 contention gate. FP8 default-on remains held for a separate maintainer-adjudicated latency/quality admission decision. No immediate author run is requested; the current hold remains.

plotarmordev (commented):
Source-only review at 1c552f4d6666cf4f3f5dc83c1b0a522e8e8c4c7c: exact Edit-reference and invalid-argument corrections accepted against Main CPU receipts; original branch fast-forward and conservative description verified. Maintainer plotarmordev owns current-main integration and cumulative gates, numerical/server-side FP8 admission, and #180 contention qualification; historical scores were not rescored. TP=2 fair / TP=3 off / TP=4 skip mixed-prefill defaults and the existing TP=3 dense,kda exception are preserved. HOLD remains; no approval, merge, live qualification or immediate author benchmark request.

plotarmordev (commented):
## Hold update

Public head checked: `1c552f4d6666cf4f3f5dc83c1b0a522e8e8c4c7c`.

**Why it is held:** The conservative split still needs final integration and qualification. It must not introduce a new scheduler default before #180's contention gate passes.

**What we tested:** The CPU scoring, threshold and comparison checks passed. We did not run the full generated-code panel: `scripts/quality/code_eval.py` executes generated code in a host subprocess, and Python `-I` is not a sandbox. The related scheduler contention run did not establish progress for the waiting request. These are separate limits, not a claim that all of this PR failed.

**Next action — me, @plotarmordev:** Choose an isolated evaluation boundary, finish the shared scheduler integration, and define the remaining numerical/quality checks. Retain FP8 opt-in and current topology-specific defaults unless their separate gates pass.

**Author action now:** None. Do not run generated code on a serving host or repeat benchmarks to satisfy this hold. I own the next integration and test-plan work.

PR #153, opened 2026-09-10
fix: scope shard completeness to the active snapshot and follow valid blob links

plotarmordev:
## What
Count direct safetensor files in the active target snapshot, following valid Hugging Face blob symlinks while excluding dangling links.

## Why
Stale snapshots previously inflated the completeness count even when the snapshot selected for loading was incomplete. The submitted snapshot-scoping change also missed valid cache symlinks until the retained rework enabled link following.

## Verification
The behavioral snapshot regression passed for stale shards outside the active snapshot, an empty repository, a valid blob link, and a dangling blob link. These were temporary local filesystem fixtures, not model loads.

This remains a shard-count completeness check, not a manifest or content-integrity check. Missing refs retain the existing latest-snapshot fallback convention. No model loading or remote cache synchronization ran.

### DIFF
--- a/start.sh
+++ b/start.sh
@@ -429,7 +429,7 @@ count_shards() {
         printf '0'
         return
     fi
-    find "$repo_path/snapshots/$ref" -maxdepth 1 -type f -name '*.safetensors' 2>/dev/null \
+    find -L "$repo_path/snapshots/$ref" -maxdepth 1 -type f -name '*.safetensors' 2>/dev/null \
         | wc -l | tr -d '[:space:]' || true
 }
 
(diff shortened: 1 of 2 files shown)


plotarmordev:
**Maintainer merge note.** Reason: real bug — `count_shards` counted `*.safetensors` across all snapshots, so a stale snapshot could mask an incomplete active one (fixture: main counts 120 stale shards, fix counts 1). The submitted `-type f` missed valid HF blob symlinks (counted 0 on a real cache); our follow-up uses `find -L`, with a symlink + dangling-link fixture that fails before the fix and passes after.

Tests run on the rebased branch: `test_snapshot_scoped_count.py`, `test_bringup_robustness.py`, `test_start_overrides.py`, `test_numeric_config.py`, `test_warm_restart_stdout.py`, `test_image_recipe.py`, `bash -n start.sh` — all pass. Pre-existing upstream failure `test_indexer_workspace.py:788` (pins `stock`, main defaults `rightsize`) is unrelated and fails identically on pristine main; a separate fix follows. Note: `tests/` changes move the recipe stamp — next `start.sh` without `SKIP_BUILD=1` rebuilds the image once.

PR #154, opened 2026-09-10
fix: keep API credentials out of launcher argv and off the worker

plotarmordev:
## What
Pass the API credential through the head Docker process environment with a value-less environment option, and stop forwarding it to the headless worker.

## Why
The old launcher expanded the credential into both local Docker arguments and the remote worker command. The worker does not serve the API and does not need the credential.

## Verification
The behavioral launcher regression passed with configured and empty authentication. Isolated executable recorders observed head environment inheritance, credential-free arguments and output, and no worker credential propagation. No actual Docker or SSH process ran.

The credential remains available in the head process/container environment to suitably privileged readers. This change protects argv, not environment inspection or shell tracing. Live bearer-auth behavior was not exercised.

### DIFF
--- a/start.sh
+++ b/start.sh
@@ -1300,12 +1300,8 @@ launch_cluster() {
              GLM53_ADAPTIVE_K_MIN_STEPS GLM53_ADAPTIVE_K_SATURATE GLM53_ADAPTIVE_K_HIST GLM53_DENSE_FP8; do
         serve_env+=" -e $v='${!v:-}'"
     done
-    # VLLM_API_KEY is read by the head (rank 0) API server for bearer auth; the
-    # worker runs --headless so it only needs the var for argv-parity, and
-    # start.sh below passes it explicitly on the head. Keep it out of the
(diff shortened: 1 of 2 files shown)


plotarmordev:
**Maintainer merge note.** Reason: `VLLM_API_KEY` was expanded into the worker ssh command and the head `docker run` argv (visible in process listings on both nodes); the headless worker never needs it. Production hunks kept as submitted; the source-string test was replaced by a behavioral test that runs the real `launch_cluster` with stubbed docker/ssh recorders and asserts the value is absent from all argv and reaches only the head container env.

Tests on the rebased branch: `test_api_key_argv.py`, `test_start_overrides.py`, `test_bringup_robustness.py`, `test_numeric_config.py`, `test_warm_restart_stdout.py`, `test_image_recipe.py`, `bash -n` — pass. Same pre-existing `test_indexer_workspace.py:788` upstream failure, unrelated. Recipe stamp moves (tests/ changed).

PR #155, opened 2026-09-10
fix: pin drafter selection without losing explicit revision overrides

plotarmordev:
## What
Select the published exact drafter revision for download, resolution, and synchronization. Preserve caller overrides and explicit empty values across configuration loading. Reject missing pinned weights instead of falling back silently to cached main.

## Why
Mutable Hub main and a repo-wide draft-file check could select different weights on fresh and warm installs. Pinning must also control the selected snapshot and sync marker, not just the download command.

## Verification
The behavioral revision regression passed for stale main, missing pinned files and directories, caller precedence, explicit empty configuration, and an isolated HF download recorder. No Hub download or rank launch ran.

An unset revision now chooses the published pin. A custom drafter needs its own matching exact revision or an explicit empty revision. Empty follows cached main; refresh is needed to update it. Skipping downloads with a missing pin fails closed. Worker contents under deliberately skipped sync and live generation were not verified.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -101,6 +101,7 @@ SERVED_MODEL_NAME=GLM-5.3-Flash-EXL3
 SPEC_METHOD=dflash
 DFLASH_MODEL=incoai/GLM-5.3-Flash-DFlash2
 # Exact checkpoint behind the published 2026-08-30 TP=2 receipts.
+# Override with an exact commit; empty follows the cached main ref (refresh to update).
 DFLASH_REVISION=dc77ff1c99eeb2df044ee3d4f0094eb033fee410
 DFLASH_TOKENS=7
 # Drafter TP. 2 = measured keep on this 2× GB10 kit (shard draft across ranks).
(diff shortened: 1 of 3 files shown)


plotarmordev:
**Maintainer merge note.** Reason: `download_dflash` ran `hf download` with no `--revision`, and the Hub shipped three different drafter weights under one ID in four days — a re-sync could silently change decode behavior. Pin defaults to `dc77ff1c…`, the revision behind the published TP=2 receipts. Four defects fixed on top of the submission: silent fallback to cached main when the pinned snapshot is absent (now fails closed), `DFLASH_REVISION` missing from caller-export capture, `${VAR:-}` swallowing an explicit empty (now `${VAR-}` = unpinned as documented), and a dead `have=1` completion check. Static test replaced with behavioral resolver/override tests; the new tests fail on the submitted code and pass after.

Tests on the rebased branch: `test_dflash_revision_pin.py`, `test_start_overrides.py`, `test_bringup_robustness.py`, `test_numeric_config.py`, `test_warm_restart_stdout.py`, `test_image_recipe.py`, `bash -n` — pass. Same unrelated `test_indexer_workspace.py:788` upstream failure. Recipe stamp moves.

PR #156, opened 2026-09-10
fix: port upstream GLM chat-template corrections

plotarmordev:
## What and why

Ports the chat-template corrections from zai-org/GLM-5.3-Flash commit 690b705278a3a58e538fcb37c2ca8b5f9511213c. Assistant history with null content no longer emits literal `None`. Tool-name concatenation uses Jinja string coercion, and result-order validation stops when sorting is already impossible. Existing thinking controls and parser delimiters are preserved.

All 6 changed hunk bodies match upstream exactly. The change is limited to files/chat_template.jinja. This is the rebased maintainer replacement for #127; contributor attribution is retained.

## Verification

Reused the standalone upstream-hunk/render comparison: all assertions passed, including null-content reproduction, unchanged marker sequences, and 8 invalid-ID fallback cases. The existing chat-template test file passed all 7 tests.

The host-side tests directory run reported 62 passed, 2 live checks deselected, and 1 existing unrelated failure: the indexer recipe test pins an obsolete default. Its test and launcher are unchanged from the base. No test was weakened or removed. GPU, installed-image, and serving checks were explicitly excluded from this local run.

## Limits and release gate

The live tool/reasoning smoke is a separate orchestrator gate; attach its reviewed evidence before merge. Local rendering does not establish streaming parser behavior, generated parallel-call reliability, or installed-image behavior. No performance improvement is claimed.

The template belongs to the image recipe-stamp inputs, so a normal post-merge launch may rebuild on stamp drift. A candidate read-only template mount can isolate the live smoke from image rebuilding, but does not verify the rebuild path. No image build was performed in this review.

### DIFF
--- a/files/chat_template.jinja
+++ b/files/chat_template.jinja
@@ -73,8 +73,8 @@ For each function call, output the function name and arguments within the follow
                 {{- emit_audio() -}}
             {%- endif -%}
         {%- endfor -%}
-    {%- else -%}
-        {{- content }}
+    {%- elif content is not none -%}
+        {{- content -}}
     {%- endif -%}
 {%- endmacro -%}
 {%- macro tool_response(text) -%}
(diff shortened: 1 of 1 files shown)


plotarmordev:
**Maintainer merge note.** Reason: faithful port of the upstream zai-org chat-template commit `690b705` — all 6 hunks byte-identical to upstream; fixes literal `None` leaking into assistant history when `content` is null, `~` tool-name concatenation, and early breaks in the tool-result ordering loops. Rebased over #150 (thinking-off effort gate) with the template tests re-run.

Verification: template unit tests + 8 ordering-fallback cases pass; CPU render harness shows the only rendered delta is the removed `None` bytes with all parser markers unchanged. **Live smoke on a real 2×GB10 TP=2 boot with the candidate template** (bytes verified inside the head container): 7/7 requests pass — plain thinking on/off → `READY`; parallel tool generation → two `lookup` calls with distinct IDs and JSON args; tool continuation with empty and with null assistant content → `ALPHA BETA`, `finish_reason: stop`, no raw XML, no literal `None`. `files/` change moves the recipe stamp.

PR #157, opened 2026-09-10
fix(launcher): make long prefill threshold an explicit opt-in

plotarmordev:
## What and why

Reworks #112 as an explicit scheduler opt-in. `LONG_PREFILL_TOKEN_THRESHOLD` defaults to empty, which emits no flag and preserves stock scheduling. A positive decimal value is validated against `MAX_NUM_BATCHED_TOKENS`, captured before `.env`, and forwarded to both ranks. An explicitly empty caller export overrides a configured value.

The example documents `3584` for `MAX_NUM_BATCHED_TOKENS=7168` (MNBT/2), not a launcher default of `1024`. The supplied A/B/A observation was warm-session freeze `131/108 s` stock to `15 s`, contended prefill `+6%`, and solo prefill `+3..+11%` within baseline drift. These are kit-specific observations, not a universal performance guarantee.

## Verification

Host regression renders both rank scripts: the flag appears exactly once per rank when set and never when unset or empty. Validation and caller-empty precedence are covered. Scoped suite: `20 passed`. Full `tests/`: `76 passed`, with the pre-existing upstream `test_indexer_workspace.py:788` assertion failing because it still expects `stock` while upstream defaults to `rightsize`. No tests or unrelated defaults were changed to hide that failure. `bash -n start.sh` passed.

## Limits

Default behavior is unchanged. No image build, cluster restart, or serving request was performed for this rework. Owner GO depends on reviewing the orchestrator's `PR-112-solo-reps.json` and accepting the known upstream test defect; the package alone does not certify a new live performance result.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -140,6 +140,12 @@ MAX_NUM_SEQS=4
 # throughput/KV balance. Re-run 2048 vs 7168 before changing; never 8192
 # (indexer smem).
 MAX_NUM_BATCHED_TOKENS=7168
+# Empty = stock scheduler (no --long-prefill-token-threshold flag).
+# Measured A/B/A at MNBT=7168 with LONG_PREFILL_TOKEN_THRESHOLD=3584 (MNBT/2):
+# warm-session freeze 131/108 s stock -> 15 s; contended prefill +6%;
+# solo prefill +3..+11%, within baseline drift. Re-measure for other kits.
+# Positive integer <= MAX_NUM_BATCHED_TOKENS, or empty.
+LONG_PREFILL_TOKEN_THRESHOLD=
 GPU_MEM_UTIL=0.85
 KV_CACHE_DTYPE=fp8
 # 0 = load vision tower (image + video). 1 = language-only.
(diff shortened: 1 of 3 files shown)


plotarmordev:
**Maintainer merge note.** Reason: one uncached long prefill starves every other session for its whole duration. Verified in the image's scheduler (`long_prefill_token_threshold` clamps `num_new_tokens` in both the running and admission loops). Reworked from the submission: the launcher default is now **empty** (stock scheduler, no flag) rather than 1024 — 1024 is wrong for MNBT=7168 kits per the author's own addendum — with `LONG_PREFILL_TOKEN_THRESHOLD=3584` (= MNBT/2) documented as the measured opt-in. Wired to both ranks like `EXTRA_ARGS`, caller-export capture, validated as positive int ≤ MNBT or empty; behavioral test renders both rank scripts with the knob set/unset.

Live A/B/A on a 2×GB10 TP=2 E3 kit (850k ctx, DFlash2 k=7, fresh boot per arm, both ranks' argv verified, ~104K-token prompts, unique salts): warm-session max poll freeze during a concurrent cold prefill **131 s / 108 s stock → 15 s** with the flag; contended cold prefill +6%; solo cold prefill **+3.9%** (n=3 per arm: 68.1/65.4/65.2 s stock vs 71.5/68.0/67.5 s flag); the waiting session keeps 103,936 cached tokens with the flag vs 0 stock; zero log errors/preemptions. Host tests on the rebased branch: `test_start_overrides.py`, `test_bringup_robustness.py`, `test_numeric_config.py`, `bash -n` — pass. Credit to @jdecker76 for the finding and the same-kit measurements.

PR #158, opened 2026-09-10
feat(launcher): add opt-in default reasoning effort

plotarmordev:
## What and why

Reworks #87 as an opt-in `GLM53_DEFAULT_REASONING_EFFORT` launcher setting. Empty remains the default and sends no flag. `low`, `high`, and `max` are accepted; `medium` is rejected because the existing template maps it to Max rather than a distinct effort level.

Both ranks receive `--default-chat-template-kwargs` with a single JSON argument. Caller exports, including an explicit empty value, take precedence over `.env`. Per-request `chat_template_kwargs.reasoning_effort` remains the override mechanism. The template itself is unchanged: omitted effort maps to Max.

Removed repeated explanatory prose, pasted receipt documentation, and the blanket recommendation from the original PR. There is no new launcher helper or wrapper. The existing PR tests are retained rather than weakened to accommodate the rebase.

## Verification

The original shell regression passed: enum rejection, both-rank exact-once argv, and empty/no-flag behavior. Scoped Python tests: `16 passed`. An offline render smoke confirmed omitted/low/high/max/medium map to Max/Low/High/Max/Max. Full `tests/`: `65 passed`, with the pre-existing upstream `test_indexer_workspace.py:788` assertion failing because it expects the old stock default. Shell syntax checks passed. No upstream test or default was changed to conceal that failure.

## Limits

No image build, serving request, or restart was performed for this rework. Owner GO requires the orchestrator's `PR-87-live-ab.json`, using the frozen `PR-87-prompts.json` set, to distinguish effort-omitted requests from the explicit-max control and document tokens, wall time, correctness, and render receipts. Host checks do not establish a performance or quality improvement, or prove the deployed vLLM request-override merge order.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -200,6 +200,11 @@ GLM53_SUPPRESS_STOPS_IN_REASONING=1
 # When a decode seq is running, do not mix a peer prefill into that step
 # (issue #6). skip = decode-only steps; N>0 = cap mixed prefill tokens; 0 = off.
 GLM53_MIXED_PREFILL_CHUNK=skip
+# Server default: empty (no flag), low, high, or max. Empty preserves the
+# template's Max fallback for omitted effort; medium is not a template level.
+# Example opt-in: GLM53_DEFAULT_REASONING_EFFORT=high
+# Per-request chat_template_kwargs.reasoning_effort overrides this default.
+GLM53_DEFAULT_REASONING_EFFORT=
 # Sparse-indexer prefill gather workspace sizing.
 #   stock     = max_model_len * 40 entries (5036.40 MB locked at 1M, measured)
 #   rightsize = the legal per-step maximum, ~+26% KV cache. Opt-in.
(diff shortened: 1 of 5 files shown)


plotarmordev:
**Maintainer merge note.** Reason: the chat template maps an omitted `reasoning_effort` to `max`, so clients that don't set it burn the maximum reasoning budget. This adds an opt-in server default (`GLM53_DEFAULT_REASONING_EFFORT`, empty = unchanged) wired via `--default-chat-template-kwargs` on both ranks; per-request `chat_template_kwargs` still overrides. Trimmed from the submission (467 → 220 lines); enum guard rejects `medium` because the template does not recognise it.

Live A/B on a 2×GB10 TP=2 kit (fresh boots, argv verified on both ranks), 3 frozen prompts, temperature 0, cap 8192: with the server default `high`, effort-omitted completions dropped **742 → 383, 8192 (cap) → 1465, 188 → 57 tokens**; the explicit `reasoning_effort: max` control was unchanged within noise (751/8192/186 vs 709/8192/188) — the default only affects requests that omit the field. Host tests on the rebased branch: `test_default_reasoning_effort.sh`, `test_start_overrides.py`, `test_bringup_robustness.py`, `bash -n` — pass. Credit to @nood-co1 for the mechanism and the original knob.

PR #161, opened 2026-09-10
launcher: caller exports win over .env for every key

plotarmordev:
## What
Caller-exported environment variables now win over `.env` for every key, including explicit empty values, by snapshotting exported names before sourcing `.env` and restoring them after. Replaces ~30 per-key `_cli_*` captures with one rule.

## Why
The hand-maintained capture list silently let `.env` override caller exports for ~30 knobs (DFLASH_TOKENS, CG_ESTIMATE, LIMIT_MM, ...). Setness-aware semantics from #155/#157/#158 are preserved.

## Verification
Host tests on the rebased branch: test_start_overrides (extended with unset/empty/value cases through the real preamble), test_dflash_revision_pin, test_api_key_argv, test_snapshot_scoped_count, test_numeric_config, test_bringup_robustness, test_warm_restart_stdout, test_image_recipe, test_chat_template, test_indexer_workspace, test_default_reasoning_effort.sh, bash -n. VLLM_API_KEY handling from #154 untouched. Credit to @nood-co1 for the original PR; source-text assertions were replaced with behavioral checks.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,5 +1,7 @@
 # GLM-5.3-Flash EXL3 — 2× DGX Spark (GB10 / SM121)
 # Copy to .env and edit. ./start.sh does this for you on first run.
+# Any key set here can be overridden per invocation (KEY=value ./start.sh
+# restart): a non-empty caller export wins over this file (#91).
 
 # --- nodes (edit these) ------------------------------------------------
 HEAD_IP=10.0.0.1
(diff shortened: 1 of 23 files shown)


plotarmordev:
**Maintainer merge note.** Reason: the hand-maintained `_cli_*` capture list silently let `.env` override caller exports for ~30 launcher knobs; one general rule (snapshot exported names before sourcing `.env`, restore after, readonly skipped) replaces it while keeping the explicit-empty semantics we added in #155/#157/#158. Trimmed the submission from +344 to 95/−98 by dropping the duplicated harness and restated docs; obsolete source-text pins in two tests replaced with behavioral checks through the real preamble. Tests on the rebased branch: test_start_overrides (extended), test_dflash_revision_pin, test_api_key_argv, test_snapshot_scoped_count, test_numeric_config, test_bringup_robustness, test_warm_restart_stdout, test_image_recipe, test_chat_template, test_indexer_workspace (21), test_default_reasoning_effort.sh, `bash -n` — all pass. No boot required (launcher preamble only); recipe stamp moves (tests/ changed).

Issue #166, opened 2026-09-11
~30s Time-To-First-Token on 2x DGX Spark, independent of load/concurrency


rtdean93:
# Bug report: ~30s Time-To-First-Token on 2x DGX Spark, independent of load

## Summary

On a fresh 2-node deployment (repo commit `1caea9a`), every request — even a
single request sent in isolation with no concurrent traffic — takes
**~29-37 seconds to return its first token**. Steady-state decode speed once
generation starts is reasonable (~16 tok/s/stream, ~62ms mean TPOT). The
~30s delay appears to be a fixed per-request tax, not a throughput/queueing
problem. Ruled out network and burst-arrival as causes (details below).
Model output is correct; the server is stable (0 failures across 52 bench
requests). This is purely a latency/root-cause question.

## Environment

- 2x NVIDIA DGX Spark (GB10), 121GiB unified memory each
- Direct ConnectX-7 QSFP link between nodes, port `enp1s0f0np0` /
  RoCE device `rocep1s0f0` on **both** nodes (symmetric — this kit's cabling
  differs from the repo's documented default of head=f1/worker=f0)
- GID index 3 on both nodes (verified: carries each node's own `::ffff:<ip>`)
- Docker 29.2.1, driver 580.173.02, CUDA 13.0
- Repo commit: `1caea9a10b26ae93b88d08e82d1e7abb0dc45a42`
- Image: rebuilt locally (GHCR pull's recipe stamp didn't match this repo
  commit — `825e33746403` vs `a53970c3f52f` — so `start.sh` rebuilt from
  Dockerfile automatically)

## Config (non-default values in `.env`)

```
HEAD_CX7_IF=enp1s0f0np0        # repo default: enp1s0f1np1
WORKER_CX7_IF=enp1s0f0np0      # matches repo default
HEAD_CX7_IB=rocep1s0f0         # repo default: rocep1s0f1
WORKER_CX7_IB=rocep1s0f0       # matches repo default
NCCL_IB_GID_INDEX=3            # matches repo default
MAX_MODEL_LEN=640000           # lowered from 850000 default — see below
ABLIT=1                        # abliteration on (load-time o_proj edit, method=auto/proj since ablit/transplant/ wasn't populated)
```
Everything else is stock defaults from `.env.example` (`SPEC_METHOD=dflash`,
[...]

Issue #162, opened 2026-09-10
reasoning_effort=max: single turn can burn the entire output budget on thinking → finish_reason=length, empty agentic turn


Squabble9:
## Summary

At `reasoning_effort=max`, a single assistant turn can spend the **entire output budget on the thinking block** and terminate with `finish_reason=length` — returning empty or mid-sentence content. In agentic harnesses this ends the turn (and usually the task) silently: the framework sees the assistant stop calling tools and treats the run as complete, but no deliverable was ever produced.

## Setup

- Recipe `1de5c41` (current `main`), TP=2 on 2× DGX Spark, GLM-5.3-Flash-EXL3 (EXL3), default recipe serving args (thinking on; `reasoning_effort` unset server-side → template default).
- Client: OpenCode (non-interactive agent mode) over the OpenAI-compatible API, `reasoning_effort: max` sent per-request, no explicit client `max_tokens`.

## Observed (reproduced twice in one benchmark session)

Two of seven `max`-effort cells in a small task benchmark died this way:

1. **Screenshot→HTML reproduction task** (one-shot, ~2.4k image tokens + short prompt): one turn, `finish_reason=length`, **32,000 completion tokens**, zero visible text, no file written.
2. **Multi-step reconciliation-report agent run**: 13 successful tool calls (reads/greps/`bun` scans), then one reasoning turn hit exactly **32,000 completion tokens**, `finish_reason=length`, final visible text cut mid-plan ("Now step 4: run the scanner."), report never written.

In both cases the 32,000 output tokens were consumed entirely inside the thinking block (nothing actionable emitted). The truncation point at exactly 32,000 in both cases suggests a hard server-side output ceiling interacting with unbounded `max`-effort thinking.

The same tasks at `reasoning_effort=low` completed correctly in ~15 minutes.

## Expected

Thinking should be budgeted so that a usable response always fits the output cap — e.g.:

- clamp thinking to reserve N tokens for content, or
- expose a thinking-budget knob (the DeepSeek-v4 DSpark recipe lineage has the analogous `thinking_token_budget` hotfix, its issue #31), or
- on `finish_reason=length` with all-thinking output, retry/degrade the turn at a lower effort instead of returning an empty turn.

At minimum, documenting that `max` can starve content on long prompts — and that clients should treat `finish_reason=length` with empty content as retriable at lower effort — would save agentic users a debugging round.

## Workaround

[...]

Issue #147, opened 2026-09-09
Open-PR triage 2026-09-09 (community, non-maintainer)


plotarmordev:
Independent community triage of all 39 open PRs, checked against main at `9c0794b` (merge of #139, adaptive-k + FP8 dense decode). This is an independent sweep, not a maintainer decision.

**Method:** read every PR body, diff, and review thread; grep each PR's knob or fix against main to determine merged / superseded / conflicting status. Key verifications on main: no fine-grained APC code exists; the mixed-prefill gate is still v1; `download_dflash` calls `hf download` with no `--revision`; `count_shards` still counts across snapshots; `VLLM_API_KEY` is still interpolated into worker ssh argv (start.sh:1300) and head docker argv (start.sh:1408); the E3 grouped path uses its own `exl3_fat_moe_gather` and never calls `apply_exl3_batched_fat` (the E2 helper #140/#144 optimize); chat template line 7 still maps unset `reasoning_effort` to `max`. No CI status checks are configured on any PR (statusCheckRollup empty repo-wide), so "mergeable" below means git-mergeable only.

| PR | Title | Tag | Runtime path | One-line reason |
|---|---|---|---|---|
| #67 | Pin DFlash2 drafter revision | **P0** | boot / spec-decode weights | Silent drafter weight drift is live in main (`hf download` without `--revision`) — a fresh sync can change decode behavior overnight; mergeable. |
| #66 | Keep VLLM_API_KEY out of argv and off worker | **P0** | boot / security | Still leaking in main 9c0794b (start.sh:1300, :1408) — key readable from `/proc/<pid>/cmdline`; cheap safety fix; mergeable. |
| #112 | LONG_PREFILL_TOKEN_THRESHOLD (440s freeze -> 20s) | **P0** | scheduler / prefill fairness | One long prefill starves all other sessions for its whole duration at long context lengths; testable today via EXTRA_ARGS without merging; conflicting (start.sh). |
| #83 | Per-KV-group APC retention (sparse drafter) | **P1** | prefix cache / KV scheduler | Fixes mutual eviction of two >=56K conversations (0% hits, 60-90s re-reads) — a real long-context stability defect; conflicting. |
[...]

PR #130, opened 2026-09-05
APC: opt-in per-group retention, draft eviction priority, and safe DFlash replay

ratulsarna:
Adds opt-in per-group prefix-cache retention and safe DFlash replay, building on #83.

- `GLM53_APC_RETENTION_INTERVAL_SWA=0`, with the global interval unset, preserves target-history checkpoints while prioritizing draft-cache eviction.
- Reconciles reused draft windows with the final hybrid-cache replay boundary.
- Leaves retention behavior unchanged when the new option is unset.
- Applies the cache overlays consistently on both ranks alongside adaptive-k and dense-FP8.
- Credits nood-co1 / BlockbrainLabs for the per-group retention foundation in #83.

Validation: per-group retention, hybrid replay, launcher parity/pre-stop guards, and numeric configuration suites pass against pristine pinned vLLM sources.
[Fresh two-Spark serving check](https://github.com/ratulsarna/spark-serve/blob/c87355cd1406eccc79943ba1941fd53389df3ba3/services/glm-5.3-flash-exl3/qualification/2026-09-11-refresh.md): API/tools/vision pass; two 100K histories and a late edit return correct facts with retained prefixes.

PR #167, opened 2026-09-11
template: add opt-in before-last-user Reasoning Effort placement

### LINKED ISSUE #111: Follow-up to #63: changing reasoning_effort still invalidates the entire prefix cache (54.5 s vs 5.4 s at 42k; ~10 min at 500k)
## Summary

#63 fixed the *thinking* toggle — thank you, it works, we verified it independently. But the same head-of-prompt position means **changing the effort level still invalidates the whole prefix cache**, because the rendered line changes from `Reasoning Effort: Max` to `Reasoning Effort: High` at roughly token 8.

Clients that expose a reasoning picker (OpenCode's Ctrl+T cycles default → low → high → max) therefore pay a full cold re-prefill on **every press**.

## Measured

Live engine, one conversation grown to ~42k tokens, changing only `chat_template_kwargs` between turns:

```
1 cold (default)                ctx 41836   53.5s
2 default, cached               ctx 41885    4.4s
3 default, cached               ctx 41917    5.5s
4 {"thinking":true}             ctx 41961    5.4s   <-- FREE  (this is #63 working)
5 {"thinking":true} cached      ctx 41967    6.5s
6 back to default               ctx 41982    5.5s   <-- FREE
7 {"thinking":true,
   "reasoning_effort":"high"}   ctx 42029   54.5s   <-- FULL COLD RE-PREFILL
8 same, cached                  ctx 42065    6.3s
9 back to default               ctx 42107   10.1s
```

Turn 7 costs essentially exactly the cold-prefill time (54.5 s vs 53.5 s). **At 42k that is a minute; at the 400–550k contexts this recipe is built for it is roughly ten minutes.**

Confirmed at the token level via `/tokenize`, same messages, `add_generation_prompt: true`:

| comparison | shared prefix |
|---|---|
| thinking off vs on | **34 / 34 tokens (100%)** — #63 working |
| thinking off vs on + `reasoning_effort=high` | **8 / 34 tokens (23.5%)** |

## Why it matters more than it looks

The effort axis is a soft steer. On our kit the three settings produce nearly identical output length on the same prompt (default 146 / low 178 / high 154 tokens, all correct), and none of them surfaced `reasoning_content` in non-streaming responses. So the picker's *benefit* is small and its *cost* is a full re-prefill of the whole session.

## Options

1. **Move the effort marker to the end of the prompt** (next to the `<think>` open), so it is a suffix change rather than a prefix change. Biggest win, but changes the prompt shape the model was tuned on — your call whether that is acceptable.
2. **Emit a constant head line and carry the level elsewhere**, e.g. always `Reasoning Effort:` at the head with the value late.
3. **Document it** and let integrators decide — a line in the README next to the reasoning section saying that changing effort mid-conversation costs a full re-prefill would have saved us a production incident.

We took option 3 locally plus a config change: our gateway now exposes a single variant that flips `thinking` only and never sends `reasoning_effort`, so the picker is free. That is a workaround, not a fix, and it costs us the effort axis entirely.

## Environment

2× DGX Spark (GB10), TP=2, image `ghcr.io/miaai-lab/glm-5.3-flash-2x-dgx-sparks:exl3`, `Mia-AiLab/GLM-5.3-Flash-EXL3-TR3-4bpw` @ `25a44fdb`, template as of #63, DFlash2 k=7, fp8 KV, 1M context, MNBT=2048, prefix caching on.

sovereignbrah:
## Current maintainer scope — opt-in, still draft

Current head: [`7f6bb1eaa14989c442bc18703a4996e492ba42fc`](https://github.com/MiaAI-Lab/GLM-5.3-Flash-EXL3-2x-DGX-Sparks/commit/7f6bb1eaa14989c442bc18703a4996e492ba42fc). The maintainer introduced the opt-in revision; this metadata correction is maintainer-owned and preserves the original author's proposal and measurements below.

- **HEAD remains the default.** Opt in with `chat_template_kwargs.reasoning_effort_placement="before_last_user"`; the other accepted value is `"head"`.
- PRE relocation occurs only with thinking enabled, a generation prompt requested, and a structured user turn present. History/no-user renders retain HEAD behavior; thinking-off renders omit the directive. Invalid placement values are rejected. Literal role-marker text is not used to identify user turns.
- Current-main integration, structured-role handling, rendering boundaries and default-equivalence corrections are already addressed. They are not new source-fix requests.
- CPU placement/equivalence checks do not establish answer quality, reasoning cost, cache benefit or performance qualification of this head. Earlier author tables below concern the earlier candidate, not a default-change approval.

### Evidence hold and next owner

The [maintainer bounded panel](https://github.com/MiaAI-Lab/GLM-5.3-Flash-EXL3-2x-DGX-Sparks/pull/167#issuecomment-5658992044) retained **34/36 passing checks and two distinct failures**: low-effort HEAD arithmetic exhausted its 512-token allowance without an answer; low-effort PRE arithmetic returned **167,838 rather than 201,003**. Neither failure may be erased by raising the old cap or retrying to pass, and these results do not establish that PRE caused both. The previously reported Max reasoning-cost caveat remains material.

Keep **draft, merge hold and the exact-head changes-requested review**. The existing correctness/reasoning-cost and predeclared HEAD/PRE quality/cache gates remain: separate failure accounting, completed-request cache counters, reasoning/output usage, latency and unchanged controls, with the pinned TheGrill v0.2.0 published checksums/receipts supplementing—not replacing—answer/tool checks. No present gate is declared passed.

**Next-action owner: maintainer (plotarmordev)** to schedule and pin a supported prospective study and available validation resources; the author retains ownership of their evidence contribution once that work is commissioned, and the maintainer owns final exact-head re-review. No immediate live run or repeat source-fix work is requested. Unblock only after the stated quality/cost/cache gates and separate correctness checks are satisfied on the final pinned candidate, then exact-head re-review approves readiness.

## Original author proposal and historical receipts

The following text is retained verbatim for attribution and measurement history. Its default-PRE, no-user-tail, history-render and unconditional cache-safety statements describe the superseded proposal, **not the current contract above**; its related-PR status and test claims are historical.

<details>
<summary>Original author description (superseded implementation framing; retained evidence)</summary>

## Description and Motivation

**Knob/path:** `files/chat_template.jinja` — where the `<|system|>Reasoning Effort: <Low|High|Max>` directive renders.

**Before:** the directive sat at char ~39, right after `[gMASK]<sop>`. #63 made it unconditional so a thinking toggle kept the prefix cache; a5cc004 (#150) re-gated it on thinking for output quality, at the cost of the cache. Either way, changing `reasoning_effort` between turns re-prefilled the whole conversation, because vLLM chains prefix-cache block hashes from token 0 (#111: 54.5 s vs 5.4 s at 42k; ~10 min at 500k).

**After:** the directive is emitted once, **immediately before the last `<|user|>` turn**, only when thinking is on and a generation prompt is requested (a conversation with no user message falls back to the tail). Everything before that point renders byte-identically across thinking off / low / high / max, so an effort change or a thinking toggle re-prefills only the final user turn.

**Expected direction of measured numbers:** TTFT on an effort change goes from a full cold prefill to a warm follow-up; decode, acceptance and KV pool are untouched (the template renders the same token count, one directive, one position). Thinking-off output is byte-identical to the current template.

**Why before the last user message and not after it (#111 option 1, "next to the `<think>` open")?** I measured both. Both keep the cache. Tail placement was **rejected** on output quality: at `high` on a one-shot code fixture the model emitted a second `</think>` and a duplicated code block **10/22** times (temperature 0, seeded reps). Before-last-user matched the old head placement 14/14 on the same fixture. My read: the model was tuned on `<|system|>…` → `<|user|>` → `<|assistant|>` order; a system line landing directly before `<|assistant|>` is out of distribution, a system line before a user turn is not.

**Safe on both nodes:** template-only; the launcher mounts the same file on both ranks and the worker never renders prompts. No knob, default or overlay changes (the only non-template edit is one commented `EXTRA_ARGS="--enable-prompt-tokens-details"` example in `.env.example`, for reading `cached_tokens`). `GLM53_DEFAULT_REASONING_EFFORT` (#158) keeps working: the default kwargs reach the same `effective_reasoning_effort` variable, and a no-effort request still renders Max when the knob is empty.

### Measurements

2× DGX Spark (GB10), TP=2, this kit's recipe: `Mia-AiLab/GLM-5.3-Flash-EXL3-TR3-4bpw` @ `25a44fdb`, DFlash2 k=7 with the adaptive-k patch, dense FP8, `ABLIT=1`, 900k context, 3584-token hybrid pages, prefix caching on, `EXTRA_ARGS="--enable-prompt-tokens-details"` (the launcher does not add that flag; it is only needed to read `usage.prompt_tokens_details.cached_tokens`, TTFT tells the same story without it). All arms rendered by the server's own template via `/tokenize` + `/detokenize`, directive moved by string surgery, re-tokenized, sent as token ids through `/v1/completions` at temperature 0, seed = rep, 2 in flight. Harness: `tests/bench_effort_placement.py` (this PR).

**Cache proof** — 16,579-token prompt, effort low → high → low, `cached_tokens` / TTFT:

| placement | low (cold) | high | low |
|---|---:|---:|---:|
| head (current) | 0 / 14.9 s | **0 / 15.0 s** | 14,336 / 2.6 s |
| before last user (this PR) | 14,336 / 2.4 s* | **14,336 / 2.5 s** | 14,336 / 2.5 s |
| tail | 0 / 14.7 s | 14,336 / 2.5 s | 14,336 / 2.5 s |

\* warm from the preceding arm — same prefix, which is the point.

**Cache proof at scale** — same sequence, the low → high step only (the one that used to be a full miss), then the warm repeat. Head's residual hits at 47k/152k are the pages it shares with the smaller prompt from the previous pass; everything after them was re-prefilled:

| prompt tokens | head: low → high | before last user: low → high | warm repeat (either) |
|---:|---:|---:|---:|
| 17,907 | 0 cached, **16.1 s** | 14,336 cached, **3.5 s** | 14,336, 3.5 s |
| 46,861 | 14,336*, **28.8 s** | 46,592 cached, **0.9 s** | 46,592, 0.9 s |
| 152,280 | 46,592*, **92.3 s** | 150,528 cached, **2.1 s** | 150,528, 2.1 s |

\* shared with the previous, smaller prompt; not an effort-change hit. The 47k row is #111's scenario (54.5 s vs 5.4 s there); the warm cost is the partial last page plus the final user turn, which is why 47k is faster than 18k (269 vs 3,571 uncached tokens).

Through an OpenAI-style client (top-level `reasoning_effort`, 16,573-token prompt): low → high → none → max → low kept 14,336 cached on every call after the first; a two-turn conversation kept 14,336 at turn 2 and a tier switch at turn 2 cost nothing extra.

**Obedience + quality** — mean reasoning tokens in `<think>`, correct answers / n (exec for code, parse for JSON, exact value for math), mean wall time:

| fixture | placement | low | high | max |
|---|---|---|---|---|
| code | head | 0 tok, 2/2 | 4 tok, **12/12**, 2.1 s | 192 tok, 12/12, 11.5 s |
| code | **before last user** | 0 tok, 4/4 | 4 tok, **14/14**, 3.2 s | **1,196 tok**, 4/4, 44.5 s |
| code | tail | 0 tok, 2/2 | 31 tok, **12/22** (10 malformed), 7.5 s | 230 tok, 12/12, 13.7 s |
| code | none | 214 tok, 2/2 | 206 tok, 2/2 | 212 tok, 2/2 |
| json | head | 0 tok, 2/2 | 5 tok, 2/2 | 196 tok, 2/2 |
| json | **before last user** | 9 tok, 4/4 | 4 tok, 4/4 | 282 tok, 4/4 |
| json | tail | 4 tok, 2/2 | 7 tok, 2/2 | 214 tok, 2/2 |
| math | head | 112 tok, 1/2 | 138 tok, 2/2 | 403 tok, 2/2 |
| math | **before last user** | 105 tok, 2/4 | 138 tok, 4/4 | 322 tok, 4/4 |
| math | tail | 109 tok, 1/2 | 164 tok, 2/2 | 340 tok, 2/2 |
| math | none | 448 tok, 2/2 | 356 tok, 2/2 | 389 tok, 2/2 |
| tool† | head | 0 tok, 2/2 | 0 tok, 2/2 | 18 tok, 2/2 |
| tool† | **before last user** | 0 tok, 2/2 | 0 tok, 2/2 | 24 tok, 2/2 |

† tool-call conversation: system + tool block, user → assistant `<tool_call>` → tool result → follow-up user; the check is exactly one `<tool_call>multiply` with `a=8`, `b=9` and no `</think>` in the answer. 12/12, no malformed output at either position.

Reading: the tiers are obeyed at the new position exactly as at the head (low ≈ 0, high ≈ 4, max ≈ hundreds on code/json). `none` (no directive) is not "max", it is its own ~200-token mode, which is why the directive has to exist somewhere. Math at `low` is weak at every placement, including head; that is a tier property, not a placement effect.

**Caveat, stated plainly:** at `max` on the code fixture the new position reasons ~6× longer than head did (1,196 vs 192 tokens, 44 s vs 12 s; every answer correct). JSON `max` is +44%, math `max` is −20%. `high` is unchanged everywhere. If `max` latency on code matters to you more than cache-safe effort changes, that is the trade.

### What this means for clients (no proxy needed)

- **Top-level `reasoning_effort` already reaches the template.** vLLM's chat endpoint passes it into the template kwargs (`build_chat_params` in `chat_completion/protocol.py`) and sets `enable_thinking = effort != "none"` when the client did not set it. Measured on the native port, one-shot code fixture, temperature 0: `low` → 0 reasoning tokens (1.6 s), `high` → 4 (1.8 s), `max` → 1,277 (30 s), `none` → thinking off (1.8 s). `chat_template_kwargs.reasoning_effort` behaves the same (`max` → 924, `low` → 0). Nothing between the client and vLLM is required for cache-safe tier changes.
- **Gotcha, pre-existing and unchanged here:** the template only knows `low` / `high`; anything else (`medium`, `minimal`, `xhigh`) renders **Max** (`chat_template.jinja:7`). Measured: `medium` → 937 reasoning tokens, 25 s, i.e. the longest tier. An OpenAI-style client that defaults to `medium` gets Max without knowing. Now documented in the README. Mapping `medium → high`, `minimal → low`, `xhigh → max` is a one-line template change; I kept it out of this PR so the diff stays about placement, happy to open it separately.
- **`none` is not Max.** With thinking on and no directive at all the model settles into its own ~200-token mode on every fixture (table above). The directive has to exist somewhere; this PR only moves it.
- **`GLM53_DEFAULT_REASONING_EFFORT` (#158) composes.** The default kwargs land in the same `effective_reasoning_effort` variable. Verified with the knob set to `high`: a request with no effort field rendered `High` before the last user turn and reasoned 4 tokens; with the knob empty it renders Max as before.
- **Multi-turn:** only the final user turn is re-prefilled on a change. History renders (`add_generation_prompt=false`) carry no directive, so a client that renders history itself never sees a stray system line.
- **Streaming is unchanged.** `stream: true` on the native port at `high` (math fixture): first `reasoning` delta at 0.43 s, first `content` delta at 3.4 s, `reasoning` strictly before `content`, no `</think>` in the content stream. `max` on the code fixture: 0.41 s → 32.9 s, same shape. At `low`/`high` on the one-shot code fixture the model answers with no reasoning delta at all (42 tokens, 0.44 s TTFT), which matches the ~0–4 reasoning tokens in the table.
- **Malformed-output signature:** a `</think>` inside `content` (or a second fenced block where one was asked for) is the tail-placement failure mode. `tests/bench_effort_placement.py` checks for it; useful to keep around when anyone next touches the template.

### What I did not test
- Contexts beyond ~152k for the cache proof (#111's 500k case). The block-hash argument is length-independent and the 18k → 47k → 152k rows scale as expected, so I stopped there.
- Concurrency above 2 in flight, other quants or kits, and `max` reasoning length on long prompts (phase 0 generates 4 tokens by design).
- Tool calls with thinking on at `max` on anything harder than the two-step arithmetic fixture (12/12 there).

## Related Issues

**Resolves #111** (effort changes invalidate the prefix cache; proposes tail placement, this PR measures why before-last-user is the right spot).

History of this line: #63 (merged, emit unconditionally), #150 / a5cc004 (merged, re-gate on thinking), #158 (merged, `GLM53_DEFAULT_REASONING_EFFORT`), #87 (closed, superseded by #158), #127 (closed).

Open work this touches or complements:

- #125, #84, #83, #130 (fine-grained APC hits, per-group retention, DFlash replay): complementary, no overlap. They make more of a prefix hit; this PR stops the prefix from being invalidated in the first place. With head placement a tier change was a full miss regardless of hit granularity, so their numbers should only improve on top of this. Different code paths (template vs. engine overlay).
- #51 (`DEFAULT_MAX_NEW_TOKENS`): the only other open PR that edits `files/chat_template.jinja` (assistant/tool hunks). `git merge-tree` of #51 against this branch merges cleanly, no conflict markers.
- #95 (`GLM53_APC_NO_STORE`, per-request skip of prefix-cache writes): orthogonal. It lets a request opt out of caching; this PR makes the cached prefix survive an effort change. Nothing shared.
- #37 (`/reset_prefix_cache` overlay): handy for re-running the cache proof from cold; the harness does not need it (it uses distinct arms).
- #162 (`max` burns the whole output budget): not fixed here, and worth reading next to the caveat above. At `max` on the code fixture the new position thinks longer than head did, so the case for a thinking budget gets stronger, not weaker. `high` is unaffected.
- #120, #106, #166 (cache miss / TTFT reports): an effort or thinking change was one guaranteed full-prefix miss per request; this removes that cause. It does not claim to explain the rest of those reports.
- #135 (OpenClaw, thinking-on default): unchanged. This PR does not touch the thinking default or the parser.
- #147 (open-PR triage): template-only, no launcher knob, no new env.

---

## Testing

- `python3 tests/test_chat_template.py` — 10 tests pass on the new template; 5 of them fail on the current one (the shape is pinned).
- 20 equivalence checks (not committed): for one-shot, multi-turn, multi-turn + tools and tool-call + tool-result conversations, thinking-off renders byte-identical to the current template, and thinking-on renders exactly "current template with the directive moved before the last `<|user|>`".
- `bash tests/test_default_reasoning_effort.sh` — passes (launcher untouched).
- Live: engine restarted with the new template on the 2-node deployment, `/health` 200; 32-check live verification through vLLM and an OpenAI-style proxy: render shape per mode with and without tools, thinking-off JSON, thinking-on at low/high/max, tool call + tool continuation (off and high), a five-image request, the cache sequence above, and the two-turn re-prefill bound. 31/32 on the first pass; the miss was one arithmetic slip on the math fixture at `high`, rerun 6/6 at `high` and 6/6 at `max`.
- `python3 tests/bench_effort_placement.py` — the numbers above: four runs totalling 118 phase-1 completions on code/json/math, plus one run with `--phase0-tokens 16500,42000,128000 --arms head,pre --fixtures tool` for the scale table and the tool row (12 completions).
- Streaming: five `stream: true` requests on the native port (`high`/`low`/`max` code, `high`/`low` math) with `stream_options.include_usage`; delta order and content cleanliness as stated in *What this means for clients*.
- Native port, no proxy: eight one-shot code requests through `/v1/chat/completions` with top-level `reasoning_effort` `low` / `medium` / `high` / `max` / `none`, no effort field, and `chat_template_kwargs` `max` / `low` — the reasoning-token counts in *What this means for clients*.

---

## Checklist:

</details>

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -130,6 +130,9 @@ MTP_TOKENS=2
 # lower at 850k: the boot refuses a pool that cannot hold one max-length request (13 GiB = ~820k).
 # Stock launcher capture-size default is "1 2 4 8 16 24 32".
 # EXTRA_ARGS="--cudagraph-capture-sizes 1 2 3 4 5 6 8 9 10 12 15 16 20 24 32 --kv-cache-memory-bytes 15032385536"
+# Add --enable-prompt-tokens-details to report usage.prompt_tokens_details.cached_tokens per request
+# (prefix-cache hits; tests/bench_effort_placement.py reads it). Off by default in vLLM; combine with any of the above.
+# EXTRA_ARGS="--enable-prompt-tokens-details"
 # Tuning (defaults shown): GLM53_ADAPTIVE_K_ALPHA=0.25 GLM53_ADAPTIVE_K_MARGIN=1.0
 # GLM53_ADAPTIVE_K_MIN_STEPS=4 GLM53_ADAPTIVE_K_SATURATE=max GLM53_ADAPTIVE_K_HIST=200
 # Runtime override without a reboot: <CACHE_ROOT>/glm53_adaptive_k.json ({"mode","set","margin",...}).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,29 @@
 # Changelog
 
+## 2026-09-11 — `Reasoning Effort:` directive moved before the last user message (cache-safe effort changes)
+
+`files/chat_template.jinja` now emits `<|system|>Reasoning Effort: <Low|High|Max>` once, immediately before the
+**last** `<|user|>` turn, only when thinking is on and a generation prompt is requested (a conversation with no
+user message falls back to the tail). Everything before that point is byte-identical across thinking off / low /
+high / max, so a per-request `reasoning_effort` change or a thinking toggle re-prefills only the final user turn.
+Before, the directive sat at char ~39 and every change was a full prefix-cache miss (#63 kept it unconditional
+for that reason; a5cc004 re-gated it on thinking for output quality, at the cost of the cache).
+
+- Cache proof (~16.5k-token prompt, 3584-token pages): low → high → low keeps **14,336 / 16,579** cached tokens on
+  every call, TTFT ~2.5 s; head placement: **0** cached on each change, TTFT ~15 s. At ~47k / ~152k tokens an
+  effort change costs **0.9 s / 2.1 s** (46,592 / 150,528 cached) against **28.8 s / 92 s** at the head.
+- Tool-call conversations (tool block + prior call + result + follow-up user): 12/12 correct calls at low/high/max
+  at either position; streaming `reasoning` deltas still arrive before `content`, nothing leaks into `content`.
+- Tail placement (after the last user turn, just before `<|assistant|>`) also keeps the cache but was **rejected**:
+  at `high` on a one-shot code fixture the model emitted a second `</think>` and a duplicated code block 10/22
+  times. Before-last-user matched head 14/14 there and on the JSON and math fixtures.
+- Caveat: at `max` on that code fixture before-last-user reasons ~6× longer than head (1,196 vs 192 tokens; every
+  answer correct). `high` and `low` reasoning lengths are unchanged.
+- Thinking-off output is byte-identical to the previous template. History renders (`add_generation_prompt=false`)
+  carry no directive.
+- `tests/test_chat_template.py` pins the new shape (10 tests; 5 fail on the old template).
+  `tests/bench_effort_placement.py` is the live A/B used for the numbers above.
+
 ## 2026-09-07 — E3 grouped fat-expert MoE prefill (`EXL3_FAT_GROUPED`, now the default)
 
 Cold prefill **+37–45%** on this 2× GB10 kit (16k: 1,155 → 1,578 tok/s; 128k: ~1,150 → 1,629; 256k: 1,087 → 1,576),
--- a/README.md
+++ b/README.md
@@ -539,8 +539,10 @@ curl -s http://127.0.0.1:8888/v1/chat/completions \
 
 Thinking defaults on. Disable it with the **top-level** JSON field
 `"chat_template_kwargs": {"enable_thinking": false}`. This closes the empty
-thinking block in the generation prompt. The `Reasoning Effort:` line itself
-renders unconditionally since #63 (prefix-cache stability), thinking on or off.
+thinking block in the generation prompt and drops the `Reasoning Effort:` line.
+That line is emitted once, **immediately before the last user message** (not at
+the prompt head as before), so toggling thinking or changing the effort only
+re-prefills the final user turn — see *Client request defaults* below.
 
 Do not send a literal nested `extra_body` object over raw HTTP; `extra_body` is
 an OpenAI Python SDK option that merges its contents into the top-level request.
@@ -556,7 +558,7 @@ template reads. None of them need a restart, and none are enforced by
 
 | Field | Send | Why |
 |---|---|---|
-| `reasoning_effort` | `high` for reasoning work | Unset = **Max** (`files/chat_template.jinja:7`); `low` is the model card's lightest simple-Q&A mode. Keep it constant per route — see below |
+| `reasoning_effort` | `high` for reasoning work | Unset = **Max** (`files/chat_template.jinja:7`, or `GLM53_DEFAULT_REASONING_EFFORT`); `low` is the model card's lightest simple-Q&A mode. Safe to change per request since the directive moved before the last user message — see below |
 | `max_tokens` | ≥ `32768` with thinking on | Max-effort reasoning runs well past 8k output tokens. Too small a cap truncates mid-thought and the reply comes back with empty `content` |
 | `chat_template_kwargs.clear_thinking` | `true` for multi-turn agents | Replaces earlier turns' reasoning with `<think></think>` (`chat_template.jinja:154`), keeping the current tool-call chain. Cuts context, not answer quality |
 | `top_p` / `temperature` | leave unset | `generation_config.json` already supplies `0.95` / `1.0`; the boot log prints the override line. Sending `top_p=1.0` explicitly overrides that and is worse |
@@ -572,10 +574,33 @@ $ curl -s $BASE/v1/chat/completions -d '{...,"reasoning_effort":"low"}' | jq '.c
 ["annotations","audio","content","function_call","reasoning","refusal","role"]
 ```
 
-**Do not vary `reasoning_effort` per request within a conversation.** The effort
-word lands at char 39 of the prompt, so changing it is a full prefix-cache miss
-on an otherwise-warm conversation, not a partial one.
+**Changing `reasoning_effort` (or toggling thinking) per request is cache-safe.**
+The `<|system|>Reasoning Effort: <Low|High|Max>` directive is emitted once,
+immediately before the **last** user message, and only when thinking is on.
+Everything before that point renders byte-identically across off / low / high /
+max, so a change costs one re-prefill of the final user turn, not the whole
+conversation. Measured on this kit (~16.5k-token prompt, 3584-token pages,
+`--enable-prompt-tokens-details` on): low → high → low kept **14,336** cached
+tokens on every call (TTFT ~2.5 s); the old head placement dropped to **0** on
+each change (TTFT ~15 s). At ~47k and ~152k tokens the same effort change cost
+**0.9 s / 2.1 s** before-last-user against **28.8 s / 92 s** at the head.
 `tests/test_chat_template.py` pins that shape.
+
+Why before the last user message and not after it (just before `<|assistant|>`)?
+Both keep the cache, but tail placement was measured and **rejected**: at
+`high` on a one-shot code fixture the model emitted a second `</think>` and a
+duplicated code block 10/22 times (temperature 0, seeded reps); before-last-user
+matched the old head placement 14/14. Thinking off is unaffected either way.
+Caveat: at `max` on the same code fixture before-last-user thinks about 6× longer
+than head did (1,196 vs 192 reasoning tokens, all answers correct); `high` is
+unchanged. Re-measure with `tests/bench_effort_placement.py` (to read
+`usage.prompt_tokens_details.cached_tokens` set `EXTRA_ARGS="--enable-prompt-tokens-details"`;
+the launcher does not add that flag).
+
+Effort words the template does not know (`medium`, `minimal`, `xhigh`) render
+**Max** (`chat_template.jinja:7`), so an OpenAI-style client that sends
+`reasoning_effort: "medium"` gets the longest tier. Send `low` / `high` / `max`,
+or `none` to turn thinking off for that request.
 
 Needs: Docker (no sudo) on both nodes, passwordless SSH head → worker,
 `hf` / `huggingface-cli` + `curl` + `rsync` on the head, ~180 GiB free per
@@ -714,6 +739,7 @@ After CUDA compile, Python overlay edits (`overlay/exl3.py`, tests) are a cheap 
 | `overlay/patch_ablit.py` | install the load_weights hook; bind-mounted and run on both ranks |
 | `ablit/` | direction vectors + `LAYER_MAP.json` from drowzeys' published recipe; `fetch_transplant.py` + `transplant/` for the donor o_proj byte-copy |
 | `tests/test_ablit.py` | recipe integrity, orthogonalization math, TP-shard equivalence, transplant byte-copy + TP slice, hook gating |
+| `tests/bench_effort_placement.py` | live A/B of the `Reasoning Effort:` directive position (head / before-last-user / tail / none): phase 0 cache retention across effort changes on ~16.5k / 42k / 128k prompts (`--phase0-tokens`), phase 1 obedience + answer checks (exec / JSON / exact value / tool call) per fixture × tier × arm, malformed-answer detection |
 | `tests/test_default_reasoning_effort.sh` | `GLM53_DEFAULT_REASONING_EFFORT` enum guard (`""`/`low`/`high`/`max`; `medium` rejected) and the `--default-chat-template-kwargs` flag at both rank sites, sliced out of `start.sh` and evaluated |
 | `scripts/boot-shape-warmup.sh` | post-`/health` DFlash2 k=7 BLOCK ladder + sampler/kpool arms |
 
--- a/files/chat_template.jinja
+++ b/files/chat_template.jinja
@@ -5,7 +5,12 @@
 {%- set thinking_enabled = true -%}
 {%- endif -%}
 {%- set effective_reasoning_effort = reasoning_effort if reasoning_effort is defined and reasoning_effort in ['low', 'high'] else 'max' -%}
-{%- if thinking_enabled and effective_reasoning_effort is not none -%}<|system|>Reasoning Effort: {{ effective_reasoning_effort | capitalize }}{%- endif -%}
+{#- The Reasoning Effort directive is no longer emitted here. It is emitted once, immediately
+    before the LAST user message (see the message loop), so a thinking on/off or effort change
+    never touches the prompt head: vLLM chains prefix-cache block hashes from token 0, and a
+    change at char ~39 re-prefilled the whole conversation. Placing it after the last user
+    message (just before <|assistant|>) was measured and rejected: at High it produced a
+    malformed double-</think> answer 10/22 times. Before-last-user matched head 14/14. -#}
 {%- set clear_thinking = clear_thinking if clear_thinking is defined else false -%}
 {%- if tools -%}
 {%- macro tool_to_json(tool) -%}
@@ -141,7 +146,9 @@ For each function call, output the function name and arguments within the follow
     {%- endif %}
 {%- endfor %}
 {%- for m in messages -%}
-{%- if m.role == 'user' -%}<|user|>{{ visible_text(m.content) }}
+{%- if m.role == 'user' -%}
+{%- if add_generation_prompt and thinking_enabled and effective_reasoning_effort is not none and loop.index0 == ns.last_user_index -%}<|system|>Reasoning Effort: {{ effective_reasoning_effort | capitalize }}{%- endif -%}
+<|user|>{{ visible_text(m.content) }}
 {%- elif m.role == 'assistant' -%}
 <|assistant|>
 {%- set content = visible_text(m.content) %}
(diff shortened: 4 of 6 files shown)


plotarmordev (changes requested):
Holding this PR and requesting changes; it is not part of the independent launcher/benchmark fixes being merged.

The broader frozen evaluation of our reviewed, semantically equivalent placement variant completed 1,056 raw/chat/continuation requests. It recorded 18 baseline-pass/candidate-fail pairs versus 4 reverse pairs, plus 13 paired reasoning-cost flags. The sensitive cases include literal-marker handling and conversations without a user turn. These results concern the reviewed placement variant, not a claim that this exact upstream commit received every test. Code-output semantics were not executed/evaluated.

Please:
1. Rebase onto current main and resolve the README conflict.
2. Address the paired correctness regressions and supply a reproducible per-mode comparison, including literal-marker data and no-user histories.
3. Retain the documented Max-cost caveat and provide an explicit reasoning/completion-cost acceptance gate. Prefix reuse alone does not justify changing the default behavior. An opt-in proposal may keep the existing default, but is not itself a waiver of the correctness gate.

The larger cache-isolation qualification was not completed, so there is no new blanket performance or default-change approval here. #177's forced-tool finish-reason issue was not isolated to this PR and is tracked separately.

plotarmordev:
## Reviewed — changes/evidence requested; do not merge yet

Maintainer source fixes are now **pushed to this PR**, preserving author history: [`7f6bb1eaa149`](https://github.com/MiaAI-Lab/GLM-5.3-Flash-EXL3-2x-DGX-Sparks/commit/7f6bb1eaa14989c442bc18703a4996e492ba42fc). The integration includes main `b3bd1f6928930727b0e9521478340dfd074b16c8` (including #184 and #51). Scoped CPU checks passed on this source. No fresh GPU/model qualification is claimed for this refreshed head; earlier runtime receipts remain tied to their recorded candidate and runtime identities.

### Incorporated source fixes

HEAD placement remains the default; `before_last_user` remains an explicit opt-in. The template uses structured message roles rather than literal role-marker text, preserves HEAD for history/no-user renders, omits the directive when thinking is disabled, rejects invalid placement values, and benchmarks actual HEAD/PRE rendering. The existing 224-case default/explicit-HEAD equivalence coverage is retained; refreshed template and token-default CPU checks passed.

### Why it is still held

The bounded runtime quality panel passed **34/36** checks. Low-effort HEAD arithmetic exhausted its 512-token allowance without an answer; low-effort PRE arithmetic stopped normally with **167,838 instead of 201,003**. These are distinct failures, not evidence that PRE caused both. The cache benefit is real within that acquisition: high/max effort changes reused **7,168 cached tokens with PRE versus zero with HEAD**. The full bounded run retained 42 generation requests and 7,930 completion tokens.

### Specific author changes requested

1. Investigate the truncated HEAD case and wrong PRE answer separately. Preserve their prompts, controls and outcomes in the regression protocol; do not make the old run pass by increasing its cap or retrying.
2. Keep HEAD default and PRE opt-in, including the rendering boundaries and default-equivalence checks above.
3. Predeclare a representative HEAD/PRE quality/cache study covering low/high/max effort, JSON, arithmetic, tool arguments and streaming. Retain actual completed-request cache counters, reasoning/output usage, TTFT and latency, with appropriate unchanged controls. Any different prospective budget/workload is a new study, not a replacement receipt.
4. Demonstrate that the proposed cache/latency benefit does not come with an unacceptable correctness or reasoning-cost change under the declared scope. TheGrill performance receipts supplement—not replace—those answer/tool checks.

### Required qualification before marking ready

Use the published [**TheGrill v0.2.0 experimental prerelease**](https://github.com/plotarmordev/thegrill/releases/tag/v0.2.0), from source `4166e9f0c4467ce516c8d35ae363a80abe7b6494`. Install the native-target archive and verify **that release's checksums and receipts**. The published binaries were rebuilt by CI: earlier local staging hashes are **not** their identities. The release page is authoritative for downloads; the source-pinned installation guide retains pre-publication wording and older v0.1.0 examples.

Follow the [measurement/evidence scope](https://github.com/plotarmordev/thegrill/blob/4166e9f0c4467ce516c8d35ae363a80abe7b6494/docs/performance/README.md), [claim-specific recipe guidance](https://github.com/plotarmordev/thegrill/blob/4166e9f0c4467ce516c8d35ae363a80abe7b6494/docs/performance/SHARED-RECIPES.md), and [report template](https://github.com/plotarmordev/thegrill/blob/4166e9f0c4467ce516c8d35ae363a80abe7b6494/docs/performance/SHARED-REPORT-TEMPLATE.md). Pin the executable, workload/producer, actual backend/runtime and model revisions, and declared settings. Qualify supported request controls before the prospective acquisition; use the full `/v1/chat/completions` endpoint. Keep unchanged-deployment controls and every failed/partial capture. Do not strip fields, replace rejected samples, change gates after seeing results, or treat missing usage as zero. Share a sanitized summary and evidence identities, not credentials, private prompts or deployment configuration.

**For this PR, please keep it draft and mark it ready only after the requested changes and separate correctness/numerical checks pass, and the relevant TheGrill 0.2.0 reports complete and satisfy the predeclared PR-specific acceptance gates. Then request maintainer re-review of the exact final head.** Exit 0 or `DESCRIPTIVE` alone is not a performance PASS; default C1 does not establish fairness, cache-state safety, capacity or kernel correctness. An inconclusive comparison is not a no-regression certificate, and an inconclusive unchanged-deployment performance control does not by itself prove a broken server. Investigate variation that prevents the claimed conclusion rather than discarding the control.

TheGrill publication is not blanket backend support or persisted-model-KV/eviction proof; unsupported claims need separate real-backend evidence or must remain explicitly unqualified. This is the review gate for this PR, not a claim that every project must adopt TheGrill.

**Disposition:** reviewed; changes/evidence requested; merge hold. Tooling availability is now resolved, so only `awaiting-tooling-release` is being removed. Draft status, `merge-hold`, `changes-requested`, `reviewed` and `awaiting-test-evidence` remain.

plotarmordev (changes requested):
Changes/evidence requested on `7f6bb1eaa14989c442bc18703a4996e492ba42fc`. Maintainer fixes are pushed and scoped CPU checks passed; this is not approval of the remaining runtime claims. Specific remaining changes, acknowledged positive/negative evidence, and exact readiness gates are in the [updated maintainer handoff](https://github.com/MiaAI-Lab/GLM-5.3-Flash-EXL3-2x-DGX-Sparks/pull/167#issuecomment-5658992044).

Use [TheGrill v0.2.0 experimental prerelease](https://github.com/plotarmordev/thegrill/releases/tag/v0.2.0) with its published CI-built artifact checksums/receipts, not staging hashes. Keep the PR draft until its separate correctness/numerical checks and relevant predeclared performance gates are satisfied; then request exact-head re-review. Tooling availability is resolved, but the merge hold remains. No additional GPU qualification is claimed by this review.

plotarmordev (commented):
## Hold update

Public head checked: `7f6bb1eaa14989c442bc18703a4996e492ba42fc`.

**Why it is held:** Moving the effort directive improves cache reuse in some cases, but cache reuse does not establish answer quality or acceptable reasoning cost.

**What we tested:** The new cache-placement check used distinct renders with equal token lengths. A long request reused 150,528 tokens with the before-last-user placement; first output was about 1.6 seconds versus about 100 seconds for head placement in that bounded comparison. This was a cache check, not a generated-code quality test. The earlier quality panel's wrong PRE answer and truncated HEAD answer remain separate unresolved observations; neither should be erased or blamed on the same cause without evidence.

**Next action — me, @plotarmordev:** Agree the bounded HEAD/PRE quality and cost comparison and arrange the test window. Preserve low/high/max effort, answer/tool checks, stream handling, unchanged controls and the earlier failing cases.

**Author action now:** Keep HEAD as default and PRE opt-in. No immediate extra hardware run is requested. The existing quality-evidence request remains open; only request readiness after the agreed final-head quality, cost and cache checks pass. A faster first token alone is not enough.

PR #174, opened 2026-09-11
test(perf): add agent-session, Responses-replay and logprob probes + 2026-09-11 findings

mmizutani:
## Maintainer disposition — draft HOLD

The corrected tooling source is published at `5f6e22163d8e7e0570ae500c1bd17ad9b8a20990`; it is not current-main merge clearance or numerical/performance admission. Keep this PR draft and unmerged.

CPU evidence covers five comparison boundary cases (including empty captures), actual capture entrypoints with inert responses, missing-rank handling, and unavailable global prefix metrics. These checks do not qualify FP8 accuracy, cache performance, or live serving. The KL calculation is conditional on shared reported support, not full-distribution KL. Historical measurements below remain author-reported, scoped to their original configurations; they are not a current FP8 qualification.

Remaining source/tooling gate: maintainer `plotarmordev` owns current-main integration and resolution of the remaining per-request missing-usage/`cached_tokens` zero fallback before proposing a tooling-only landing. Counter resets, nonfinite metrics and numerical underflow are not newly qualified. Separately, any numerical admission requires a finalized comparison protocol and explicitly authorized capture; the draft #182 protocol is not finalized. No immediate author benchmark request and no node access is authorized by this disposition.

The corrected historical document remains frozen at SHA256 `79ea542b0c40fb254ac88a1cac529deef28b7988e02800a628e256f4ad678fd2`.

## Original author description — historical only

The following description is preserved for provenance, not endorsed as current qualification:

> ## Summary
> 
> Three stdlib-only probes used for the 2026-09-11 headroom investigation, plus the findings document:
> 
> - `tests/bench_agent_sessions.py` — N concurrent growing sessions; per-turn `prompt_tokens` / `cached_tokens`, global prefix-cache counter deltas and wall time. This is the detector for the multi-session prefix-cache cliff (it scored **0.000 global hits from turn 1** before the fixes in the companion APC PRs and 0.886–0.928 after).
> - `tests/replay_responses_traj.py` — replays a mini-swe-agent trajectory through `/v1/responses` and reports the server-side `cached_tokens` per API call. It separates "the prompt forked" from "the server evicted the prefix": calls that missed live (0 cached) hit on an idle engine with the same payload (32 k–50 k cached).
> - `tests/bench_logprobs.py` — fixed-text prompt-logprob capture plus a KLD/argmax comparison for numeric gates. Used here to qualify dense FP8 vs BF16 dense: KLD 0.002–0.016 nats, argmax agreement 97.0–99.7 % on four fixed texts.
> 
> `docs/headroom-2026-09-11.md` records the findings, measurements and the reconciliation A/B between the PR #83 and PR #130 retention implementations (same-probe boots B3–B7: 3×/4× 30 k repeats 1.000 hits with the #83 semantics vs 0.000 with the #130 rework on this workload; B6 isolated the cause from the DFlash block-drop cherry-pick).
> 
> ## Usage
> 
> ```bash
> python3 tests/bench_agent_sessions.py --agents 4 --turns 8 --base-tokens 60000 --chunk-tokens 6000 --out /tmp/agent.json
> python3 tests/replay_responses_traj.py --trajectory <traj.json> --calls 16,17,18,19,20 --out /tmp/replay.json
> python3 tests/bench_logprobs.py capture a.json   # then, after a config change:
> python3 tests/bench_logprobs.py compare a.json b.json
> ```
> 
> ## Notes
> 
> Draft: tooling + evidence for discussion; the divergence findings are included deliberately so the #130 behavior is documented as measured, not argued.

### DIFF
--- a/docs/headroom-2026-09-11.md
+++ b/docs/headroom-2026-09-11.md
@@ -1,0 +1,401 @@
+# Headroom investigation — 2026-09-11 (2× DGX Spark, GLM-5.3-Flash-EXL3)
+
+Live, on-kit investigation of remaining speed/efficiency levers. Every number
+below was measured on this kit during the session; raw artifacts are under
+`logs/headroom-20260911/`. The prior live serve was shut down before the
+session; the agent sandboxes that shared the head's UMA were also stopped.
+
+## TL;DR
+
+| # | Lever | Status | Measured effect |
+|---|---|---|---|
+| 1 | Adaptive verification length (`GLM53_ADAPTIVE_K=ema`) | **keep ON** | prose **+10.3 %** (A/B/A medians 29.36 vs 26.63 tok/s); structured neutral (65.9 vs 65.9) |
+| 2 | `GLM53_SPINWAIT_MS=16` | **keep ON** | 30.44 tok/s prose on B1 vs 29.36 B0 A-arm (directional, not isolated); repo sweep +0.95 % decode / −85 % EngineCore CPU |
+| 3 | Long-prefill boot warmup rung | **fixed** | the only mid-serve JIT in the prior 5 h serve was `BuildPrefillChunkMetadataKernel`; rungs added, ARG_MAX staging fixed |
+| 4 | Prefix-cache collapse under ≥3 long sessions | **diagnosed, unfixed** | reproducible hard cliff: 1–2 sessions hit ~96–100 %, 3+ sessions hit 0 %; independent of the 918 k-token pool and of `VLLM_PREFIX_CACHE_RETENTION_INTERVAL` |
+| 5 | Dual-rail CX7 NCCL | **verified** | single HCA 12.84 GB/s vs dual 20.94 GB/s peak busbw; `NCCL_CROSS_NIC` / `MERGE_NICS` no further gain |
+| 6 | Host DRAM headroom | **at ceiling** | 2 GiB MemAvailable under concurrent 40 k prefills even with sandboxes stopped — do **not** raise `GPU_MEM_UTIL` |
+| 7 | Dense FP8 (`GLM53_DENSE_FP8=dense,kda`) | **measured, provisional** | structured **75.0** tok/s (+13.7 %), prose **34.4** (+29 % vs B0 stock, +13 % vs B1 spinwait-only); KLD 0.002–0.016 nats, argmax 97.0–99.7 % |
+
+## 1. Method
+
+- Boots are `./start.sh restart` with caller env overrides (caller wins over
+  `.env`; `tests/test_start_overrides.py`).
+- B0: stock capture list, adaptive-k **on** (needed to capture the `{3,5,8}`
+  graphs) and A/B'd live through the override JSON; `GLM53_SPINWAIT_MS=2`.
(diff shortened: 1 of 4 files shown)


plotarmordev:
### Review follow-up — `81c83fc2e4cb`

The three probes and the explicit configuration tables are useful. In particular, B5/B6 explicitly used 0/0; we should not dismiss those results as merely unset-versus-enabled defaults.

Please scope the causal interpretation to what the experiments isolate: these are different overlay stacks, not a single-variable retention comparison. The reported hit-rate contrast should remain recorded, but it does not alone identify the responsible mechanism or justify replacing the replay guard on main. Wording that separates observations from attribution is sufficient; no new run is being requested from you.

Keyed-server support is also absent from the probes; the existing benchmark bearer-key convention is the natural follow-up for that usage. We own any further isolated runtime comparison and the merge decision.

plotarmordev (commented):
Final source/publication review at `5f6e22163d8e7e0570ae500c1bd17ad9b8a20990`.

The historical wording request is addressed: the B3/B4–B6 contrast is scoped to whole overlay stacks, B5/B6 explicit 0/0 remains disclosed, and the document does not justify replacing main’s replay guard. Local/unmerged components and warmup claims are labeled. The findings document is unchanged by the subsequent probe corrections (SHA-256 `79ea542b0c40fb254ac88a1cac529deef28b7988e02800a628e256f4ad678fd2`); historical numbers remain author-reported, not newly qualified.

Source review accepts the CLI argument guard, capture-length/text-name checks, unavailable rather than fabricated comparison values, conditional-overlap KL labeling, missing-counter diagnostics, preservation of captures without rank-1 entries, and handle/dead-parameter cleanup. Final review found a reachable empty-capture exclusion count of -1 at `869f710`; Main corrected it in this head and supplied a five-case CPU regression receipt failing before and passing after. The original four-case receipt did not cover that case. This is diagnostic evidence only, not a numerical admission gate or performance qualification.

The author branch was fast-forwarded after unchanged-public-head, ancestry, full-tree and pushed-range secret/content gates. The PR remains draft and unmerged. Main supplied CPU receipts; this reviewer ran no tests, builds, formatters, live endpoint calls or GPU/node operations. Bearer support remains out of scope. Current-main integration, any further runtime qualification, readiness and landing remain maintainer-owned.

plotarmordev (changes requested):
## Hold update

Public head checked: `5f6e22163d8e7e0570ae500c1bd17ad9b8a20990`.

**Change needed:** The probes must not count missing usage or missing cache details as zero. That can make an accounting failure look like a real cache result.

**What we tested:** With our local correction, a real 4-agent × 8-turn run produced all 32 usage/detail records. This checks accounting on that run; it is not a model-quality or throughput pass. It also does not test the missing-data failure path by itself.

**Next action — me, @plotarmordev:** Publish the correction already prepared for `tests/bench_agent_sessions.py` and `tests/replay_responses_traj.py`, then have the published head reviewed. The author does not need to redo our local work.

**Before ready:** Missing fields must remain unavailable, affected totals/ratios must be withheld, and the probe must return a nonzero/incomplete result. Check complete and missing-data responses, then repeat the bounded session check on the published version. Keep historical measurements separate from these new checks.

PR #175, opened 2026-09-11
fix(launchers): make PYTORCH_CUDA_ALLOC_CONF overridable

RealDealCPA-VR:
## Problem

`start.sh` hardcodes the allocator config on the serve container:

```
-e PYTORCH_CUDA_ALLOC_CONF=expandable_segments:True
```

That default is right for serving, but derived images that attach a
hidden-state KV connector — drafter self-distillation capture lanes, for
instance — have to disable `expandable_segments`, because the connector
refuses it. Today the only way is to patch `start.sh` locally, which then
has to be stashed and re-applied around every pull. Since this repo
force-pushes history, that local edit also has to stay uncommitted to keep
`git merge --ff-only origin/main` working, so it is one `git checkout` away
from vanishing silently.

## Change

```
-e "PYTORCH_CUDA_ALLOC_CONF=${PYTORCH_CUDA_ALLOC_CONF-expandable_segments:True}"
```

The single-dash form is deliberate. `${VAR-default}` substitutes the default
only when `VAR` is **unset**, so an explicitly empty value passes through —
which is exactly what disabling the allocator option requires. `${VAR:-default}`
would swallow it.

## Impact

No behaviour change unless the variable is set. Every existing user gets a
byte-identical launch command; the default is unchanged. Verified on a 2x
DGX Spark (GB10, TP=2, E3 grouped MoE, DFlash2 k=6): the running container
reports `PYTORCH_CUDA_ALLOC_CONF=expandable_segments:True` with the variable
unset, same as before.

https://claude.ai/code/session_0138mTmrVuGrqPBSXpxz1gQq


plotarmordev:
Maintainer follow-up before merge: Preserve the unset-only allocator default, caller export precedence over .env, and explicit-empty semantics; quote worker environment arguments so both ranks receive the same value. The TP4 launcher is unchanged.

Applying normal follow-up/prerequisite commits, preserving the original PR ancestry without a force push. Prepared exact head: `0ea1735661d057225bce3670755d4882f4e41b38`. Focused checks passed: bash -n start.sh; python tests/test_launcher_rank_parity.py; python tests/test_start_overrides.py; python tests/test_numeric_config.py; python tests/test_bench_decode_auth.py; python -m pytest -q tests/test_nccl_multi_hca.py. The final head also passed the secret/content gates; independent HELP review checked the corrected scope. No deployment-local files or live-run artifacts are included.

plotarmordev (approved):
Approving exact corrected head `0ea1735661d057225bce3670755d4882f4e41b38` after independent HELP review and the focused checks recorded above. Preserve the unset-only allocator default, caller export precedence over .env, and explicit-empty semantics; quote worker environment arguments so both ranks receive the same value. The TP4 launcher is unchanged.

Proceeding with a normal, head-pinned merge. No hosted checks are configured/reported; local results are not being represented as hosted CI. No new GPU or performance qualification is claimed. Follow-up: run merged-branch hosted validation when CI becomes available.

PR #182, opened 2026-09-13
perf(sm121): speed up EXL3 decode and KDA FP8 prefill

rodman80:
## Summary

Two independently guarded, opt-in SM121 (GB10) optimizations:

1. **EXL3 thin-decode fast path** (`GLM53_EXL3_MOE_FAST=1`) — K4/N256 kernels with a 1-stage register / 8-stage shared pipeline plus optional gate/up transform reuse.
2. **KDA large-M FP8 hybrid path** (`GLM53_KDA_FP8_FAT=1`, requires `GLM53_DENSE_FP8=dense,kda`) — FP8-Marlin for small M, torch native FP8 `_scaled_mm` for large M on the same logical FP8 weights.

Exact-head qualification is recorded in `docs/qualification-evidence.md` (candidate `4c0f422`; runtime code from `fe71e6d`; image `glm53eval/pr182:fe71e6d`). Author commits sit on top of the maintainer fixes (`f91fe07` preserved, no force-push): `fe71e6d` TP=1 harness fix, `abc4ee4` bench control fix, `4c0f422` evidence docs.

## Scope and defaults

Both flags default OFF and fail closed to stock behavior. Qualification scope is **homogeneous NVIDIA GB10 / SM121** only. Non-SM121 behavior is fail-closed by construction (native per-device capability check, Python capability check, per-boot functional probe, no persistent cross-device readiness cache) but has not been performance-qualified.

## EXL3 thin-decode path

Native change (`overlay/patch_exl3_decode_pipeline.py`, additive only): two K4/N256 kernels (shared vs independent gate/up transform) plus a `glm53_fast_moe_version` symbol. Dispatch requires K==4, N256-compatible dims, flag=1, SM121 per call. Transform reuse only when gate/up SUH tables are the identical allocation (proven by all-expert `torch.equal` at load). Requesting the flag on an image built without the patch raises at load.

SASS/resource inspection of the exact-head binary (host CUDA 13.0 toolkit, no installs): the fast kernel shows **REG 127, 16 B stack frame, 6 local stores, 0 local reloads** — all observed local reloads eliminated (37 stores → 6, all write-only; 88 B → 16 B frame) versus stock K4/N256 (REG 128, 37 STL / 39 LDL). Occupancy class unchanged. This supersedes the earlier ptxas-only note; "spill-free" is retired as an absolute claim.

## KDA large-M FP8 path

`Glm53DenseFp8Method.apply` routes on tensor metadata (no sync): retained weights + no bias + matching K + flattened M > 64 → `_scaled_mm`, else Marlin. Load-time retention only for `kda` in-proj `[12576, 4096]` on SM121 after a live M=65 `_scaled_mm` probe per layer (~0.54 ms/layer, ~18 ms total, ~266 KB transient; failure retains nothing). Keeps raw `[N,K]` e4m3 + fp32 scales (~49.2 MiB/layer-rank); the transposed view is the col-major operand, zero copy. `o_proj`/`f_b`/`g_b`/dense stay Marlin by measurement. Fused Triton rowwise-quant with eager fallback (bitwise-exact where applicable).

## Exact-head correctness

The requested harness bug is fixed (`fe71e6d`): standalone entries entered with `_TP=None` and died in `process_weights_after_loading` (`AssertionError: tensor model parallel group is not initialized`, reproduced). Both entries now run inside a real single-rank lifecycle (genuine NCCL group + TP=1 coordinators, exception-safe destroy, leak = error; no mocks), with `Tp1LifecycleTests` pinning the failure, the fix, and clean restore. Also fixed (`abc4ee4`): the bench's Marlin control silently ran fat-vs-fat under `FAT=1`; it now forces a Marlin-only control layer.

## Numerical repeatability

The frozen ≥0.99 argmax gate is **not satisfiable even by unchanged stock repeats** (measured ~0.938 structured, ~0.972 prose). A five-stock-run study found a 13-position fragile pool; every flip was a near-tie (top-1 margin ≤ 0.5). For B-thin (3 runs vs 5 stock): no position showed systematic B-consensus against A-consensus, extras were not reproducible between B runs, B3 had zero extras outside the pool (max cross-arm top-1 margin ~0.425). For FAT: repeats consistent with stock except near-ties; BF16-relative kernel error identical for both paths (~0.0266); one arm-consistent tie resolved toward stock in the third run. No mathematical equivalence is claimed. **Whether this tie-aware analysis suffices for admission remains a maintainer decision (see questions below).**

## Exact-head performance

TheGrill v0.2.0 experimental prerelease, source `4166e9f0c4467ce516c8d35ae363a80abe7b6494`, CI tarball SHA256 `4d89b3db20baae380902b2b92c1c289ad8c69f1ba95a372c62c077ec86fa5b6e`, binary SHA256 `3cb208b667558bdfe5120f6393af3b997ab6e10c2c1aa1c548ca494acc4a9564`. Workloads: routine decode `6d075fb8617280ad062393e169f4b162d12840338bbb2749e6871be5672b5846`, routine prefill `f0ddfda5f5bb35b6c1637fd27e02c5ae3ef9aeb37ed5e61fc0b953b52a61edd6`, mixed `886c96c9efeaa4f6bac85e3d19af3408469edd55eb1fdc3a82aedaa3dacce06f`. Selected comparisons are DESCRIPTIVE (never a PASS); gates were predeclared before candidate data.

### EXL3 thin (isolated, FAT off, dense off)
Decode workload A/B/A2, all 8/8 complete: candidate **+4.8% to +8.4% across cells** (structured C1 ≈ +8.3%); restored A2 within ±2.4%. Clean isolated thin-path evidence.

### KDA FAT (isolated, dense,kda both arms, FAST off)
Decode workload A/B/A2: candidate within **±1.4%**, A2 within ±1.2% — decode non-regression as designed (M≤64 never leaves Marlin). Engineering prefill/mixed A/B: **prefill +18.0% / +18.7%, mixed decode +5.6%, mixed prefill +10.0%**, c1 tied. Prefill/mixed improvements are retained as **engineering A/B evidence**: the corresponding TheGrill profiles require reported-prefix-zero accounting but this backend returns `prompt_tokens_details: null` (INVALID capture retained, never weakened or rerun).

### Combined (separately collected, dense,kda held both arms)
Decode A/B/A2: **+5.7% to +9.9%**, A2 within ±2.1%; 100k prefill ≈ +20.0%. These combined values were directly measured; isolated percentages were not added mathematically.

## Memory and startup cost

Controlled worker-rank fresh-boot comparison: stock **79.65 GiB** vs FAT **81.28 GiB** loaded model (**+1.63 GiB/rank**, matching 34 × 49.2 MiB). KV explicitly pinned to **14.0 GiB in both arms**; tested 100k-context and C6 configurations remained usable. This is not a claim that all possible KV-capacity configurations are unchanged. The M=65 retention probe was kept: negligible measured cost against its device/shape safety benefit.

## Sanitizer and soak

`compute-sanitizer --tool memcheck` (via the existing host toolkit mount, no system installs): **0 errors** on the KDA GPU harness (17 checks pass) and on the EXL3 FAST=1 battery (119 cases; tensors compare ALL PASS). Bounded soak on the combined candidate (~25 min: C3/C6, cold prefill, mixed, text, cancel-probe): health HTTP 200 throughout, zero restarts/OOM/CUDA errors. Two soak-probe artifacts of mine (default-thinking token budget, short-answer assert) were investigated and cleared. 25 minutes is a bounded soak, not long-term stability proof.

## Portability

Homogeneous SM121/GB10 only. Fail-closed safeguards: native device check, Python capability check, per-boot functional probe, no persistent cross-device cache. Other devices untested.

## Known qualification limitations

- No `decide` PASS/REGRESSION envelope (descriptive + predeclared gates; prefill profiles backend-unqualified as above).
- Numerical admission uses the tie-aware interpretation pending maintainer agreement.
- One disclosed 4-minute overlap (engineering capture during grill setup/warmup, excluded from measured results); one combined-16k single-sample wobble; prior zero-wave baseline campaign retained as historical failure (warmup emitted a tool call after text; collector correctly rejected it).
- Earlier-head numbers below are historical only.

## Testing

- 19/19 CPU tests in-image, incl. 3 TP1 lifecycle regression tests.
- EXL3 corrected battery 119/119 PASS (distinct top-8 routing from a 32-expert pool; uniform/correlated/skewed; shared+independent SUH; streams; graphs with changed data+routing; fallback) + full-model serving at production geometry.
- KDA GPU harness PASS (retention, dispatch spies, numerics, bitwise graph replay, fail-closed).
- Overlay suite 5/5 under both flag states. `git diff --check` clean.

## Historical evidence

Earlier exploratory campaigns (pre-`f91fe07` heads) reported larger headline deltas (e.g. combined C1 +20.8%, EXL3 C1 +8.6%, FAT prefill +19.9%). They are superseded by the exact-head results above and retained in `docs/sm121-perf-paths.md` labeled historical.

## Maintainer decision questions

1. The frozen numerical gate fails unchanged stock repeats. Expanded exact-context analysis found only near-tie run-to-run flips and no systematic B-thin divergence. Is this tie-aware repeatability analysis sufficient for numerical admission, or would you prefer a new prospectively frozen numerical gate/protocol?
2. The published TheGrill prefill/mixed profiles require reported-prefix-zero accounting, but this vLLM backend returns `prompt_tokens_details: null`. Should the +18–20% prefill / mixed engineering A/B evidence remain explicitly unqualified, or is there another approved evidence path for these claims?

PR remains draft; re-review not requested. Guidance on the two questions above is welcome whenever convenient.


plotarmordev:
will look into it! thank you

plotarmordev:
## Reviewed — changes/evidence requested; do not merge yet

Maintainer source fixes are now **pushed to this PR**, preserving author history: [`f91fe07fec72`](https://github.com/MiaAI-Lab/GLM-5.3-Flash-EXL3-2x-DGX-Sparks/commit/f91fe07fec72e01b493641265f3dad54821c8b7f). The integration includes main `b3bd1f6928930727b0e9521478340dfd074b16c8` (including #184 and #51). Scoped CPU checks passed on this source. No fresh GPU/model qualification is claimed for this refreshed head; earlier runtime receipts remain tied to their recorded candidate and runtime identities.

### Incorporated source fixes

Fail-closed flag/load handling, flattened-row dispatch, safe strided activation handling, aligned scale/division semantics, the real per-layer M=65 retention probe, logprob-screening corrections and valid distinct-top-8 routing are preserved. Current TP3 exclusions are integrated. All affected test/benchmark constructors now supply the required projection prefix. Refreshed fast-path, KDA tensor/dispatch, logprob, routing, dense-FP8, adaptive-K, numeric and launcher checks passed. **Both optimization defaults remain off.**

### Positive evidence and unresolved failures

Earlier native evidence passed the corrected **119-case** battery and **18 timing-shape output comparisons**. Across **5,400 timing samples**, standalone fast-layer median latency was **6.619–11.053% lower**, with maximum stock-repeat drift **1.040%**. This is not full-model throughput or qualification of the refreshed code.

All 32 planned one-token full-model numerical requests completed, but **stock A/A2 also failed the frozen numerical gates**. That makes overall qualification inconclusive; it does not explain away B-thin's **87.5% structured-context agreement against both controls** (the controls agreed on all 16 positions), or its 24-token-context choice differing from both. Empty shared top-20 support stays unscorable, not zero KL.

The third A performance warmup emitted tool calls after text; the collector correctly rejected it. **Zero measured waves**, no B/A2 performance captures and no replacement acquisition. Earlier B-fat startup confirmed 34 retained KDA layers per rank and roughly 1.63 GiB additional loaded-model memory, but varying available KV memory did not establish a controlled capacity result. Static fast-kernel resource fields were retained; complete SASS auditing was blocked by unavailable `nvdisasm`, so no spill-free claim was made.

### Specific author changes requested

1. **Fix standalone GPU-harness initialization:** `tests/bench_fp8_fat.py` and `tests/test_kda_fp8_fat_gpu.py` must initialize and clean up real single-rank vLLM model-parallel state, or execute through a genuinely initialized runner. Upstream `process_weights_after_loading` now calls `get_tensor_model_parallel_world_size`; these entry points do not initialize it. CPU execution of the pinned functions with their actual initial `_TP=None` reproduces `AssertionError: tensor model parallel group is not initialized`. The constructor-prefix updates alone do not satisfy this prerequisite. Do not mock world size in qualification.
2. Investigate stock repeatability **and** the thin-arm discrepancies against both controls. Keep identical teacher-forced token IDs, unchanged numerical gates and failed/unscorable cases. Pass numerical admission before collecting performance evidence.
3. Keep distinct-top-8 routing at the actual expert geometry. Qualify thin/fat paths separately, with graph/fallback coverage, actual retention, memory/KV accounting, and the required state/correctness checks.
4. Complete the relevant SASS/spill/occupancy inspection, isolate probe startup overhead and supply scoped soak/portability evidence—or explicitly narrow the supported claims for maintainer agreement. Microbenchmarks do not establish full-model quality or capacity.
5. Qualify a supported prospective text-output workload. Do not strip tool calls, weaken output admission or replace rejected captures to manufacture throughput approval. Any changed protocol needs a separately declared new study with retained historical failures.
6. Keep both defaults off and submit exact-head numerical plus relevant TheGrill performance evidence before marking ready.

### Required qualification before marking ready

Use the published [**TheGrill v0.2.0 experimental prerelease**](https://github.com/plotarmordev/thegrill/releases/tag/v0.2.0), from source `4166e9f0c4467ce516c8d35ae363a80abe7b6494`. Install the native-target archive and verify **that release's checksums and receipts**. The published binaries were rebuilt by CI: earlier local staging hashes are **not** their identities. The release page is authoritative for downloads; the source-pinned installation guide retains pre-publication wording and older v0.1.0 examples.

Follow the [measurement/evidence scope](https://github.com/plotarmordev/thegrill/blob/4166e9f0c4467ce516c8d35ae363a80abe7b6494/docs/performance/README.md), [claim-specific recipe guidance](https://github.com/plotarmordev/thegrill/blob/4166e9f0c4467ce516c8d35ae363a80abe7b6494/docs/performance/SHARED-RECIPES.md), and [report template](https://github.com/plotarmordev/thegrill/blob/4166e9f0c4467ce516c8d35ae363a80abe7b6494/docs/performance/SHARED-REPORT-TEMPLATE.md). Pin the executable, workload/producer, actual backend/runtime and model revisions, and declared settings. Qualify supported request controls before the prospective acquisition; use the full `/v1/chat/completions` endpoint. Keep unchanged-deployment controls and every failed/partial capture. Do not strip fields, replace rejected samples, change gates after seeing results, or treat missing usage as zero. Share a sanitized summary and evidence identities, not credentials, private prompts or deployment configuration.

**For this PR, please keep it draft and mark it ready only after the requested changes and separate correctness/numerical checks pass, and the relevant TheGrill 0.2.0 reports complete and satisfy the predeclared PR-specific acceptance gates. Then request maintainer re-review of the exact final head.** Exit 0 or `DESCRIPTIVE` alone is not a performance PASS; default C1 does not establish fairness, cache-state safety, capacity or kernel correctness. An inconclusive comparison is not a no-regression certificate, and an inconclusive unchanged-deployment performance control does not by itself prove a broken server. Investigate variation that prevents the claimed conclusion rather than discarding the control.

TheGrill publication is not blanket backend support or persisted-model-KV/eviction proof; unsupported claims need separate real-backend evidence or must remain explicitly unqualified. This is the review gate for this PR, not a claim that every project must adopt TheGrill.

**Disposition:** reviewed; changes/evidence requested; merge hold. Tooling availability is now resolved, so only `awaiting-tooling-release` is being removed. Draft status, `merge-hold`, `changes-requested`, `reviewed` and `awaiting-test-evidence` remain.

plotarmordev (dismissed):
Changes/evidence requested on `f91fe07fec72e01b493641265f3dad54821c8b7f`. Maintainer fixes are pushed and scoped CPU checks passed; this is not approval of the remaining runtime claims. Specific remaining changes, acknowledged positive/negative evidence, and exact readiness gates are in the [updated maintainer handoff](https://github.com/MiaAI-Lab/GLM-5.3-Flash-EXL3-2x-DGX-Sparks/pull/182#issuecomment-5658993179).

Use [TheGrill v0.2.0 experimental prerelease](https://github.com/plotarmordev/thegrill/releases/tag/v0.2.0) with its published CI-built artifact checksums/receipts, not staging hashes. Keep the PR draft until its separate correctness/numerical checks and relevant predeclared performance gates are satisfied; then request exact-head re-review. Tooling availability is resolved, but the merge hold remains. No additional GPU qualification is claimed by this review.

rodman80:
## Evidence update since `CHANGES_REQUESTED` (head `4c0f422`, still draft)

Public body rewritten to the exact-head evidence (`docs/qualification-evidence.md`); stale claims removed. Author commits on top of `f91fe07`, no force-push. No runtime-code changes in this update, both defaults still OFF, labels untouched, re-review not requested.

- **Harness fix** (`fe71e6d`): standalone KDA entries now run inside a real TP=1 lifecycle (genuine NCCL group + coordinators, exception-safe destroy, no mocks); `Tp1LifecycleTests` pins the original `AssertionError`, the fix, and clean restore. **Bench control fix** (`abc4ee4`): the "Marlin" control silently ran fat-vs-fat under `FAT=1`; now forces Marlin-only.
- **Kernels**: EXL3 battery 119/119 PASS (distinct top-8 fixtures); KDA harness PASS incl. bitwise graph replay; layer +8–17% thin, fat-prod 1.8x@M65 → 6.2x@M3584 vs true Marlin.
- **SASS** (exact-head `.so`): fast kernel 0 local reloads (6 write-only STL, 16 B frame) vs stock 39 reloads (88 B frame). "Spill-free" retired as absolute wording.
- **Sanitizer**: memcheck 0 errors on both GPU harnesses (host toolkit mount, nothing installed).
- **Numerical**: frozen ≥0.99 argmax gate fails stock-vs-stock too (~0.938 structured, ~0.972 prose); 13-position near-tie pool over 5 stock runs; no systematic B-thin/FAT divergence (B3: zero extras outside pool; F2/F3 resolve suspects toward stock). Tie-aware admission proposed, not asserted.
- **Performance**: THIN decode A/B/A2 +4.8–8.4% (A2 ±2.4%); FAT decode ±1.4% (A2 ±1.2%); engineering prefill +18.0%/+18.7%, mixed +5.6%/+10.0%; COMBINED decode +5.7–9.9% (A2 ±2.1%), 100k prefill +20.0%. TheGrill prefill/mixed profiles are backend-unqualified (`prompt_tokens_details: null`, INVALID retained, never weakened) — those stays engineering evidence.
- **Memory**: +1.63 GiB/rank measured (79.65 → 81.28 GiB), KV pinned 14 GiB both arms, 100k + C6 usable. Probe 0.54 ms/layer. Bounded soak (~25 min, combined candidate) clean, health 200 throughout.

Two methodology questions for maintainer guidance (also at the end of the PR body):

1. The frozen numerical gate fails unchanged stock repeats. Expanded exact-context analysis found only near-tie run-to-run flips and no systematic B-thin divergence. Is this tie-aware repeatability analysis sufficient for numerical admission, or would you prefer a new prospectively frozen numerical gate/protocol?
2. The published TheGrill prefill/mixed profiles require reported-prefix-zero accounting, but this vLLM backend returns `prompt_tokens_details: null`. Should the +18–20% prefill / mixed engineering A/B evidence remain explicitly unqualified, or is there another approved evidence path for these claims?

plotarmordev:
Thanks—the rewritten body addresses the stale-body request. This answers the [two methodology questions](https://github.com/MiaAI-Lab/GLM-5.3-Flash-EXL3-2x-DGX-Sparks/pull/182#issuecomment-5677264055) at `4c0f422ebf5f11dbec86b25d3f101cc4e9c1540e`; it is not an immediate re-review or approval.

**1. Numerical admission: a new prospective study, not retrospective PASS.** The [reported stock-repeat failures and tie analysis](https://github.com/MiaAI-Lab/GLM-5.3-Flash-EXL3-2x-DGX-Sparks/blob/4c0f422ebf5f11dbec86b25d3f101cc4e9c1540e/docs/qualification-evidence.md) justify revisiting the gate, but a rule derived from those same observations cannot qualify them. I am not adopting the evidence document's “CONDITIONAL PASS”; old failures and unscorable cases remain intact.

**I (plotarmordev) own protocol finalization before requesting any new captures.** The executable specification must freeze the contexts/token identities, separate thin/fat arms, run counts/order, calibration-versus-held-out evaluation split, statistic and numerical acceptance allowance, uncertainty/multiple-position handling, and invalid-capture policy. It must explicitly define tie/consensus handling, margin units and the no-stock-flips case. Positions selected as unanimous across all stock runs have zero leave-one-out disagreement by construction; that is not an estimate of unseen flip noise. Likewise, a 2-of-3 candidate majority against stock consensus can reject ordinary stock tie noise. We need independent stock calibration/evaluation and a justified sensitivity/false-rejection tradeoff, not either shortcut or an unspecified allowance. No observed stock flips must not silently mean zero noise or an undefined margin bound; insufficient calibration remains inconclusive. This is a maintainer-owned specification hold, **not an author rerun request under an unspecified gate**.

**2. Prefill/mixed: retain the engineering results as explicitly unqualified; no TheGrill code change is currently warranted.** The [pinned parser](https://github.com/plotarmordev/thegrill/blob/4166e9f0c4467ce516c8d35ae363a80abe7b6494/crates/grill-perf/src/wire.rs) preserves missing accounting, and the [eligibility check](https://github.com/plotarmordev/thegrill/blob/4166e9f0c4467ce516c8d35ae363a80abe7b6494/crates/grill-perf/src/model.rs) requires the declared cache evidence—missing is not zero.

The [public September 9 smoke](https://github.com/plotarmordev/thegrill/blob/4166e9f0c4467ce516c8d35ae363a80abe7b6494/docs/performance/2026-09-09-recipe-smoke.md#hardware-software-and-public-recipe-parameters) records `--enable-prompt-tokens-details` on the same base-image digest pinned in this [Dockerfile](https://github.com/MiaAI-Lab/GLM-5.3-Flash-EXL3-2x-DGX-Sparks/blob/4c0f422ebf5f11dbec86b25d3f101cc4e9c1540e/Dockerfile). For a **future, separately authorized fresh campaign**, preserve existing `EXTRA_ARGS` and append that flag; the [launcher already appends extra arguments on both ranks](https://github.com/MiaAI-Lab/GLM-5.3-Flash-EXL3-2x-DGX-Sparks/blob/4c0f422ebf5f11dbec86b25d3f101cc4e9c1540e/start.sh). That smoke supports flag acceptance, not qualification of the current overlay stack: it used `observe`. Field availability, `cached_tokens == 0`, and `cached_tokens <= prompt_tokens` establish reporting/internal consistency, **not faithful cache accounting under the overlays**. Exact-runtime accounting semantics and request-level control evidence remain unverified; enabling the flag alone promises neither fidelity nor eligibility. [Server-wide metrics](https://github.com/plotarmordev/thegrill/blob/4166e9f0c4467ce516c8d35ae363a80abe7b6494/docs/performance/PROVIDER-ACCOUNTING.md#what-these-numbers-are-and-are-not) or [switching to observe](https://github.com/plotarmordev/thegrill/blob/4166e9f0c4467ce516c8d35ae363a80abe7b6494/docs/performance/README.md) cannot satisfy the unchanged cache claim or repair old INVALID captures.

Keep draft/defaults off and all historical failures/INVALID captures. Next action is mine: publish the finalized numerical protocol and settle the exact-runtime accounting validation contract before any acquisition request; readiness still requires the subsequent scoped evidence to satisfy those predeclared gates. No restart, hardware run, or immediate re-review is requested now. Verification here was public-source/document review only, not independent runtime reproduction.

plotarmordev (commented):
## Hold update

Public head checked: `4c0f422ebf5f11dbec86b25d3f101cc4e9c1540e`.

**Why it is held:** The original numerical rule also failed unchanged stock repeats. We cannot turn those old results into a pass by choosing a new rule after seeing them.

**What we checked:** Your updated evidence document and the earlier review. The stale-description request is addressed. Your kernel, sanitizer and performance results remain recorded; we did not independently repeat them in this maintenance run. Missing cache accounting is still missing—not zero—and the unqualified prefill/mixed reports stay unqualified.

**Next action — me, @plotarmordev:** Publish the numerical test plan before asking for more captures: fixed inputs and token positions, separate thin/fat settings, calibration and evaluation sets, repeat order/counts, tie handling, uncertainty limits, and invalid-run rules. I also own the runtime accounting check and the choice between #182 and #196.

**Author action now:** None. No new measurements or rebase requested. I am clearing the old native change request as a status correction, not approving the PR. Keep the merge hold and defaults off until the agreed checks pass.

plotarmordev:
### Keeping the optimization review in this PR

@BadAd84, thank you for rebasing this work in #196 and recording your build/boot results. I am closing #196 so we keep one implementation review and one set of acceptance rules here. The rebase notes and measurements remain available there; the branch is not being deleted.

The blocker here is **numerical correctness and how we judge normal run-to-run variation**, not the ability to rebase. The core optimization tests, benchmarks and `docs/qualification-evidence.md` in #196 are unchanged from this PR. Rebased code and a successful boot do not settle the numerical question.

Your existing #196 report is useful additional evidence: both options enabled, a successful build/boot, the reported speculative-acceptance result, and 7/8 identical temperature-zero outputs with a similar stock observation. These are your reported results; we have not independently reproduced them or accepted them as a numerical pass.

**Useful next contribution:** Please bring the existing evidence and any observations about stock-versus-candidate variation into this thread, with the tested commit, workload and test settings (sanitized). You do not need to repeat runs just to move the discussion here. @rodman80 and @BadAd84 should work against the same eventual acceptance rules, not competing PR-specific rules.

**My next action:** I (@plotarmordev) still owe the agreed test plan: fixed inputs, separate optimization settings, unchanged controls, repeat counts/order, tie handling and pass/fail limits. No new hardware run or rebase is requested until that is published. Existing failures and incomplete results stay in the record. Defaults remain off and the merge hold remains. I will reuse useful integration work from #196 when the final version is prepared.

plotarmordev:
FYI, I'm working on finalizing an exact runtime baseline with validation runs on my hardware. will post the finalized acceptance protocol and the results when complete, its a full day process. So rn this PR is the main prio

rodman80:
Take your time, man! gotta make sure everything is in the right spot before merging, I'm 200% fine with that. It'll be a nice speedup for everyone when it's finished, that's what I'm hoping for.

plotarmordev:
Hi @rodman80 — we finished the testing we promised. 818 requests, 18 boots. Here is the full picture and what we did with it.

## Thin (decode fast path) — passed everything we could measure → merged as #217

| Question | Answer |
|---|---|
| Does it change the model's answers? | No difference found — every check stayed inside stock's own noise (0 consensus flips, 0 reproducible shifts, bar 11 and 0.529) |
| Does it make decoding faster? | Yes: **+7.7% / +8.9% / +14.6%** on current main (flag off→on→off, fresh boot each) |
| Does speed recover when off? | Yes (A2 within ±3.5%) |

One honest note: our formal rule couldn't issue an official "pass" because it also fails on stock-vs-stock. We say "no difference found," not "certified." #217 keeps your commits and credit.

## Fat (prefill FP8 path) — failed quality, in a very specific way

| Finding | Value |
|---|---|
| Reproducible answer shifts | **8** (threshold 0.529; observed up to 4.31 and one infinite) |
| Worst case | a reference token goes from ≈1% to ≈78% probability (4.31-nat gap) |
| Vanished answers | stock's chosen word fell out of the candidate's top-20 in **all 6** candidate runs (stock: never in 8 runs) |
| Coordinated tie flips | 15 positions flipped the same way in all 6 runs — odds ~1 in 1,500 if random |
| Where it happens | exactly at the M>64 boundary where `_scaled_mm` turns on; M≤64 prompts on the Marlin path stayed **bit-for-bit identical** |

That last row is the key: the old math path is clean, the new math path shifts answers. The speed is real, but it's computing meaningfully different probabilities.

## New tool so you can verify this yourself — #218 (merged)

`tests/bench_numerical_calibrated.py`. The old fixed rule (99% agreement) was broken — stock fails it against itself. This one measures the model's own wobble first, then judges a change against *that*. Four checks, stdlib-only, same capture format and fixed texts as the existing panel. `python3 tests/bench_numerical_calibrated.py selftest` runs offline, no server. Re-run offline on the retained campaign captures it cleared the thin arm (WITHIN: no text reached the flip bar, zero separations, mean KL inside the stock null in all 26 texts) and flagged the fat arm (FLAG: `E-M04` mean KL 1.78e-2 against a 1.31e-2 bar, with single separations on `E-L2`/`E-M03`/`E-M05`/`E-M07` and tie-rate annotations on `E-B65`/`E-M02`).

## Where this leaves the PR

- **Thin:** alive and merged as #217 — already making our kit faster. Thank you for it.
- **Fat:** two options — fix the M>64 divergence (the panel in #218 should reproduce it on your kit), or drop it and let #217 carry the PR's value.
- Either way is fine with us. Happy to answer any question about the study.

plotarmordev:
### Follow-up: screening-tool correction and qualification limits

@rodman80 — #219 has now merged as a correction to the numerical screening tool introduced in #218. SIX reviewed the exact corrected head before merge.

| Item | Current status |
|---|---|
| Thin decode path | Merged separately in #217; the measured decode gains remain +7.7%, +8.9%, and +14.6% in the three resolved workloads of our two-node off/on/off comparison. Two other workloads were unresolved. |
| FAT prefill projection | Not included in #217. The earlier campaign found reproducible numerical differences on the M>64 path; that is not, by itself, proof of worse task accuracy. |
| Original formal qualification | **INCONCLUSIVE for both arms.** The stock-control failures prevent calling either arm formally qualified. |
| Corrected screening tool | #219 fixes position-local calibration, evidence validation, Poisson thresholds, and missing-token handling. It remains a heuristic screen, not a validated qualification or equivalence test. |

### FAT prefill findings — original frozen study

These are observations from the original #182 campaign, **not a replay with the corrected #219 tool**.

| Finding | Recorded result |
|---|---|
| Reproducible position-specific shifts | **8** exceeded the study’s **0.529-nat** separation threshold. |
| Largest finite example | **4.31 nats** at `E-M01/1`. |
| Missing reference token | At `E-M04/2`, the stock-reference token was absent from **all 6 candidate top-20 captures**. Absence from top-20 is censored evidence—not proof of zero probability or an infinite true shift. |
| Boundary controls | Shifts appeared at **M=65/66**, where FAT engages; the **M=63/64 Marlin-path controls** remained clean. M is the operation’s row count, not a universal prompt-length rule. |
| Unanimous near-tie changes | **15 positions** changed the winning token in the same direction across all 6 candidate runs. |

These findings motivate investigation of the FAT path, but do not establish task-accuracy loss. **The formal FAT verdict remains INCONCLUSIVE because stock-control checks failed.**

Corrections to my earlier comment: “passed everything we could measure” and “failed quality” were too strong. Thin had no reproducible additional numerical shift detected by that study; FAT showed numerical changes requiring investigation. “Odds ~1 in 1,500 if random” was also an imprecise interpretation of the randomization p-value, not a probability that the change was harmless. The speed gains were measured during testing, not evidence that we had permanently deployed thin.

The earlier WITHIN/FLAG replay used the previous tool version and reused campaign data. **We have not replayed thin/fat with the corrected v3 tool**, so those old outputs must not be presented as validation of this version. Existing valid captures remain reusable, but v2 calibration receipts must be regenerated as v3.

This follow-up required no GPU runs or deployment changes. We are not requesting another hardware run or rebase from you to close this tooling correction. The remaining FAT decision stays with us as maintainers: investigate a correction under an agreed acceptance method, or omit that path. Existing observations about the M>64 change are welcome; the tool alone does not establish qualification.

plotarmordev:
@rodman80 @BadAd84 — we’ve merged the thin decode portion as **#217**. On our two-node setup, it improved decode speed by roughly **8–15% in three workloads**.

For the FAT prefill portion, **we’d like your opinion on the poorer agreement we measured against stock, and help finding a way to keep the speedup while reducing those differences.**

We compared the probabilities the model assigns to tokens with and without FAT. Some variation between runs is expected, but with FAT enabled we found repeatable changes larger than the variation our study used as its reference.

| What we found | What it means |
|---|---|
| **8 repeatable probability shifts** exceeded the study’s threshold. | This wasn’t just one unusual response. The differences recurred at specific positions. |
| In one case, a token’s probability moved from roughly **1% to 78%**. | The model became much more likely to choose that token. |
| In another case, the stock-reference token disappeared from **all 6 FAT top-20 captures**. | It consistently fell outside the recorded top choices—not necessarily to zero probability. |
| The **M=63/64** Marlin controls stayed clean, while changes appeared at **M=65/66**, where FAT engages. | That points toward the FAT projection path as a useful place to investigate. Here, M is the operation’s row count. |
| **15 near-tie positions** switched their winning token in the same direction across all 6 FAT runs. | Some close choices also shifted consistently rather than varying independently between runs. |

I described this earlier as “failed quality.” More precisely, we measured poorer agreement with stock—not a demonstrated drop in answer accuracy. Our formal study also failed its stock-control checks, so neither arm received a conclusive formal qualification. Those limits matter, but the repeatable FAT differences still deserve a closer look.

**How do you interpret these results?** Do they look like an expected precision tradeoff from this FP8 projection, or something in the implementation that could be improved? Is there a way to bring it closer to stock while retaining a useful prefill speedup?

We’d like to work toward a solution rather than simply discard the prefill work. The separate screening tool has been corrected in **#219**, but these findings are from the original study—not a new run with that corrected tool. We can share the recorded cases and agree together on what a useful follow-up comparison should demonstrate.

rodman80:
@plotarmordev Hey, that's very good news! I'm happy the thin part is already on code so everyone can enjoy some speedup. While you take a deserved break, I'll investigate the fat results. It's a pretty nice prefill speedup, I can't give up on it yet, lol. 

My first impression is this might be on the activation: in Marlin, it activates straight up from BF16, but in the fat solution, there's a FP8 quantization before that, so the actual activation runs on FP8 before the rest of the work is done. I'll do some digging and see what we can do to preserve the speedup trying to maintain quality.

Marlin Conceptually:

BF16 activation
×
FP8 weight
→ Marlin W8A16

FAT v1 Conceptually:

BF16 activation
→ per-row e4m3 activation quantization
→ FP8 activation
×
FP8 weight
→ aten._scaled_mm W8A8

The current FAT row quantizer uses one scale across an entire K=4096 activation row:
scale = max(abs(row)) / 448

Therefore FAT v1 introduces activation quantization that stock Marlin does not. This will be my first investigation line.

sdougbrown:
This looked interesting so I tested it out as well. 

**1. Thin-decode path: no measurable gain on TP3 once coop-EP is active.**
After merging the thin split (`3b986c7` upstream side) we A/B'd `GLM53_EXL3_MOE_FAST` on `bench_concurrency --levels 12 --modes code --ctx 0`: FAST=0 {133.7, 143.0} vs FAST=1 {137.3, 144.7, 132.8} agg tok/s — distributions overlap. x4 was flat (92.2 vs 92.4). The structural reason: on TP3 the cooperative-EP kernel (1.6.0, `extensions/cooperative_moe/tp3/`) already dispatches rows 1-35 via a measured per-row policy, so the small-M shapes your TP2 receipts are built on never reach `exl3_moe` here. The fast kernels only see mid-M stock calls (policy rows 36-63, 64+), where the pipeline-depth advantage flattens. coop-EP + FAST=1 showed no interaction regressions (42/42 layers eligible, zero preemptions across all cells).

**2. KDA FP8 hybrid (`GLM53_KDA_FP8_FAT=1`): silent no-op on TP3 — the retention guard pins the TP2 shape.**
Ran the PR head's `overlay/exl3.py` (source sha `9b6a08b1f372`) with the coop footer on all three ranks, `GLM53_DENSE_FP8=dense,kda`, `GLM53_KDA_FP8_FAT=1`:

- Boot logs show no retention/probe activity — the load-time `[12576, 4096]` match never fires.
- All measured cells are within noise of the same stack without the PR overlay: warm-tail TTFT on a 224k prefix `2.68 / 4.07 / 5.66 s` (vs `2.71 / 4.10 / 5.72 s`), cold 224k prefill `296.7 s` (vs `295.5 s`), x12 decode `142.7` agg (inside the 132.8-144.7 band above).

Cause: the guard pins the full TP2 in-proj `[12576, 4096]`. TP3 shards the KDA in-proj row space (12576/3 = 4192 rows/rank), and TP3 boots additionally pad attention heads 64 -> 66 via `hf_overrides` (`num_attention_heads: 66`), which changes the per-rank matrix again. So `KDA_FP8_FAT_SHAPES = {(12576, 4096)}` can never match on TP3, and flag=1 retains nothing.

Question: is extending the shape set (or keying retention on the unsharded logical shape pre-shard) in scope here? The large-M dispatch is exactly the regime our prefill-bound workload lives in — warm turns at 222k depth run marginal rates around 800 tok/s, and decode at 12 lanes presents M ~ 96 per step, above your M>64 threshold. If a TP3-aware shape set lands, we can re-run this battery and report real numbers. Happy to help scope what the per-rank shapes look like under TP3 + head padding.

rodman80:
Hey, @sdougbrown ! Sorry, I don't personally own a TP3 setup, so I can't reproduce it here. I believe for THIN, I think by taking the time to measure the behavior of M in a TP3 set like we did with TP2 and following the same idea, you probably can also have some speedup as well, maybe not in the same 8-15% range, but more speed is always good.
Unfortunately, I cannot help more with that unless I win the lottery and buy another Spark :)

Considering THIN has been merged in PR #217 I'd suggest you open another draft PR to try to port these gains to TP3 as well.

About the fat KDA FP8 hybrid, I'm trying a different approach for the speedup, since it seems that the FP8 conversion is causing a little drift. I'm trying to find a way to use BF16 in the same way Marlin would and get the speedup anyway. Should have more news on that soon.

Rod

sdougbrown:
I'll absolutely look into that, I just didn't want to step on toes and start drafting that up if you intended to 🙂

rodman80:
Follow-up from the FAT investigation: the attribution work confirmed that the reproducible v1 shifts were dominated by BF16->FP8 activation quantization, not `_scaled_mm` itself.

I tested a replacement large-M path that keeps activations BF16 and materializes the same logical FP8 KDA weights in BF16. Small-M/decode stays on Marlin; only M>512 prefill uses the BF16 GEMM.

In the prospective stock / FAT-v1 / BF16-v2 study, v1 reproduced the old M>64 pathology while v2 did not. The v2 projection is ~10x closer to stock at layer level, cold/mixed prefill improves ~11-14%, and decode remains unchanged. The main tradeoff is +3.29 GiB/rank of retained model memory.

I split the implementation into a clean draft PR here: #233

#182 remains useful as the research/evidence history; the new draft contains only the BF16 large-M implementation. TP3 is intentionally out of scope because I only have the TP2/2-node cluster available for validation.

PR #171, opened 2026-09-11
feat(apc): add opt-in 64-token prefix-cache hits

mmizutani:
## Description and Motivation

The hybrid coordinator includes `KpoolTailManager` in its fine-grained lookup veto even though that scratch group opts out of prefix caching. This PR lets participating cache groups use 64-token lookup boundaries when the runtime layout supports them, instead of always falling back to the 3,584-token hybrid page.

**`GLM53_FINEGRAINED_APC=0` is the example, launcher, and runtime default.** Set `1` and restart to opt in. Current evidence demonstrates a conditional benefit, but does not qualify default enablement or resolve earlier draft-acceptance concerns.

- Scope lookup compatibility to participating groups and verify scratch alignment from live manager/spec attributes. On this layout, fine enablement requires `hash_block_size % index_kpool == 0`. Participating coarse-only managers retain the fallback; unsafe or unverifiable scratch layouts refuse initialization when fine hits would otherwise be enabled.
- Validate existing patch-owned helper/gate regions canonically, including AST scope and owned-name bindings. Drift fails without rewriting the target. Remove the connector receipt and speculative upstream-fixed silent no-op detector.
- Retain exact `0`/`1` validation, explicit-empty rejection, kill switch, pre-stop artifact checks, both-rank wiring, and Dockerfile integration. Caller overrides use main's existing generic mechanism.
- Preserve #130 retention/replay behavior, including the 2,048-token DFlash replay requirement. The README documents the knob and tests; the [99-line feature note](docs/DESIGN-apc-fine-grained-hits.md) covers the contract, safety invariant, verification, and limitations.

Based on upstream main `d8ad1831182412dfc7323d90f4ad91b8d636451e`.

### When enabling is worthwhile

A promising workload to test is **a long, unchanged prefix that remains cached, followed by substantial new text**—for example, an existing conversation followed by a large tool result. A usable fine-grained boundary must survive the retention and DFlash replay checks; a long conversation alone does not ensure a benefit.

**Keep `GLM53_FINEGRAINED_APC=1` only after repeated on/off comparisons on representative requests show a worthwhile improvement:** more actual cached tokens and reduced time to first output and/or total response time, without regressions in answer correctness or draft acceptance. Check completed-request cache counters and timings; an enabled 64-token alignment or an intermediate replay-clamp target is insufficient evidence.

[...]

PR #195, opened 2026-09-15
Rebase #125 onto main (fine-grained hybrid prefix-cache hits) + diagnostic log

BadAd84:
Rebased version of #125 (by @rwl4) onto current main, resolving the conflicts against 9b9eb13 (per-KV-cache-group APC retention with DFlash SWA replay):

- Kept main's DFlash SWA replay machinery (BASE_HELPER, INIT/CONVERGE hooks, v3 eagle-verify) and layered #125's additive layer on top: `FINE_MARK`, `FINE_OLD/FINE_NEW` (filter `unsupported_partial_hit_managers` by `group.kv_cache_spec.participates_in_prefix_caching`), three-block idempotent `main()`.
- Merged the bench (`DEFAULT_PAGE_TOKENS = 64` + HEAD's newer /reset_prefix_cache notes) and the README (keeps HEAD's table + reset-route text, adopts the 64-token page-math note).
- Added a one-shot `logger.warning` when fine-grained lookup remains disabled by participating managers - on our TP=2 kit the merged overlay applies cleanly but warm hits still land on 4x3584 boundaries, so the log names exactly which manager still blocks 64-token hits on this coordinator build.

Built and boot-tested on 2x GB10 TP=2 (image builds clean through the in-image test suite, incl. test_apc_per_group_retention composition).


plotarmordev (commented):
## Hold update

Public head checked: `9f3eb7865b0ffe9b9675d0dda4ad89548b8ce2c1`.

**Why it is held:** We need to choose how this rebased version of #125 fits alongside #125 and #171. This is a maintainer choice, not a failed author test.

**What we tested:** Earlier integrated cache candidates gained reuse on long and medium append requests, but reused fewer tokens on a short append. Those runs were not on this new head, so they do not qualify or reject it. A successful boot is also not proof of cache correctness.

**Next action — me, @plotarmordev:** Choose the implementation and supported scope, then review and qualify that exact integrated version. Preserve the replay-safety guard and the useful negative cases.

**Author action now:** None; no new rebase or run requested. If this is selected, the agreed comparison must include cold, repeated, short-append and long-append requests, completed-request cache counters, correct answers, and separate state/numerical checks.

BadAd84:
Closing - keeping this local; thanks for the context on direction.

PR #196, opened 2026-09-15
Rebase #182 onto main (SM121 EXL3 thin-decode + KDA large-M FP8 prefill)

BadAd84:
Rebased version of #182 (by @rodman80) onto current main (a35eaab), resolving conflicts against the per-group-retention/SWA-replay overlay work:

- `overlay/exl3.py`: kept main's `_glm53_use_marlin()` TP3 unaligned-KDA guard and layered #182's `kda_fp8_fat_enabled()` + large-M `_scaled_mm` dispatch (they are complementary; the fat path's validated shape set is TP2-local so the TP3 guard is unaffected).
- `tests/test_numeric_config.py`, `CHANGELOG.md`: keep-both merges (main's kv-capacity-log/mixed-prefill tests + #182's optimization-flag validation tests).

Built and boot-tested on 2x GB10 TP=2 with both opt-in flags ON (`GLM53_EXL3_MOE_FAST=1`, `GLM53_KDA_FP8_FAT=1` on top of `GLM53_DENSE_FP8=dense,kda`, adaptive-k ema):
- image builds clean through the full in-image test suite
- `kda fp8-fat retained for kda [12576x4096] (+49.2 MiB/rank)` engaged at load
- cold prefill 1310-1514 tok/s at ~23k prompts; decode prose 27.8-33.3 tok/s, code 39 tok/s
- spec-acceptance gate PASS (pos0 0.85, monotone decay to pos6 0.11 - no vllm#53030 pinning)
- temp-0 determinism 7/8 on repeated structured task (same one-flip behavior as main without this stack)


plotarmordev (commented):
## Hold update

Public head checked: `48cadc467c23b44741335003eeacf52fca76bb02`.

**Why it is held:** This is the rebased version of #182. We have not yet agreed on a numerical pass/fail rule that handles ordinary variation in the unchanged model. A rebase does not settle that question.

**What we checked:** The new head and its relationship to #182. We did not run this head on hardware. The performance results in the PR remain author-reported, not independently reproduced here.

**Next action — me, @plotarmordev:** Publish the agreed numerical test plan and settle the relationship between this PR and #182. It must define the inputs, separate optimization settings, repeat counts, unchanged controls, tie handling, and pass/fail limits before new measurements.

**Author action now:** None. Please do not run another benchmark or rebase just to clear this hold. Once the plan is agreed, the selected final version needs numerical checks and claim-specific performance evidence. Keep the options off by default.

plotarmordev:
Thank you @BadAd84 for doing the rebase and sharing the build/boot measurements.

I am closing this as a duplicate of the ongoing optimization review in #182, not because the rebase or your results are unwelcome. The actual blocker is the numerical-correctness test plan and its pass/fail rules; a rebase does not resolve that. Keeping two PRs for the same proposal would split that review and could make one look ready while the same blocker remains.

The core optimization tests, benchmarks and qualification document are unchanged from #182. Your conflict-resolution notes and additional measurements remain useful and are preserved here. I have linked and credited them in [the main discussion](https://github.com/MiaAI-Lab/GLM-5.3-Flash-EXL3-2x-DGX-Sparks/pull/182#issuecomment-5690218616). Please contribute any existing stock-versus-candidate observations there; no fresh benchmark or rebase is requested now.

I (@plotarmordev) own publishing the agreed test rules before asking either contributor for new runs. #182 remains held with both options off by default. We can use this branch's integration notes when preparing the final version. Closing the PR does not delete your branch or discard your credit.

PR #217, opened 2026-09-18
perf(sm121): EXL3 thin-decode fast path (thin-only split from #182)

plotarmordev:
Draft — thin-only split of the EXL3 half of #182 (`pr/sm121-exl3-kda-perf`, head `4c0f422`), updated onto current `main` (`0437c09`, i.e. after #216's Dockerfile test-chain fix) by merge commit. Parent effort: #182. No re-review request yet; the ask at the bottom is a maintainer disposition decision.

Base note: the branch was prepared on `b0e39f6` and updated with a merge commit after #216 landed, so the diff below is against `0437c09`. Dockerfile merged cleanly: this branch's edits are the exl3 decode-pipeline `COPY`/`RUN` regions and the `glm53_fast_moe_version` self-check assertion, #216's repair is the test-chain `RUN` and the `instanttensor` `RUN`, and the two regions do not overlap. Verified after the merge — diff vs `main` is exactly the three thin edits, diff vs the previous head is exactly #216, no orphan continuation lines, and `docker build --check` reports no warnings.

## Scope

One opt-in optimization, hardware-scoped to SM121/GB10:

`GLM53_EXL3_MOE_FAST=1` — SM121 K4/N256 **thin-decode** kernels for the EXL3 routed experts. `overlay/patch_exl3_decode_pipeline.py` adds two kernels to `exllamav3_ext` at image build time (1 fragment stage instead of 3, 8 shared-memory stages instead of 3; the second variant reuses the gate input Hadamard for the up GEMM). The stock `exl3_moe` kernels are untouched, all source anchors are validated before anything is written, and a second application is refused. Dispatch (native, per call) requires K == 4, N256-compatible dimensions, `GLM53_EXL3_MOE_FAST=1` and SM121; transform reuse additionally requires the gate/up SUH pointer tables to be the identical allocation, which `overlay/exl3.py` creates only in fast mode and only after the load-time per-expert `torch.equal` proof.

**Fat is excluded, in full and by design.** #182 also carried a KDA FP8 large-M dispatch (`GLM53_KDA_FP8_FAT`, `_scaled_mm` dispatch, `KDA_FP8_FAT_*`, fused rowwise-quant Triton kernel, weight retention, `bench_fp8_fat.py`, `test_kda_fp8_fat{,_gpu}.py`, `bench_kda_fp8.py`, `test_kda_logprob_compare.py`) and its qualification record. That work produced **reproducible numerical shifts at the engaged M>64 geometry** — 8 T2 exceedances (gaps 0.54–∞ nats, including the E-M04/2 reference absent from all six candidate top-20s), T3b failures in the B/M families, 15 unanimous tie flips, A2 T4 p=6.7e-4 — while its own in-arm M=63/64 control in the same boots was bit-identical to stock, i.e. the failure is specific to the geometry the optimization engages. Nothing from that half is in this branch: no FAT env name, dispatch, kernel, test, bench or doc text.

## Defaults and fail-closed behaviour

* `GLM53_EXL3_MOE_FAST` defaults to `0` (`start.sh`, and `start-tp3.sh` keeps unsetting it). With the flag off there is no behaviour change: the stock kernels run and the pointer tables are exactly what `main` builds. The only edit the port makes to the off-path is that the gate/up SUH alias is created *only in fast mode* — one deviation from #182's head, which aliased whenever the load-time equality proof held; the alias is content-identical either way, so no measured arm is affected, and `FAST=0` is now byte-identical to `main`.
* The flag is literally `0` or `1`. The launcher rejects anything else — including explicit empty and surrounding whitespace — before `restart` stops a running pair; `overlay/exl3.py` and the native dispatcher re-check at load.
* `FAST=1` fails closed at model load when the image has no native kernels (`glm53_fast_moe_version` absent or != 1) or when the fused `exl3_moe` path is unavailable (`EXL3_FUSED_MOE=0`, missing symbol, any fused-state build error): a requested fast path never silently degrades to the Python loop. The per-call predicate is by design not a failure — geometry the fast kernel does not cover (K != 4, non-N256 dimensions, non-SM121) legitimately keeps the stock kernel, and the GPU battery asserts that the fallback geometry matches stock.
* Scope is homogeneous SM121 / GB10 only; non-SM121 hardware is untested here.
* Both ranks receive the same flag value (head explicitly, worker through the serve_env list), and `tests/test_launcher_rank_parity.py` now asserts that.

## Numerical evidence (sealed study) and its verdict

Prospective study `pr182q-20260917` (closed 2026-09-18), hash-frozen protocol `protocol-numerical-FROZEN.md` sha256 `588c5393a0f4385a…` (full digest in the study's `FREEZE.json`), calibration addendum `calibration-addendum.json` sha256 `c9539c22172d7c1d…`: m\*=1.5 nats (grid cap), stock consensus-flip-rate 95% upper bound r_up=7.47e-4, tau_C=0.241 nats, classes CONF 5,783 / TIE 3,768 / UNSTABLE 1,196; stock null shared-top-k KL mean 0.0128 / max 0.0131 (all 35 balanced splits 0.0127–0.0131). Window 2 ran 8 stock + 6 thin captures on E material with interleaved boots.

* **T1 consensus flips: 0** (critical value 11 overall; 5 primary scope).
* **T2 reproducible shifts: 0** (max observed gap 0.054 nats vs bar 0.529).
* **T3b relative KL: pass every family** (means inside the stock null).
* **T4 tie-flip rate: F1 0.0887 < F0 0.0902** (p=0.42; A2 exhaustive p=0.477).
* **A2 exhaustive randomization** (all 3,003 balanced relabelings, per scope): T1 p=1.0, T4 p=0.48.

**Formal verdict: INCONCLUSIVE** — via the predeclared control self-test only. The absolute T3a gate (KL ≤ 0.01) fails on stock-vs-stock itself (stock null 0.0142–0.0146 > 0.01 on every split), and one bimodal long-context stock position pushed the tau-transfer bound (tau_E 0.5288 vs 0.500). Both control failures are instrument-calibration properties of the material, not candidate behaviour. This is **not** a pass and **not** a clearance.

Honest limits: single deployment (2×GB10, TP=2, dflash-7, dense=kda); teacher-forced positions, no multi-step decode; a formal PASS requires a successor study id with the absolute gate re-derived and a quantile-based tau transfer.

## Speed evidence

* Kernel-level, exact head: **+8% to +17%** layer latency fast vs stock; SASS 127 regs / 16 B frame / **0 local-memory reloads** vs stock 128 regs / 88 B frame / **39 reloads**; parity battery 119/119 PASS at rel RMSE ~1e-8; `compute-sanitizer --tool memcheck` 119 cases, 0 errors.
* Serving A/B/A2, **maintainer-measured at head `3d6ffbd` on current `main`** (image `sha256:f267534bbe52b792dbcaa3d1439bb4f5aaa02ccde7349eb7bc7eea5becac5116`, TheGrill v0.3.0, `glm-routine-decode-v3`, fresh two-node boot per arm, same image in all arms, `GLM53_DENSE_FP8=kda` and `GLM53_KDA_FP8_FAT=0` in every arm, thinking-off, no tools). **Descriptive A/B/A2 — no PASS envelope is claimed.** Decode tokens/s (median) A / B / A2:

| cell | A `FAST=0` | B `FAST=1` | A2 `FAST=0` | B vs A | A2 vs A (noise) |
|---|---|---|---|---|---|
| structured-1 | 72.74 | 78.36 | 72.52 | **+7.71%** | −0.31% |
| code-1 | 67.83 | 73.86 | 68.04 | **+8.88%** | +0.30% |
| json-1 | 51.32 | 58.81 | 49.50 | **+14.59%** | −3.54% |
| structured-2 | 68.86 | 78.07 | 71.67 | withheld (raw +13.4%) | +4.07% |
| prose-1 | 31.93 | 34.91 | 33.40 | withheld (raw +9.3%) | +4.62% |

  The cells with resolved ranges also improve on achieved completion throughput (+7.93% / +9.13% / +14.36%) and wave latency (−7.34% / −8.37% / −12.56%). Every arm published 20 waves (15 measured, 15 eligible) with 0 invalid captures, 0 retries and 0 replacement runs; every response reports `cached_tokens=0` on the declared-cold legs. Two cells are **withheld by the tool's own range-overlap rule** — their raw medians are labelled raw and no change is claimed for them.
* Disclosed caveats for that measurement: the server-side generation override caps every trial at 256 output tokens (identical in all arms, below the workload's 400-token cap); the A2-vs-A drift above is the noise control the deltas are read against; arm A (the `FAST=0` baseline) was re-entered through the operator's reuse path after an operator bug, with no request served before its capture; and **the run declarations recorded a placeholder image digest** (`sha256:IMAGE_ID`) — the real identity is independently evidenced by the arm containers' `Image` field, the image inspect and the in-image gate, and this collector does not attest declarations against the server. Full record: `~/.local/state/pr182-qualification-20260917/SPEED-TEST-RESULT.md` (+ `.json`).
* Serving A/B/A2 (**historical, author-measured at the pre-rebase head**): +4.8% to +8.4% decode across all cells (structured-1 +8.3%), with the A2 stock repeat within ±2.4%. Superseded for this rebased branch by the current-`main` measurement above.
* Provenance: the SASS, parity-battery, sanitizer and layer-latency numbers, and the historical A/B/A2, are author-reported measurement recorded in the parent effort (#182, branch head `4c0f422`); this branch does not carry that qualification record and does not re-derive those numbers. The current-`main` A/B/A2 is separate and was measured by the maintainer on this branch's own rebased image. The numerical study is separately hash-frozen and cited with its own identities.

## Not claimed / unknown

* Client **tool workloads are not validated** on the surface #215 (`tool_choice:none` decode guard) introduced — nothing here asserts behaviour for them.
* Composition with the cooperative-MoE path is **untested** (the TP3/TP2 profile generators are repinned, see below, but no combined numerical or speed run exists).
* No combined headline percentage, no `decide` PASS envelope, no absolute "zero spills" claim (6 write-only STL remain), no non-SM121 claim.

## Rebase notes

* Ported: `overlay/patch_exl3_decode_pipeline.py`, the thin parts of `overlay/exl3.py`, `start.sh` thin wiring, `Dockerfile` COPY/RUN + `glm53_fast_moe_version` self-check, the thin test/bench files, `tests/test_numeric_config.py` thin contract, docs.
* Dropped from #182 (in addition to the fat items above): `tests/bench_decode_conc.py`, the `tests/bench_decode.py` and `tests/_run_cold_prefill.py` campaign-harness edits, `docs/qualification-evidence.md` (its `#182`-specific fat and combined-campaign sections), and the CHANGELOG text that described them. The thin-relevant content in that PR's post-head commits was checked: `f91fe07`, `fe71e6d` and `abc4ee4` touch only `tests/bench_fp8_fat.py`, `tests/test_kda_fp8_fat*.py` and (new in `fe71e6d`) `tests/_vllm_tp1.py`, all excluded here, so nothing from them is in scope.
* Deliberate deviation from #182's head: there the gate/up SUH pointer alias was applied whenever the load-time equality proof held, including `FAST=0`; here it is created only in fast mode, so the off-path keeps exactly `main`'s pointer tables. The alias is content-identical by that proof, so no measured arm changes; the CPU contract test asserts the new three-way behaviour (off, fast+shared, fast+independent).
* Conflicts with current `main`: `start.sh` (main moved on with #207/#212/#215) — the four thin wiring points were re-applied onto main's structure; no guard was relaxed, the existing validations (retention, SWA, mixed-prefill, `tool_choice`, no-store, KV-capacity-log) are untouched, and the new 0/1 check uses the same `_glm53_validate_bool_flag` helper as the other switches. Other `overlay/exl3.py` hunks applied unchanged.
* One non-thin file had to move with it: `main` added `extensions/cooperative_moe{,/tp3}/prepare_profile.py`, which pin `overlay/exl3.py` byte-for-byte and refused to build a profile ("review compatibility before repinning"). After review their `STOCK_SHA` is repinned to the new digest: with the flag unset (the launcher default, and what `start-tp3.sh` enforces) the module builds the same tables and takes the same paths as before, and the guards still refuse any other content. Dated deployment provenance in `docs/astra-results.md` / `docs/ASTRA-PORT_REPORT.md` is left untouched; `docs/cooperative-moe-handoff.md` records the new pin and that the overlay must be regenerated before the next deploy.
* Beyond #182: the rank-parity test gains the new wired knob as an asserted value on both ranks, the launcher validation gets its own CPU test, and the alias gate above.

## Checks run (CPU only; no hardware, no image build)

`bash -n start.sh` and `bash -n start-tp3.sh` clean; `git diff --check` clean; `tests/test_exl3_decode_pipeline.py` 13/13; `tests/test_exl3_routing.py` 2/2; `tests/test_launcher_rank_parity.py` OK (with the new `FAST=0` / `FAST=1` rank scenarios); the new `tests/test_numeric_config.py` case passes; `tests/test_tp3_contribution.py` 9/9 after the repin. A full CPU sweep of `tests/test_*.py` on this branch and on `b0e39f6` produces an identical non-pass set: 12 pre-existing failures on `main` (tests that need the in-image vLLM sources or a live server), of which `tests/test_numeric_config.py::test_tp4_rejects_retention_override` and `tests/test_start_overrides.py` are also red on `b0e39f6` and are untouched here; the image-only `tests/test_exl3_overlay.py` self-check is skipped in both. `tests/test_exl3_thin_fast_gpu.py` and `tests/bench_exl3_thin.py` need a GPU and are not part of these runs.

## Ask

Maintainer disposition on taking the thin path as an opt-in, defaults-off step. The numerical study is formally inconclusive-with-clean-relative-evidence; the current-`main` serving comparison is now measured (`+7.71% / +8.88% / +14.59%` median decode rate in the three range-resolved cells, two cells withheld, descriptive only), so speed and quality should be judged as two separate claims.


plotarmordev:
## Test summary before merge

We qualified the thin path before merging. Here is everything in one place.

### 1. Speed — measured by us, on current main

Same server image for all three runs. Fresh boot per run. Flag off → on → off.

| Workload | Off (A) | **Thin on (B)** | Off again (A2) | Gain (B vs A) |
|---|---|---|---|---|
| structured-1 | 72.74 tok/s | **78.36** | 72.52 | **+7.7%** |
| code-1 | 67.83 | **73.86** | 68.04 | **+8.9%** |
| json-1 | 51.32 | **58.81** | 49.50 | **+14.6%** |
| structured-2 | 68.86 | 78.07 | 71.67 | +13.4% (raw)* |
| prose-1 | 31.93 | 34.91 | 33.40 | +9.3% (raw)* |

*These two exceeded the tool's conservative overlap rule, so we report them as raw values, not verdicts.

Achieved throughput: +7.9% / +9.1% / +14.4%. Wave latency: −7.3% / −8.4% / −12.6%. 20 waves per arm, zero invalid captures, zero retries.

### 2. Answer quality — does it change what the model says?

818 requests, 18 fresh server boots, 10,828 scored positions, compared against a measured picture of the model's own run-to-run noise (stock vs stock):

| Check | Result |
|---|---|
| Consensus flips (systematic answer changes) | **0** (bar was 11) |
| Reproducible probability shifts | **0** (largest 0.054, threshold 0.529) |
| Per-family checks | pass, all families |
| Tie-flip rate vs stock's own | lower than stock (p=0.42) |
| Randomization cross-check (3,003 relabelings) | p=1.0 / 0.48 — nothing hides |

**One honest caveat:** our formal pass/fail rule could not issue an official "pass" — that same rule also fails when comparing the model against itself. So the correct claim is **"we found no difference beyond the model's own noise,"** not "certified identical."

### 3. Safety of the switch

Off by default. Turns on only with `GLM53_EXL3_MOE_FAST=1`, only on SM121 (GB10), and fails closed to the stock kernel anywhere else. Kernel-level evidence: same math, zero slow-memory reloads (stock: 39), 119/119 numerical parity checks, memory-sanitizer clean.

Kernel work by @rodman80 from #182 — credit preserved in the commits.

mrexodia:
Independent validation on two GB10 DGX Sparks, testing the merged thin path on `357fce7` with `GLM53_EXL3_MOE_FAST=0/1`.

**Result: 4–10% higher decode throughput than stock in this profile, with prefill essentially unchanged.** The existing cooperative kernel remained competitive in a separate one-boot reference.

### Matched serving results

Same locally built image/native module, weights, adapter and settings; stock/FAST environments differed only in the FAST flag. Order: **stock → FAST → FAST → stock**, two independent boots per arm. Values are the equal-weight mean of two within-boot medians: three repetitions per boot, except synthetic128k once per boot.

| Workload, tok/s | Stock | FAST | Change | Cooperative reference† |
|---|---:|---:|---:|---:|
| Prose C1 | 33.39 | 35.69 | +6.9% | 36.31 |
| Prose C2 aggregate | 32.61 | 34.83 | +6.8% | 35.80 |
| Prose C4 aggregate | 47.52 | 52.38 | +10.2% | 52.54 |
| Structured C1 | 73.69 | 80.27 | +8.9% | 81.56 |
| Poetry C1 | 21.63 | 22.50 | +4.0% | 23.49 |
| Code C1 | 27.61 | 29.04 | +5.2% | 31.64 |
| 32,753-token Pi: server decode | 26.37 | 27.44 | +4.1% | 29.32 |
| Same Pi capture: cold input/TTFT | 1410.87 | 1417.81 | +0.5% | 1405.57 |
| Synthetic ~32k cold prefill | 1450.26 | 1456.74 | +0.4% | 1452.27 |
| Synthetic ~128k cold prefill | 1443.46 | 1452.35 | +0.6% | 1449.89 |

† One preceding boot on the same updated image, existing cooperative binary/runtime, regenerated adapter. **Not an equally replicated third arm.** Cooperative and FAST were never enabled together.

Profile: TP2/CX7, DFlash2 k7 probabilistic/draftTP2, adaptive EMA `[2,4,7]`, dense+KDA FP8, grouped E3, fused cap32, batch7168, four slots, fixed14GiB KV/rank, vision on, ablation off, `mixed-prefill=skip`. This is a matched local profile, not untouched defaults. Configured context850k, **not validated to capacity**. All experimental arms used auto/safetensors loading; startup performance was not compared.

Prose400/structured200: temperature0/top_p1, thinking off, observed output lengths matched. Poetry/code512: temperature1/top_p0.95, seeds11/23/47. Exact Pi messages/tools and token IDs were invariant; synthetic cold-prefix nonces varied by design. Outputs/acceptance vary, so these are finite-sample estimates, not output-equivalence or significance claims.

For prose C1/C2/C4, server decode **ms per draft event** fell **93.33→86.06 /93.64→86.42 /133.66→120.46**. C1 accepted/event actually fell **2.157→2.078**, so its gain was not increased acceptance. This metric is summed request decode time/draft events, **not GPU kernel or verify-step tracing**.

Existing `skip` scheduling delayed concurrent starts: C2/C4 mean-request TTFT aggregates were **6.59→6.06s /9.81→8.74s**. Submitted concurrency is not continuously active decoding. Client aggregate throughput uses `sum(O−1)/(last output−first output)` across the wave; Pi decode uses attributable server timing to avoid tool-parser buffering inflation. Prefill is input/TTFT with verified zero cache hits.

### Functional and stress coverage

Across all five boots, including the reference and startup warmups: **533 completed inferences +8 intentional disconnects**, 541 HTTP200 submissions. Arithmetic, tool round trips, images, Responses, cached-usage reporting and positive repeat reuse passed before/after each boot. FAST additionally passed streaming/nonstreaming `tool_choice:none`, cancellation/recovery, and **60/60 owned large-session/recovery answers**. No observed CUDA OOM/illegal-memory/NCCL errors, unexpected restarts, or preemptions. Relevant per-wave and replay counters reconciled; 24/24 startup warmups and sampler-cache postcondition MET each boot.

- Exact Pi: 30 requests total, cold hits0 /repeat hits28,672.
- Larger Pi-derived histories: actual GLM inputs **114,276 /232,066**; three cold/repeat/repeat triplets each. Repeat reuse worked; generated tool calls were unscored, not executed.
- Four-session return/fork schedule: **495,425 aggregate history tokens**, 21 requests. Correct answers survived, but **all four post-burst returns lost usable cache hits**, requiring **83–91s full prefills** despite zero preemptions. This FAST-only stress result is **not evidence of a FAST-versus-stock cache regression**. Summary/foreground HTTP requests overlapped, but output intervals did not; the summary was not adopted as compaction.

### Native qualifications

On **each GPU**, the upstream119-case stock/FAST/stock comparator passed; filtered memcheck and synccheck reported zero errors. Additional production-cap32 boundaries, zero-weight clearing, fallback, graph/stream tests and dispatch profiling were retained.

**Not an all-clean native approval:**
- Stock and FAST racecheck each exited99 with **six warnings, zero errors**, in inherited Hadamard scratch reuse. No barrier fix deployed.
- An additional relative-L2≤1e-5 screen failed head rows1/33-uniform and worker33-uniform. Unchanged-stock repeats failed the same cases. Elementwise `1e-3 + 1e-3*abs(reference)` and upstream peak/repeat envelopes passed; the stricter screen remains failed, not retroactively relabeled PASS.
- Two broader host-test failures also reproduce on untouched upstream: TP4 retention rejection and all-env-key override/ABLIT expectations. Targeted thin-flag validation passed.
- Optional real-checkpoint parity was skipped in native-gate containers without the HF cache. Full-model serving is not a teacher-logit/task-quality or arbitrary-stream race-freedom proof.

An initial local-image packaging attempt failed before model serving because it retained a stale self-test. Updating that test to the matching source left native/adapter bytes unchanged; both-GPU image checks were rerun. No performance samples came from that failed attempt.

Overall this supports the advertised **incremental decode improvement over stock**, not a large upgrade over an already enabled cooperative profile. Original serving configuration restored afterward; no experimental kernel promoted. Raw prompts, histories and responses remain private.

PR #222, opened 2026-09-18
test(launcher): align TP4 retention checks with DFlash support

surlebeat:
## Description and Motivation

Merged #207 enabled DFlash SWA retention on TP3/TP4, but `tests/test_numeric_config.py::test_tp4_rejects_retention_override` still expects nonempty retention overrides to be **rejected** on TP=4. The numeric suite fails on current `main`:

```
$ python3 tests/test_numeric_config.py
AssertionError: ('0', '')
```

The validator returns 0 for `"0"` under `SPEC_METHOD=dflash`; the stale test expects 2. The failure is also acknowledged in merged #217's description.

The stale test is replaced by the current launcher contract:

- accepted values `""`, `"0"`, `"3584"`, `"14336"` → rc 0;
- malformed values `"-1"`, `"1"`, `"1.5"`, `" 3584"`, `"1003520"`, `"bad"` → rc 2 (`"1003520"` is 3584×280 — valid multiple but over the 1,000,000 maximum);
- an explicit SWA override requires `SPEC_METHOD=dflash`: `mtp`/`none` + `"0"` → rc 2, empty → rc 0.

Launcher and runtime behavior are unchanged — this is a test-only correction of the expected contract, not a claim of new TP4 GPU validation.

## Related Issues

- Aligns the test suite with merged #207 (DFlash SWA retention on TP3/TP4).

---

## Testing

- Reproduced the failure on this branch before the change: `AssertionError: ('0', '')`.
- After the change the full numeric suite passes: `numeric config tests: PASS`.
- `python3 -m py_compile tests/test_numeric_config.py` and `git diff --check` clean.
- CPU-only; no model launch involved.

---

## Checklist:

- [x] My change is reproducible and verified (a boot and, for behavior changes, measurements). — Test-only; verified by the commands above.
- [x] Defaults still work out of the box; a new knob has a sane fallback consistent with the existing ones. — No knob changes.
- [x] I updated the README (and docs/ where relevant) for any knob, default, or measured number I changed. — None changed.


sdougbrown:
`👾 AI Agent` — independent verification from a TP3 (3x GB10) deployment.

On upstream/main `ca85576`, `python3 tests/test_numeric_config.py` fails at the first case: `AssertionError: ('0', '')` — `start-tp4.sh` forwards both retention knobs, so the `("0", 2)` / `("14336", 2)` expectations no longer hold.

Checked out this PR's head and ran the same command inside `ghcr.io/miaai-lab/glm-5.3-flash-2x-dgx-sparks:exl3-instanttensor` (python 3.12, no GPU required): `numeric config tests: PASS`.

The `SPEC_METHOD="$3"` parameterization is the part our local fix missed — without it the retention validation exits 2 for the wrong reason, so the acceptance cases could not distinguish the knob rule from the dflash requirement. The broader malformed matrix (`-1`, `1.5`, leading space, `bad`) also covers cases we had not asserted. LGTM from our side; no regressions in the rest of the file.

surlebeat:
Superseded upstream: 9030573 ("test: restore launcher and APC CPU regression coverage") rewrote tests/test_numeric_config.py with this PR's coverage, generalized — validate_retention() now takes spec_method and test_swa_retention_interval_needs_the_dflash_drafter covers the SWA-requires-DFlash contract across all three launchers (start.sh / start-tp3.sh / start-tp4.sh, per #207). Closing as absorbed. Thanks!

PR #233, opened 2026-09-19
perf(sm121): speed up KDA large-M prefill with BF16 GEMM

rodman80:
## Summary

Large-M KDA prefill path: `KDA in_proj` keeps stock FP8-Marlin W8A16 for M <= 512 and serves larger-M prefill from a load-time BF16 copy of the same logical FP8 weight (e4m3 tensor x the stored Marlin scale) via ordinary BF16 GEMM. Opt-in with `GLM53_KDA_BF16_LARGE_M=1` (default off). The measured dispatch boundary is fixed at M>512: small-M/decode stays on Marlin and large-M prefill uses the retained BF16 logical weight.

## Motivation

FP8-Marlin W8A16 is good at small M but slow for the observed KDA prefill geometry (decode M=1..220, prefill M>=768, band 221-511 empty). PR #182 explored a W8A8 large-M path: fast, but it produced reproducible full-model probability shifts at its M>64 boundary, and attribution isolated the cause in BF16->FP8 activation quantization (not the GEMM backend). This PR removes that term instead of fixing the backend: activations stay BF16.

## Implementation

- `overlay/exl3.py`: `kda_bf16_large_m_enabled/logical_weight`, chunked (512-row, ~8 MiB scratch) load-time BF16 materialization using the STORED dtype-rounded scale captured before Marlin prep, tiny-M cuBLAS probe, metadata-only dispatch, fixed-512 boundary stored on the layer, production dispatch counters.
- Fail-closed at load for a KDA in_proj layer with the feature on: TP != 2, non-SM121 capability, unexpected shape/dtype all raise instead of silently running Marlin. Small M stays Marlin by design (fallback, not an error).
- The 512 boundary is intentionally fixed, not user-configurable: it is the configuration actually measured and qualified on TP2/GB10, and 512 is not claimed optimal for other hardware.
- `start.sh`: the single boolean knob with pre-stop validation and both-rank forwarding.
- No W8A8 code: no `_scaled_mm`, no rowquant, no Triton kernel, no raw-FP8 retained copy. No TP3 changes, no cooperative-EP changes, no THIN changes.

## Dispatch scope

TP2-local KDA `in_proj_qkvbfg_a [12576x4096]` only. M=512 -> Marlin, M=513 -> BF16; ordinary decode shapes remain Marlin; bias/odd-K/missing-retention fall through to Marlin.

## Numerical evidence

- Layer level (real layer-0 weights, rel RMSE vs Marlin, M>=768): old W8A8 ~0.0266, fp8-x-fp32-scale 0.002938, shipped fp8-x-stored-bf16-scale **0.002526** (~10.5x closer than W8A8); M=63/64/65/66 exact continuity (all Marlin).
- Prospective stock / W8A8-control / BF16 study: W8A8 reproduced the old boundary failure (M=65: 7.99x stock-null mean KL, 23 coordinated flips; a reference token 0.778->0.0051) while BF16 did not (M=65: 1.10x, 0 flips; M=66: 0.88x, 0 flips; same position 0.768-0.782 alongside stock). Remaining BF16-vs-stock differences sit inside measured stock/runtime variation.
- Engineering evidence, not an equivalence claim: no mathematical/bitwise equivalence asserted.

## Performance (TP2 / 2xGB10)

- KDA in_proj large-M: BF16 ~3.1-3.7x faster than Marlin for M>=768 (at M=7168: Marlin ~27.21 ms, BF16 ~7.36 ms).
- Serving, same image, flag-only arms: cold ~16k TTFT 11.41 s -> 10.12 s (-11.4%); cold ~100k 70.86 s -> 62.12 s (-12.3%); mixed C3 + cold ~100k prefill TTFT 76.21 s -> 65.43 s (-14.1%); decode unchanged (same Marlin path).

## Memory cost

Retained copy 98.25 MiB/layer/rank x 34 KDA layers = +3.26 GiB/rank theoretical, **+3.29 GiB/rank measured** (model load 79.65 -> 82.94 GiB). KV reservation and the tested 850k context remain unchanged (883,552 KV tokens, 552 blocks), while idle headroom is reduced approximately 7.7 GiB -> 4.7 GiB.

## Validation

- `tests/test_kda_bf16_large_m.py` (23 CPU tests): flag validation, fixed-512 constant, chunk-invariant construction, stored-vs-fp32-scale regression, fail-closed retention, fixed M<=512/M>512 dispatch table (incl. the old 63/64/65/66 boundary), counters.
- `tests/test_kda_bf16_large_m_gpu.py`: retention (98.25 MiB, bf16, threshold 512), M<=512-all-Marlin / M>512-all-BF16 with exact row counts, BF16-vs-Marlin relmax ~5.5e-3, m8 bitwise parity, graph capture/replay bitwise on both branches, fail-closed (bad flag/TP=1).
- `tests/bench_kda_bf16_large_m.py`: Marlin-vs-BF16 numerics + crossover timing receipt (real weights: M<=512 bitwise 0.00000, M>=768 0.26-0.30x Marlin time, rel RMSE 0.00253).
- Clean-port smoke on the built image: 34/34 retentions, KV unchanged, cold-prefill in range, boundary-text logprobs sit with stock/v2 (not with the W8A8 pathology scale).
- Launcher parity + pre-stop validation tests updated for the knob.

## Scope and limitations

Qualification scope: TP=2 homogeneous GB10/SM121. TP3 is intentionally out of scope because it has different KDA sharding/head-padding geometry and was not testable on this cluster. Not claimed: equivalence, universal quality preservation, zero memory cost, all-workload improvement, formal numerical certification. The maintainer's exact original study texts were not available locally; running them against this draft is valuable independent follow-up.

## Relationship to #182

#182 remains the historical research/evidence thread (W8A8 exploration, attribution, A/B/C campaign) and is untouched by this PR. This draft contains only the BF16 large-M implementation. Decode work from #217 is separate and unmodified here.


plotarmordev:
## Qualification results and completed integration

Reviewed head: `c9754afa84b09d6043158926990b30a360df7965` (GPU-study head; the integration fixes below were checked at the newer head).

**Conclusion:** the BF16 large-M path is real, the isolated long-prefill speedup is measurable, and the completed testing did not establish a general answer-quality regression. Merged as **experimental and default off** after the integration fixes and CPU regression repairs below, without another broad GPU benchmark campaign. The integration fixes and exact-head validation results are recorded below.

### Built and exercised

- Verified actual dispatch on both TP2 ranks: eligible projections at scheduled M=513/768/6656 used genuine BF16 GEMM; the tested M<=512 cases used Marlin. Stock is FP8 weight-only / BF16-activation Marlin, not W8A8. A 7168-token input was scheduled as 6656+512, so input length and GEMM M must not be conflated.
- Isolated the first observed candidate-specific difference at layer0 `in_proj_qkvbfg_a` with matching observed input rows. The retained weights are FP8-derived BF16 weights, not original checkpoint BF16 weights.
- Completed a seven-arm remedy screen, component replays, then an eight-boot full campaign: four fresh calibration stock boots, the original PR, two independent boots of a lower-memory late-scale alternative, and a held-out stock boot. Each boot covered 46 numerical cases with four captures. Five full arms also covered 18 prefill fixtures with three measured repeats, 24 objective fixtures repeated twice, and fixed 128-token generation at c1/c4.
- Completed a separate five-arm speculative-decoding workload comparison, including cached multi-turn requests and a 229356-token retrieval case. All five arms returned 15/15 conversation checksums and passed that retrieval case. This is limited functional evidence, not a broad long-context reasoning claim.
- Graded 320 generated programs across ten arms on a frozen, padded 32-task HumanEval subset, with unchanged task tests in isolated CPU containers. Canonical validation passed 32/32; there were no grading infrastructure errors.

### Performance

For a 100000-token code prompt with a one-token response, median client request wall time was:

| Implementation | Seconds |
|---|---:|
| Stock, two bracketing boots |70.84 /71.97|
| This PR |62.24|
| Lower-memory late-scale alternative, two boots |66.34 /66.56|

This is approximately 12–14% less request time for this PR on that workload, **not** a universal decode-throughput or streamed-TTFT claim. The PR retains approximately 3.26GiB/rank of extra weights for the qualified geometry. Cached/concurrent workloads can lose some of the benefit through reduced cache headroom.

### Answer-quality review

| Profile | Stock controls | PR233 | Late-scale alternative |
|---|---:|---:|---:|
| Isolated profile |25/32,23/32|24/32|24/32,22/32|
| Speculative-decoding workload profile |28/32,26/32|26/32|28/32|

These are **not standard HumanEval pass@1 scores**. The prompts were padded and responses capped at 512 output tokens.

Inspection of the saved responses found:

- All 14 PR233 failures across the two profiles hit the output cap and failed Python parsing. Every PR233 response that ended normally passed the saved tests. This does not imply capped responses would necessarily pass with more tokens.
- In the isolated profile, PR233 gained `HumanEval/139` and `/134` against both stock controls, and lost `/114` against both; that loss was a capped, unfinished code-block response.
- In the speculative-decoding profile, PR233 had exactly the same per-task pass/fail outcomes as the ending stock control, with no losses on tasks both stock controls passed.
- The two stock controls disagreed on 8 tasks in the isolated profile and 4 in the second profile.
- There were no tasks both late-scale boots lost against both stock controls, and no tasks both late-scale boots gained against both controls. Cleaner component arithmetic did not produce a repeatable answer-quality advantage.
- Some alternative/stock answers had genuine logic, missing-import, or generated-self-test errors. A complete but incorrect function can coexist with a truncated explanation; truncation is not an automatic excuse.
- One alternative failure involved `doctest.testmod()` inspecting the evaluator's module rather than the detached execution namespace. Its original score is preserved, but that failure should not be presented as evidence of an incorrect algorithm.

### Numerical interpretation and deviations

The unchanged frozen numerical comparator also flagged the held-out **stock** boot (`E-S02`, `E-M06`). The global compatibility result is therefore **inconclusive**, not a candidate failure. No thresholds were retuned. Specific observed logprob differences remain evidence of different computation, not proof of worse answers.

Post-accumulation scaling reduced the isolated added projection error beyond final BF16 rounding by 14–22x in component replay. A compact retained-FP8 implementation halved persistent extra weight storage and matched the retained-BF16-code late-scale implementation on 22/22 full-M replay outputs. Those findings concern arithmetic and memory, not model accuracy; they do not justify a second maintained quality-safeguard mode at present.

The original workload profile hit a predeclared swap-growth safety limit on stock. A separate lower-memory-budget workload campaign completed, but preemptions occurred in candidates and the ending stock control. That supplemental run does not qualify the original profile or provide a clean no-preemption performance comparison.

### Integration fixes and merge

Normal-merged at `d4bd9eb4f2607b5d79a85c5d141281b3e13b2dfa` from tested head `90305735f7e9090e1bcf2c233af5a222584bac69`. The merged main tree is identical to the tested tree. The feature remains **experimental, BF16-only, and default off**; no production setting was enabled.

- Isolated CPU tests load only the required helper AST nodes, not the unrelated Torch-dependent KDA region.
- The enabled large-M path rejects FP16 at load; disabled FP16 retains its existing path. Restoring the old guard makes the new rejection test fail as expected.
- Reviewed and updated the TP3 cooperative-overlay source pin. The existing `Exl3MoEMethod` and TP3 Marlin exclusion helper are AST-identical to the pre-merge main. This preserves TP3 setup compatibility, not qualification of the new BF16 path on TP3.
- Repaired the broader CPU suite rather than waiving its failures: launcher tests now load their real helper dependencies, NIC memory inputs are deterministic fixtures, ABLIT expectations match caller-export precedence, and retention tests cover the supported interval/DFlash contracts.
- APC has a proper pytest entry point that executes all 12 original case bodies, including both-order composition and eviction behavior. Validation used pristine vLLM sources at `487ecf187d3dfe74d2cf6119a92881dba403c219`; no behavioral assertions were removed from those cases.
- Updated feature documentation, APC test instructions, and the changelog. No model reruns or service changes were made for these integration fixes.

**Exact-head verification:**

- `python3 -m pytest tests -q -rs -p no:cacheprovider`: **299 passed, 3 skipped, 70 subtests passed**, with the pinned APC source inputs supplied.
- The three skips require scheduler source installation. The scheduler's supported direct-run entry point installed the patch and passed **all 30 tests**, including those three cases.
- All **25 KDA CPU/Torch tests** passed without skips.
- The real-source dense-FP8 patch installation/idempotence script and launcher shell syntax checks passed. TP3 contribution and rank-parity tests are included in the passing CPU suite.
- Secret scanning found no leaks in the complete PR or newly pushed commit ranges. Worktree matches were verified as existing token hashes, a configuration-key name, and function arguments—not credentials. No deployment-local content was included.

The earlier **8 failures and 11 setup errors were repaired, not accepted as a baseline exception**.

### Skipped / honest gaps

No additional GPU campaign was run on the integration-fix head; GPU observations above belong to the explicitly identified study head. No 900000-token input qualification, no FP16/TP3/other-hardware qualification, and no general accuracy-equivalence claim. The numerical comparator remains inconclusive because it also flagged held-out stock. Code-task evidence is limited to 32 tasks with substantial output truncation and one PR boot per profile.

GitHub reported no hosted check runs or reviews. That is not a green hosted-CI result: the merge used the exact-head local verification above. Follow-up: we will run hosted CI against merged main when it is available.

plotarmordev:
working on the fixes before merge and merging when ready @rodman80, thank you again!

PR #235, opened 2026-09-19
feat(tp4): forward adaptive verification length (GLM53_ADAPTIVE_K*) on start-tp4.sh

100menotu001:
## Description and Motivation

`start.sh` has carried adaptive verification length (`GLM53_ADAPTIVE_K=ema` and its six companion knobs, *Faster prose decode*, 2026-09-08) with a capture-size generator that adds every enabled query length at every batch size. `start-tp4.sh` never forwarded the knobs, mounted `overlay/patch_adaptive_k.py`, or extended the capture list, so a four-node kit that set `GLM53_ADAPTIVE_K=ema` got stock k on every rank with no error and no log line.

This PR wires the overlay on the four-node launcher exactly the way the other TP4 overlays are wired (and the way #224 did for the slice): host path, existence check, scp to ranks 1–3, read-only mount, container-start application on head and workers, env forwarded to head and workers, and the identical capture-size generator from `start.sh` (identity when `off`). No new knob, no new default; `GLM53_ADAPTIVE_K` stays `off`. `GLM53_DENSE_FP8` is deliberately not wired (numerics still marked provisional in the README).

**Measured effect on a 4× DGX Spark kit** (see Testing): with `ema` the drafter still proposes 7, the scheduler verifies 3/5/8-token prefixes, and prose acceptance ratio goes 0.34 → 0.50 with fewer wasted verify positions. Single-stream decode moved structured 86.9 → 100.5, coding 57.5 → 63.8, prose 41.0 → 42.7 tok/s across a boot that also switched mixed prefill `skip` → `fair` and set `GLM53_APC_RETENTION_INTERVAL_SWA=0` (both already the `.env.tp4.example` defaults), so the decode delta is not attributed to adaptive-k alone. The claim of this PR is that the knob now reaches every rank on TP=4; the speed claim is the README's TP2 one.

## Related Issues

- #224 (same launcher wiring pattern; measurement boot ran both)
- Closes nothing; no issue was filed because the knob silently did nothing on TP=4 rather than failing.

---

## Testing

CPU, on this branch:

```
$ python3 tests/test_tp4_adaptive_k_wiring.py
ok   test_capture_sizes_ema_covers_every_k_at_every_batch
ok   test_capture_sizes_off_is_stock
ok   test_launcher_syntax
ok   test_recipe_wiring
$ bash -n start-tp4.sh && echo OK
OK
```

The test checks: wiring on both rank scripts, the seven defaults byte-equal to `start.sh`, generator `off` == the stock TP4 list `1 2 4 8 16 24 32`, generator `ema` (k-set 2,4,7, `DFLASH_TOKENS=7`, `MAX_NUM_SEQS=8`) contains n×3, n×5, n×8 for n=1..8.

**Boot on a 4× DGX Spark kit** (this branch merged with #224, `:exl3` image `9581c4c74257`, `SKIP_BUILD=1 ./start-tp4.sh restart`, `GLM53_ADAPTIVE_K=ema GLM53_ADAPTIVE_K_SET=2,4,7`, E3 grouped, fused MoE, graphs on, `DFLASH_TOKENS=7`, `MAX_NUM_SEQS=8`, MNBT 2048, 1M ctx):

```
head: patched scheduler.py (GLM53_ADAPTIVE_K=ema)      (identical on w1, w2, w3)
patched cudagraph_utils.py (adaptive-k graph lengths)
[glm53-adaptive-k] uniform decode graph query lens: [3, 5, 8]
cudagraph_capture_sizes: [1, 2, 3, 4, 5, 6, 8, 9, 10, 12, 15, 16, 18, 20, 21, 24, 25, 30, 32, 35, 40, 48, 56, 64]
GPU KV cache size: 5,126,984 tokens, Maximum concurrency for 1,000,000 tokens per request: 5.13x
/health 200 at +440 s
```

`tests/bench_decode.py`, 3 runs × 400 tokens, temperature 0, thinking off, quiet serve:

| | `off` (previous boot, same kit) | `ema` |
|---|---:|---:|
| Structured tok/s median (accept / per step) | 86.9 (0.959 / 6.71) | **100.5** (0.959 / 6.71) |
| Coding | 57.5 (0.632 / 4.43) | **63.8** (0.671 / 3.93) |
| Prose | 41.0 (0.342 / 2.40) | **42.7** (0.503 / 1.98) |

Caveats stated for the record: the `off` column was measured on the previous boot, which also had mixed prefill `skip`; one agent session was live during part of that run (structured min was 62 tok/s, so its median is depressed by contention). Cold prefill on the `ema` boot: 1,553 / 1,547 / 1,495 tok/s at 11k / 55k / 191k tokens, in line with the README's TP4 numbers; adaptive-k does not touch prefill.

---

## Checklist:


100menotu001:
Rebased onto main after #224 merged (same adjacent launcher lines); resolved by keeping both overlays. Added `test_rank_scripts_syntax`: `bash -n` on the launcher cannot see inside the quoted heredocs that become each rank's /start.sh, and a dropped `fi` there only appears as `/start.sh: syntax error: unexpected end of file` at boot, so the test now checks those bodies too.

Issue #106, opened 2026-09-02
Prefix cache stops hitting globally after ~1 h of concurrent agentic load (hits_total freezes, only a restart fixes it) — ABLIT=0 and 1


vadiksext:
## Summary

After ~50–60 min of concurrent agentic load the prefix cache stops hitting **globally**: `vllm:prefix_cache_hits_total` freezes and never moves again, every request (including a byte-identical prompt re-sent three times) is re-prefilled from token 0, `Running` drops to 1 with the rest `Waiting/Deferred` while KV usage sits at ~20 %. No error in head/worker logs, `num_preemptions_total=0`. The only thing that restores hits is a restart (there is no reset endpoint, #31). Seen twice in one night (clean case with `ABLIT=0`, 3 clients; earlier with `ABLIT=1`, 4 clients). Same symptom as #7 — but it only shows after ~1 h of continuous load, so the 4-request retest that closed #7 cannot see it.

## Environment

- 2× ASUS GX10 (GB10), TP=2 over RoCE, recipe `c190db1`, image built locally per README (`EXL3_FAT_KERNEL=1`, E2 kernel tier `kernel_ok`)
- `.env`: `MAX_MODEL_LEN=131072 MAX_NUM_SEQS=4 MAX_NUM_BATCHED_TOKENS=2048 GPU_MEM_UTIL=0.87 GLM53_INDEXER_WORKSPACE=rightsize GLM53_SPINWAIT_MS=16 GLM53_MIXED_PREFILL_CHUNK=skip SPEC_METHOD=dflash DFLASH_TOKENS=7 DFLASH_DRAFT_TP=2`
- boot log: `kv_cache_size_tokens=502059`, `kv_cache_max_concurrency=3.83`
- clients go through litellm on the head; `chat_template_kwargs.enable_thinking=false`, `temperature=1.0 top_p=0.95 max_tokens=16384`, tool calling (one `bash` tool)

## Workload

SWE-bench-style agent, 3 concurrent clients. Every step re-sends the whole conversation (identical prefix + 1–10k new tokens: tool call + tool output), contexts grow from ~5k to ~100k tokens over 100–200 steps, one request every 5–20 s per client. Nothing else touches the server.

## Timeline (stock weights, `ABLIT=0`, UTC)

| time | `Prefix cache hit rate` (cumulative, from `loggers.py`) | completions / min |
|---|---|---|
[...]

PR #238, opened 2026-09-20
perf: compact DFlash2 shared KV pages without further quantization

plotarmordev:
## Current checkpoint — fixes pushed, default OFF

Reviewed fixes and qualification documentation are pushed at `bdf6b9a` (tested runtime source `9a3aca4`). **Draft; not merged; default remains OFF.** [Consolidated measurements, performance comparisons, limitations and deferred investigation](https://github.com/MiaAI-Lab/GLM-5.3-Flash-EXL3-2x-DGX-Sparks/pull/238#issuecomment-5760233232). Stock stress qualification remains incomplete; custom success is not a blanket stock approval. Further investigation is deferred by the author.

---

## Description and Motivation

Add `GLM53_DRAFT_KV_COMPACT=1` (default `0`) to use shared MLA pages more efficiently for DFlash2's sliding-window cache. This is a cache-layout optimization, not additional quantization. It leaves weights, KV dtype, window length, target cache groups, and the PR233 KDA path unchanged.

Instead of copying a fixed draft block size, derive the largest multiple of 64 that both divides the MLA block and fits the physical page. This preserves prefix-cache alignment. Reject padded-page virtual splitting against the actual attention backends during kernel selection; applying a full manager-page stride to each smaller kernel block is unsafe.

The existing drafter patch now preflights both allocator and worker sources before writing either, writes the worker guard first, and remains idempotent. Consolidating its repeated write/parse branches keeps the overlay smaller despite the added behavior. TP2/TP3/TP4 launchers validate and forward the opt-in; defaults stay unchanged.

### CPU allocator result — not a serving benchmark

For a 3,584-token MLA block, 656-byte MLA records, and a 2,048-byte/token draft:

| Property | Existing padded layout | Compact opt-in |
|---|---:|---:|
| Draft block size | 64 | 896 |
| Draft admission bound, 2,048-token window + 2,048 in-flight tokens | 65 block IDs/request | 6 block IDs/request |
| Prefix alignment | 3,584 tokens | 3,584 tokens |
| Backing tensor allocations and pool block count | unchanged | unchanged |

This reduces block-ID demand; it does not free the preallocated backing tensors, establish a new concurrency limit, or demonstrate a tok/s gain.

The direction was motivated by [Alexbob0's draft-page sizing work](https://github.com/Alexbob0/glm53-flash-vllm-upstream-sm121/blob/9bf39c3e84194a57c630a42c0d79066159a5b787/overlay/patch_kv_drafter_group.py#L64-L83). This implementation adds geometry-derived selection, divisibility validation, and backend safety checks for the current recipe rather than transplanting that configuration.

## Related Issues

Independent of the weight/execution changes in #233. No changes to adaptive-k, cooperative MoE, quantization, or serving defaults.

## Testing

Exact head `2945711`, based on `775a58b`:

- **29 passed**: `tests/test_draft_kv_compact.py`, `tests/test_launcher_rank_parity.py`, `tests/test_start_overrides.py`, and `tests/test_numeric_config.py`.
- The new suite ran with `GLM53_VLLM_SRC` set to hash-checked pristine vLLM `487ecf187d3dfe74d2cf6119a92881dba403c219` sources. All five new tests ran; none skipped. It executes actual cache-spec, grouping, allocation, and backend-selection definitions without importing torch or vLLM's GPU runtime.
- Geometry sweep checks page fit, maximum eligible block size, and unchanged LCM. Pinned-source tests cover the reservation reduction, unchanged backing allocations, invalid alignment, backend split rejection (including an explicit stride with zero extra padding), the unpadded exception, dry-run, idempotence, and cross-file preflight failure. The explicit-stride regression failed before the guard fix and passes afterward.
- `bash -n start.sh start-tp3.sh start-tp4.sh` passed.
- Additional isolated launcher smoke: both TP2 ranks receive `0`/`1`; invalid and explicitly empty values fail before any stubbed host action. All three TP3 and all four TP4 stubbed rank launches receive the opt-in. TP3/TP4 caller precedence, including explicit empty values, passed. Every host command in these launch checks was stubbed.
- Additional CPU byte-buffer smoke exercised the pinned padded-view function using NumPy-backed operations: a compact draft-page write left neighboring slots and padding untouched. This is not CUDA-kernel verification.
- Gitleaks exact commit-range scan: no leaks. Full-worktree scan reported 32 pre-existing false positives, verified unchanged from the base: token-ID SHA256 fields and two configuration/argument identifiers. No deployment-local files or values are included.

Reproduce the new suite with the three source files listed in its module docstring:

```sh
GLM53_VLLM_SRC=/path/to/pristine-vllm-source python3 -m pytest -q tests/test_draft_kv_compact.py
```

Without that path, the two geometry/configuration tests run and the three pinned-source tests explicitly skip.

## Live qualification — A/B/A control

Operator-authorized testing is now complete. A1 = freshly restarted original layout; B = compact layout; A2 = restored original layout. B used the exact allocator patch from `2945711` plus minimal flag forwarding on the existing serving baseline, not a wholesale deployment of the upstream branch. Container image, weights, quantization, cache dtypes, context/concurrency limits, batching, and memory-utilization settings were held fixed. Both ranks booted successfully with compact blocks; the original deployment was restored and passed its health check afterward.

There were **108 completed requests**, 36 per phase, with identical prompt hashes across phases. Short decode and 8k/32k prefill panels use medians of three repetitions; concurrent decode uses three two-request rounds. Long-context/reuse pairs and the sustained-generation check ran once per phase.

| Measurement | A1 baseline | B compact | A2 restored baseline |
|---|---:|---:|---:|
| Full-context reservation bound, block IDs/request | 176 | 117 | 176 |
| Structured decode, tok/s | 78.54 | 79.85 | 78.44 |
| Code decode, tok/s | 52.51 | 52.78 | 51.91 |
| Prose decode, tok/s | 31.43 | 32.05 | 31.86 |
| Two-request code aggregate, tok/s | 75.20 | 73.32 | 75.80 |
| 8k cold-prefill TTFT, s | 7.296 | 7.420 | 7.291 |
| 32k cold-prefill TTFT, s | 27.386 | 28.513 | 27.448 |
| 100,701-token cold TTFT, s | 84.140 | 90.162 | 85.085 |
| Same 100k prompt, first reuse TTFT, s | 1.026 | 3.831 | 1.041 |
| Same 100k prompt, cached tokens | 100,352 | 96,768 | 100,352 |
| Two near-243k cold requests, total wall time, s | 415.726 | 432.883 | 427.489 |
| Their concurrent follow-ups, total wall time, s | 8.347 | 5.530 | 8.358 |
| Sustained 3,200-token generation, decode tok/s | 70.35 | 74.94 | 72.69 |
| Aggregate speculative-token acceptance | 68.07% | 68.69% | 67.16% |

**Interpretation:** The 33.5% lower reservation bound is confirmed. There are observed gains on the long-context follow-up pair and sustained generation, but this is not a general serving speedup: short decode differs by roughly 1–2%, concurrent code throughput is roughly 2.9% lower, and 32k cold TTFT is roughly 4.0% higher relative to the mean of the baseline medians. Most importantly, the first 100k reuse loses one 3,584-token prefix block and takes about 3.7x as long; the restored baseline recovers the original hit and latency. Preserving the LCM does not guarantee identical prefix-hit coverage.

All **45 explicit exact-answer checks** passed, including long-context retrieval and follow-ups, arithmetic, and sorting. All three sustained-generation checks passed the consecutive-integer check across the draft-window rollover. No request failed and no preemption was reported. Host swapout was zero in both baseline phases; B recorded 30 pages (120 KiB) on one host. Free-form text varies within the baseline as well as between phases, so these checks do not establish numerical equivalence or broad model-quality parity.

Capacity caveat: usable pool counts differed across fresh memory profiles (516 / 513 / 498). Five-second sampled peak occupancy was 58.5% / 41.7% / 33.3%; scheduling and sampling make those peaks unsuitable for a runtime-capacity claim. No higher-concurrency capacity increase or reduction in preallocated backing memory is demonstrated. The two near-243k cold requests were submitted together, but their cold prefills were effectively serialized. This is one candidate boot, not a statistical performance study; other attention backends and higher-rank GPU deployments remain untested.

Raw receipts and deployment-specific details remain private. No deployment-local configuration is included in this PR.

## Qualification hold

**Draft; not ready to merge. Keep the option off in the original deployment.** Live boot and functional checks passed, but the repeatable first-reuse regression above prevents presenting this as an unqualified performance improvement.

- **Next-action owner:** PR author @plotarmordev, to evaluate the lost prefix block and propose a fix or an explicitly documented capacity/latency tradeoff.
- **Unblock condition:** the prefix-reuse issue is resolved, or the tradeoff is explicitly accepted in review; any chosen implementation change is verified at its new head. No fresh measurements or rebase are requested before that direction is chosen.

## Checklist

- [x] Changes are consistent with the current recipe and preserve existing defaults.
- [x] README, configuration example, and changelog updated.
- [x] CPU behavior and launcher boundaries verified at the exact head.
- [x] Authorized GPU boot, functional checks, acceptance, prefix reuse, and matched serving measurements completed; original deployment restored.
- [ ] Prefix-reuse tradeoff resolved or explicitly accepted; qualification hold cleared.


plotarmordev:
> **Publication update:** fixes are now pushed at `bdf6b9a`; the local/unpublished wording below is historical. [Current consolidated checkpoint](https://github.com/MiaAI-Lab/GLM-5.3-Flash-EXL3-2x-DGX-Sparks/pull/238#issuecomment-5760233232). Default OFF; draft; no merge.

## Fresh custom OFF / ON / OFF qualification after the stock-installer correction

Tested runtime source: **`9a3aca4c0a5ad3a90b22ceb83dc71b3d0fa24f46`**, local/unpublished. Published PR head remains `2945711d8b7d563542cb3e48914b4fb4b12595cd`. The legacy stock-coordinator migration is now fixed locally; stock results and remaining limits are reported separately, not inferred from this custom run.

The original custom deployment's image, precision and serving configuration were preserved. Container environments differed only in `GLM53_DRAFT_KV_COMPACT`; no stock 14 GiB override leaked into these runs. Each arm had a fresh boot. **51 prompt hashes matched across all three arms: 153/153 requests completed, 90/90 marker/reference answers were correct, all three 3,200-token rollover sequence checks passed, and no preemptions or safety stops occurred.** Minimum sampled available RAM across the three arms was **4.73 GiB**.

Percentages below compare ON against the arithmetic mean of the two OFF measurements. Lower is better for times; higher is better for throughput. Decode, C2 and short-prefill measurements are medians of three trials.

| Measurement | OFF 1 (K0) | ON (K1) | OFF 2 (K2) | ON vs mean OFF |
|---|---:|---:|---:|---:|
| Reservation IDs per configured maximum-length request | 176 | 117 | 176 | **−33.52%** |
| Usable pool IDs after automatic profiling | 507 | 519 | 514 | Not a fixed-budget comparison |
| Cached/prefix-hit tokens on 100,701-token repeat | 100,352 | 100,352 | 100,352 | Equal |
| Repeat TTFT (s) | 1.043 | 0.728 | 1.052 | **−30.46%** |
| Cold 100,701-token TTFT (s) | 89.109 | 87.489 | 84.729 | +0.66% |
| Two near-243k cold requests, pair wall time (s) | 432.028 | 423.355 | 418.867 | −0.49% |
| Their two follow-ups, pair wall time (s) | 8.427 | 5.551 | 6.735 | **−26.78%** |
| Structured decode (tokens/s) | 74.53 | 79.31 | 78.33 | +3.77% |
| Code decode (tokens/s) | 51.23 | 50.45 | 49.64 | +0.03% |
| Prose decode (tokens/s) | 38.81 | 30.72 | 32.19 | **−13.46%** |
| C2 code aggregate (tokens/s) | 74.68 | 72.60 | 81.54 | **−7.05%** |
| 8k cold TTFT (s) | 7.654 | 7.121 | 7.342 | −5.03% |
| 32k cold TTFT (s) | 28.182 | 28.011 | 27.428 | +0.74% |
| Rollover decode (tokens/s) | 71.06 | 74.41 | 75.64 | +1.44% |

The enabled arm retained all **28,672 prefix tokens** at each tested tail length: 64, 65, 349, 896, 897 and 2,048. Both OFF arms retained only 25,088 at tail 64. Concurrent changed-suffix references and the original-prompt reference afterward were correct. Both near-243k cold requests and their follow-ups completed correctly; cold prefills were effectively serialized, so this does not establish parallel long-prefill capacity.

**Interpretation:** the repeat-TTFT benefit reproduced, but there is no universal decode speedup. Prose and C2 throughput were lower. Freeform output sequences differed, and OFF baselines drifted (about 17% for prose, 9% for C2 and 20% for long-follow-up wall time). These are descriptive measurements, not identical-output kernel timings or statistical significance claims. Automatic profiling also selected different pool sizes; a lower reservation-ID requirement is not a measured reduction in backing tensor memory or VRAM.

The migration correction preserves the previously correct generated custom coordinator bytes: SHA-256 `209d3a88ea93dcbe808a7b0cb0655c299da216e2f5dbcb5b690bf01f9938d50c`. The earlier `6d5dd89` A3/C/A4 experiment remains separate and byte-frozen: neutral receipt SHA-256 `35fd5ef14b9e9502116311c5706e0ae874056369f90e38ba2184f2be3257e7e2`; all 96 artifacts in its manifest were rechecked unchanged. These new measurements are not a replacement for failed stock runs.

**Limits:** this is qualification of the exercised custom workload only. Actual stock tests exposed automatic-budget OFF startup capacity rejection, near-limit reuse preemptions, a 14 GiB ON memory-guard stop, and an OFF sequence-reference failure. TP3/TP4 GPU behavior, other backends/architectures and tensor-level numerical parity remain unqualified. Default remains off. No push or merge is implied.

Original service restored and checked after all runs: original image, environment, launch command, mounts and patch hashes on both ranks; five original source hashes unchanged; health and a live inference smoke passed.

Neutral result receipt SHA-256: `de9dc16deb0aafbbe60bc469d71cd250a988287c52e2f069f62537f6f4e79b46`. It binds 419 private raw-evidence files through manifest SHA-256 `9e901d5db485e1c42e3b2030e080be0b96de7c19e05821ec9b4f9e76fd7aaf0a`. Raw deployment details remain private. Documentation updated locally in `bdf6b9a667bbbdade6a6eadda1de54d4253ebd72`.

[Stock-default and adjusted-stock results, including every failed arm](https://github.com/MiaAI-Lab/GLM-5.3-Flash-EXL3-2x-DGX-Sparks/pull/238#issuecomment-5759982309).

plotarmordev:
> **Publication update:** fixes are now pushed at `bdf6b9a`; the local/unpublished wording below is historical. [Current consolidated checkpoint](https://github.com/MiaAI-Lab/GLM-5.3-Flash-EXL3-2x-DGX-Sparks/pull/238#issuecomment-5760233232). Default OFF; draft; no merge.

> **Update:** the legacy-coordinator installer incompatibility is fixed in local source `9a3aca4` (not yet in the published PR head). [The follow-up report](https://github.com/MiaAI-Lab/GLM-5.3-Flash-EXL3-2x-DGX-Sparks/pull/238#issuecomment-5759982309) contains actual stock-default and adjusted-stock measurements, including the remaining preemption, memory-guard and sequence-reference failures. The original failed attempt below is retained as historical evidence.

## Stock-default qualification: startup regression found — do not merge the revised candidate yet

This follows the [custom-configuration results](https://github.com/MiaAI-Lab/GLM-5.3-Flash-EXL3-2x-DGX-Sparks/pull/238#issuecomment-5756046631). Those measurements remain valid for that configuration, but **they do not establish stock-image compatibility**. I withdraw the earlier default-off merge recommendation pending the correction below.

### What was actually tested

- Revised local source head `0871b23bfcdde91571b6a8464e73b85a6eaa3fee`; runtime changes are from `6d5dd89`. The published PR head is still `2945711d8b7d563542cb3e48914b4fb4b12595cd`; no new source commits were pushed during this qualification.
- Unmodified repository `.env.example`, with machine-specific connectivity values only. Stock settings include `MAX_MODEL_LEN=850000`, `MAX_NUM_SEQS=4`, `MAX_NUM_BATCHED_TOKENS=7168`, `GPU_MEM_UTIL=0.85`, DFlash2 k=7, `ABLIT=0`, and no custom dense-FP8/KDA/fast-MoE opt-ins. The stock configuration was not reduced or otherwise tuned to make the test pass.
- Actual default image: `ghcr.io/miaai-lab/glm-5.3-flash-2x-dgx-sparks:exl3-instanttensor`, fetched for this test. Registry digest: `sha256:447114ee77d14c9b4732ee23978ada2a0ee9027868a231d6fd42700a8b25be1d`. Matching image ID on both ranks: `sha256:ef9f5013c41adf93a5171abb809283be0f202b88fd14098e4434337715614c62`.
- The normal launcher adopted the existing complete model cache. Failure occurs during overlay application, before model loading/inference; it is not a measured checkpoint-performance result.

### Actual default-OFF startup result

The image's GPU EXL3 self-check passed and both serving containers started. **Both then exited during `patch_hybrid_prefix_hit.py`, before health became ready**, with:

```text
kv_cache_coordinator.py: expected one dflash-boundary-verify target, found 0
```

This happened with `GLM53_DRAFT_KV_COMPACT=0`. Default-off does not avoid the problem because the overlay transformation itself runs at startup.

### Independent isolation against the actual stock-image source

I extracted a fresh coordinator from that exact image and ran the published and revised prefix overlays against separate copies, under both flag values:

| Prefix overlay | Flag OFF: source application | Flag ON: source application |
|---|---|---|
| Published `2945711` | Pass | Pass |
| Revised `6d5dd89` / `0871b23` | Fail: missing boundary-verification anchor | Same failure |

**This table is a source-application comparison, not four successful GPU boots.** Only the revised default-OFF serving boot was attempted; the ON incompatibility was independently reproduced before attempting inference. The revised overlay leaves the file unchanged on failure.

Actual stock coordinator SHA-256: `1f4f11011bff2a4aab1f0deaaa1c77878c354e939e19441064bd56933c50919d`.

### Root cause and required correction

The image already contains an older `# [glm53-hybrid-apc]` transformation. In `overlay/patch_hybrid_prefix_hit.py:433–438`, that marker causes the `MIN_OLD` → `MIN_NEW` conversion to be skipped. The older form has a direct `_glm53_is_draft_swa_spec(spec)` condition but lacks the newer `_glm53_draft_swa = ...` verification preamble. The new boundary transformation at lines 479–480 unconditionally requires that newer preamble, so its anchor count is zero.

The pristine-source fixture and our already-newer custom runtime did not cover this legacy, prepatched image shape. This is a **new compatibility regression in the revised candidate**: the published prefix overlay applies to the same stock source.

**Next-action owner: `plotarmordev`, as PR author.** Unblock condition: add a fail-closed migration for the recognized older hybrid-patched form, regression coverage using the actual stock-image shape, review the corrected current head, and rerun default OFF/ON qualification. Do not remove the anchor guard or silently substitute the custom image to obtain a passing “default” result.

### Performance and restoration outcome

**Stock-default performance: N/A.** No stock serving requests reached inference, so there is no defensible stock decode/prefill/TTFT percentage, and the planned OFF/ON/restored-OFF performance run was halted at startup. The custom-config percentages in the linked comment must not be relabeled as stock results.

The original custom deployment was restored. Both ranks' original images, environment maps, mount maps, and mounted allocator/prefix/quantization overlay hashes match the pre-test snapshots. All five frozen deployment source hashes are unchanged; health is HTTP 200. No deployment-local content or credentials are published here.

plotarmordev:
> **Publication update:** fixes are now pushed at `bdf6b9a`; the local/unpublished wording below is historical. [Current consolidated checkpoint](https://github.com/MiaAI-Lab/GLM-5.3-Flash-EXL3-2x-DGX-Sparks/pull/238#issuecomment-5760233232). Default OFF; draft; no merge.

## Installer fixed locally; stock qualification is mixed, not an all-green result

Tested runtime source: **`9a3aca4c0a5ad3a90b22ceb83dc71b3d0fa24f46`**, currently local/unpublished. The published PR head remains `2945711d8b7d563542cb3e48914b4fb4b12595cd`; this comment does not claim the fix is already in that head.

### What was wrong and what changed

The public InstantTensor image contains an older coordinator with the `glm53-hybrid-apc` marker but without the current v3 verification form. The installer treated the marker as proof that migration was complete, then failed to find the boundary-verification anchor. The existing custom deployment already had the newer coordinator form, which hid this compatibility defect.

The installer now recognizes and migrates that exact legacy form, including the form with the published replay stage. Before its sole write, it validates all owned stages and rejects unsupported drift, partial/duplicate stages, and competing ordinary bindings of owned helpers, leaving rejected input unchanged. Valid stock-derived output is byte-identical to a pristine installation. This is source-shape validation, not a sandbox for arbitrary Python.

**Verification:** 61 focused CPU tests passed without skips; standalone hybrid smoke passed. SIX independently cleared the exact source head after installer regressions, binding probes, source-shape/idempotence checks, and retention-overlay composition checks. That is source/CPU clearance, not a blanket GPU qualification.

### Actual stock defaults, without a cache-budget override

Public image ID: `sha256:ef9f5013c41adf93a5171abb809283be0f202b88fd14098e4434337715614c62`.

The public stock profile retained the 850,000-token limit, four sequences, 7,168-token batching, and GPU utilization 0.85. Cached model revisions were reused: target `024db9f7e9871e8efdf21538ba55af7442be3cd5`, drafter `dc77ff1c99eeb2df044ee3d4f0094eb033fee410`. No precision or serving-policy tuning was introduced. Container environments for these two arms differed only in the compact flag.

- **OFF:** the corrected installer completed, but startup rejected the cache capacity: **13.46 GiB required versus 10.73 GiB available**. No inference ran. This is distinct from the fixed installer failure.
- **ON:** startup succeeded with the automatic **10.93 GiB** cache budget. **64/65 planned requests completed**, with **31/31 marker/reference answers correct** and the 3,200-token rollover sequence check passing. The **814,571-token cold prompt answered correctly** (TTFT 634.510 s). Minimum sampled available RAM was **5.15 GiB**. Its repeat triggered the preemption guard (**2 observed preemptions**), so the repeat was not completed and the full workload is **not qualified**.

The compact stock-default arm therefore works beyond the custom setup, but this is not proof of clean near-limit prefix reuse.

### Controlled comparison with the explicitly approved 14 GiB budget

These are **adjusted-stock**, not unmodified-default results: all three arms added only `--kv-cache-memory-bytes 15032385536`. Fresh OFF / ON / OFF boots otherwise used the same stock image and environment. The 2 GiB available-memory floor, 256 MiB swap-out ceiling, and zero-new-preemption guard were unchanged.

| Outcome | OFF 1 (T0) | ON (T1) | OFF 2 (T2) |
|---|---:|---:|---:|
| Completed / planned requests | 64 / 65 | 63 / 65 | 48 / 65 |
| Correct marker/reference answers | 31 / 31 | 30 / 30 | 15 / 15 |
| Rollover sequence check | Passed | Passed | **Failed: repeated 202** |
| Observed preemptions | 4 | 0 | 0 |
| Minimum sampled available RAM (GiB) | 2.090 | **1.668** | 2.168 |
| Why execution stopped | Near-limit repeat preempted | Near-limit cold prompt crossed memory floor | Sequence-reference failure; later cases not reached |

The failed OFF sequence repeated 202 at integer position 203, before the 2,048-token window boundary. It is preserved as a failed check, not attributed to compact mode and not replaced by a retry-to-pass. ON completed the 63 non-near-limit requests; its 30 marker references and rollover check passed, but freeform code/prose outputs are not semantic-quality proofs.

The following measurements exist for matching prompts in all three arms. Percentages compare ON with the arithmetic mean of the two OFF measurements; lower is better for times, higher for throughput.

| Measurement | OFF 1 | ON | OFF 2 | ON vs mean OFF |
|---|---:|---:|---:|---:|
| Reservation IDs per configured maximum-length request | 532 | 295 | 532 | **−44.55%** |
| 100,701-token repeat TTFT (s) | 1.114 | 0.817 | 1.090 | **−25.80%** |
| 100,701-token cold TTFT (s) | 66.544 | 70.319 | 69.253 | +3.56% |
| Structured decode (tokens/s) | 63.76 | 65.45 | 63.60 | +2.78% |
| Code decode (tokens/s) | 47.79 | 44.17 | 42.02 | −1.64% |
| Prose decode (tokens/s) | 27.70 | 26.14 | 31.35 | −11.48% |
| C2 code aggregate (tokens/s) | 62.79 | 64.59 | 64.08 | +1.83% |
| C4 code aggregate (tokens/s) | 92.35 | 95.73 | 91.18 | +4.32% |
| Two long follow-ups, pair wall time (s) | 351.286 | 175.935 | 176.102 | −33.28%* |

Decode/concurrency values are medians of three trials. **48 completed prompt hashes matched across all three arms**, including the failed OFF rollover check; that does not turn the incomplete workload into a pass. The 100k repeat had 100,352 prefix-hit tokens in all three arms. Stock responses omit cached-token usage details, so prefix-hit observations use per-group Prometheus counter deltas, not invented API usage values.

*The second OFF long-follow-up result was already essentially equal to ON; the OFF baselines varied by about 50%. Do not interpret that mean-based percentage as a stable compact-mode speedup. Code/prose outputs also varied, and the OFF decode baselines drifted. These are descriptive measurements, not an identical-output kernel benchmark or a statistical significance claim.*

**Limits:** fewer reservation IDs do not mean smaller backing tensors or a measured VRAM reduction. The 14 GiB override crossed the memory guard with compact mode enabled; it is not a safe blanket fix. Near-limit reuse preempted in both the adjusted-stock OFF arm and the automatic-budget ON arm; its cause is not established by these measurements. TP3/TP4 GPU behavior, other backends, and tensor-level numerical parity remain unqualified. Default remains off. No push, merge, or broader stock-qualification claim is implied by this report.

Original service restored and checked after all runs: original image, environment, launch command, mounts and patch hashes on both ranks; five original source hashes unchanged; health and a live inference smoke passed.

Neutral result receipt SHA-256: `de9dc16deb0aafbbe60bc469d71cd250a988287c52e2f069f62537f6f4e79b46`. It binds 419 private raw-evidence files through manifest SHA-256 `9e901d5db485e1c42e3b2030e080be0b96de7c19e05821ec9b4f9e76fd7aaf0a`. Raw deployment details remain private. Documentation updated locally in `bdf6b9a667bbbdade6a6eadda1de54d4253ebd72`.

[Custom comparison](https://github.com/MiaAI-Lab/GLM-5.3-Flash-EXL3-2x-DGX-Sparks/pull/238#issuecomment-5756046631).

plotarmordev:
## Checkpoint: fixes pushed; experimental/default OFF; further investigation deferred

**Current PR head: `bdf6b9a667bbbdade6a6eadda1de54d4253ebd72`.** The eight local commits are now pushed to this PR. Runtime source tested and independently reviewed: `9a3aca4c0a5ad3a90b22ceb83dc71b3d0fa24f46`; the final commit updates README/CHANGELOG only. **The PR remains draft. No merge and no default-ON change.**

### 1. Fixes and verification

| Change | Why it matters | Evidence / status |
|---|---|---|
| DFlash boundary lookup for compact pages | A short tail must not unnecessarily lose a full MLA prefix page. Complete-window and replay checks remain. | CPU boundary coverage and live prefix-reuse cases passed. |
| DFlash-only allocator preflight on every grouping path | Prevents enabling this layout for unsupported groups/methods. | Focused allocator and launcher checks passed. |
| Legacy stock-coordinator migration | Stock had an old `glm53-hybrid-apc` marker but lacked the newer verification form. Custom already had that form, hiding the installer bug. | Actual stock installation now succeeds under both flag values. |
| Pre-write stage and helper-binding validation | Rejects unsupported drift, partial/duplicate stages and competing ordinary helper bindings instead of accepting a stale marker. | Rejected coordinator input stays unchanged. This is not a sandbox for arbitrary Python. |

**61 focused CPU tests passed without skips**, plus standalone hybrid smoke. SIX cleared the exact runtime source after independent source-shape, idempotence, binding and retention-composition checks. That is scoped source/CPU approval, not blanket stock/GPU approval.

### 2. Fresh custom comparison — correctness checks passed; performance varies

The existing custom image, precision and serving configuration were preserved. Only the compact flag changed between fresh OFF / ON / OFF boots. No stock 14 GiB override was carried into these runs.

**153/153 requests completed; 90/90 marker/reference answers were correct; all three 3,200-token rollover checks passed. No preemptions or safety stops. Minimum sampled available RAM: 4.73 GiB.** All 51 prompt hashes matched across the three arms.

Percentages below are ON versus the arithmetic mean of the two OFF results. Lower is better for times, higher for throughput. Decode/concurrency and short-prefill values are medians of three trials. The observations describe these measurements, not guaranteed improvements.

| Measurement | OFF 1 | ON | OFF 2 | Change | Observation |
|---|---:|---:|---:|---:|---|
| Reservation IDs per configured maximum-length request | 176 | 117 | 176 | −33.52% | Lower reservation demand; not a measured VRAM reduction. |
| Usable pool IDs after automatic profiling | 507 | 519 | 514 | — | Pool sizes varied; this was not a fixed-budget comparison. |
| Prefix-hit tokens on 100,701-token repeat | 100,352 | 100,352 | 100,352 | Equal | Prefix reuse preserved. |
| 100k repeat TTFT, seconds | 1.043 | 0.728 | 1.052 | −30.46% | Lower cached-prompt latency. |
| Cold 100k TTFT, seconds | 89.109 | 87.489 | 84.729 | +0.66% | Little difference in this small sample. |
| Two near-243k cold requests, pair wall seconds | 432.028 | 423.355 | 418.867 | −0.49% | Little difference in this sample. |
| Their follow-ups, pair wall seconds | 8.427 | 5.551 | 6.735 | −26.78% | Lower wall time; OFF baselines also varied by 20%. |
| Structured decode, tokens/s | 74.53 | 79.31 | 78.33 | +3.77% | Higher measured throughput; small difference. |
| Code decode, tokens/s | 51.23 | 50.45 | 49.64 | +0.03% | Little difference in this sample. |
| Prose decode, tokens/s | 38.81 | 30.72 | 32.19 | −13.46% | Lower throughput; output and baseline variation complicate attribution. |
| C2 code aggregate, tokens/s | 74.68 | 72.60 | 81.54 | −7.05% | Lower measured aggregate throughput. |
| 8k cold TTFT, seconds | 7.654 | 7.121 | 7.342 | −5.03% | Lower measured time to first token. |
| 32k cold TTFT, seconds | 28.182 | 28.011 | 27.428 | +0.74% | Little difference in this sample. |
| Rollover decode, tokens/s | 71.06 | 74.41 | 75.64 | +1.44% | Slightly higher throughput; all sequence checks passed. |

ON retained all 28,672 prefix tokens at tails 64, 65, 349, 896, 897 and 2,048. Both OFF arms retained only 25,088 at tail 64. Changed-suffix references and the original-prompt reference afterward were correct. Long cold prefills were effectively serialized; this is not proof of parallel long-prefill capacity.

### 3. Stock outcomes — NOT a full qualification pass

Stock public serving settings: 850k context, four sequences, 7,168-token batching, GPU utilization 0.85. Public InstantTensor image ID: `sha256:ef9f5013c41adf93a5171abb809283be0f202b88fd14098e4434337715614c62`. Cached target revision `024db9f7e9871e8efdf21538ba55af7442be3cd5` and drafter revision `dc77ff1c99eeb2df044ee3d4f0094eb033fee410` were reused; this was not a fresh model download.

| Configuration / arm | Completed requests | Correct marker references | Rollover | Minimum sampled available RAM | Outcome |
|---|---:|---:|---|---:|---|
| Stock automatic budget, OFF (S0) | 0 inference requests | — | Not run | — | Startup capacity rejection: installer passed, but startup needed 13.46 GiB cache versus 10.73 GiB available. |
| Stock automatic budget, ON (U1) | 64/65 | 31/31 | Passed | 5.15 GiB | Partial completion: booted with 10.93 GiB; 814,571-token cold prompt answered correctly, but repeat triggered preemption (2 observed). |
| Adjusted stock, 14 GiB, OFF 1 (T0) | 64/65 | 31/31 | Passed | 2.090 GiB | Stopped: near-limit repeat preempted (4 observed). |
| Adjusted stock, 14 GiB, ON (T1) | 63/65 | 30/30 | Passed | **1.668 GiB** | Stopped: near-limit cold prompt crossed the 2 GiB memory floor. No preemptions observed before stop. |
| Adjusted stock, 14 GiB, OFF 2 (T2) | 48/65 | 15/15 | **Failed** | 2.168 GiB | Reference check failed: sequence repeated 202 at integer position 203. Later cases were not reached. |

The 14 GiB arms are **adjusted-stock**, not untouched defaults. The approved override was `--kv-cache-memory-bytes 15032385536`. The memory floor, swap-out ceiling and preemption guard were not weakened. Failed runs were retained, not replaced with retries-to-pass. The OFF counting failure occurred before the sliding-window boundary; it is not evidence of a compact-specific failure.

### 4. Adjusted-stock performance — useful measurements, incomplete workload

These measurements exist for matching prompts in all three 14 GiB arms. All three had 552 usable pool IDs. There were 48 matching completed prompt hashes, including the failed OFF rollover check. Untouched stock OFF could not start, so there is **no untouched-default OFF/ON inference speed comparison**.

| Measurement | OFF 1 | ON | OFF 2 | ON vs mean OFF | Observation |
|---|---:|---:|---:|---:|---|
| Reservation IDs per 850k request | 532 | 295 | 532 | −44.55% | Lower logical reservation demand. |
| 100k repeat TTFT, seconds | 1.114 | 0.817 | 1.090 | −25.80% | Lower measured time to first token. |
| Cold 100k TTFT, seconds | 66.544 | 70.319 | 69.253 | +3.56% | Higher measured time to first token. |
| Structured decode, tokens/s | 63.76 | 65.45 | 63.60 | +2.78% | Higher measured throughput; small difference. |
| Code decode, tokens/s | 47.79 | 44.17 | 42.02 | −1.64% | Slightly lower measured throughput. |
| Prose decode, tokens/s | 27.70 | 26.14 | 31.35 | −11.48% | Lower measured throughput. |
| C2 code aggregate, tokens/s | 62.79 | 64.59 | 64.08 | +1.83% | Higher measured throughput; small difference. |
| C4 code aggregate, tokens/s | 92.35 | 95.73 | 91.18 | +4.32% | Higher measured throughput; small difference. |
| Long follow-ups, pair wall seconds | 351.286 | 175.935 | 176.102 | −33.28% | Not attributable to the flag: second OFF was already as fast as ON. |

The 100k repeat had 100,352 prefix-hit tokens in all arms. Stock API usage omitted cached-token details; these observations use per-group Prometheus counter deltas. Concurrent deltas are group aggregates, not per-request cache counts.

The automatic-budget ON cold 814,571-token request had TTFT **634.510 s**. The adjusted-stock OFF cold request had TTFT **700.778 s**. **Do not call their difference a flag-only speedup:** their cache budgets differed, and neither completed the subsequent repeat.

### 5. Earlier valid custom experiment — retained, not mixed into the fresh averages

The earlier `6d5dd89` A3/C/A4 run completed 153 requests, 90 correct references and three rollover checks without preemptions or safety stops. It remains separate and byte-frozen.

| Earlier measurement | OFF 1 | ON | OFF 2 | ON vs mean OFF |
|---|---:|---:|---:|---:|
| Reservation IDs | 176 | 117 | 176 | −33.52% |
| 100k repeat TTFT, seconds | 1.040 | 0.706 | 1.033 | −31.94% |
| Cold 100k TTFT, seconds | 83.458 | 86.880 | 85.186 | +3.03% |
| Long follow-up pair, seconds | 7.857 | 5.618 | 7.580 | −27.21% |
| Structured decode, tokens/s | 80.09 | 80.08 | 77.88 | +1.39% |
| Code decode, tokens/s | 53.83 | 52.62 | 52.91 | −1.42% |
| Prose decode, tokens/s | 32.70 | 30.83 | 31.38 | −3.78% |
| C2 code aggregate, tokens/s | 80.20 | 76.56 | 74.60 | −1.08% |
| Rollover decode, tokens/s | 72.48 | 75.89 | 73.81 | +3.75% |

The repeat-TTFT benefit reproduced in the fresh run. The generation regressions varied in size. Earlier installer failures and the pre-boundary-fix prefix-reuse defect are historical failures, not successful qualification of this head.

### 6. What we know versus theories for next time

| Topic | Verified fact | Theory / next diagnostic — NOT established |
|---|---|---|
| Why custom originally worked but stock did not | Their installed coordinator source forms differed. The legacy marker caused migration to be skipped. | **Resolved source cause.** Not evidence that custom tuning was inherently required by the layout. |
| Stock OFF startup failure | Required cache exceeded the automatically available budget. | Capacity/headroom policy needs consideration; do not assume a fixed 14 GiB override is safe. |
| Near-limit reuse preemption | Occurred in adjusted-stock OFF and automatic-budget ON. | **[INFERENCE]** Cache-pressure, prefix-hit admission/accounting, or scheduler behavior may be involved. Instrument hit lengths, free/held blocks and allocation decisions before assigning a cause. |
| 14 GiB ON memory stop | Available RAM crossed the guard; this was not an observed OOM claim. | **[INFERENCE]** The fixed cache budget may leave insufficient activation/runtime headroom. Compare phase-local memory/profile allocations and baseline variation; no allocation responsible has been identified. |
| Prose/C2 regressions | Some observed rates were lower with ON. Freeform outputs differed and OFF baselines drifted. | **[INFERENCE]** Output-dependent speculation/workload differences, run variation, or layout/backend overhead may contribute. Use identical-output/controlled checks to distinguish them. |
| OFF sequence failure | The response repeated 202 before crossing the window boundary. | Its cause is undetermined. Do not label it a compact-mode bug, and do not erase it by weakening the check. |

### 7. Decision and deferred work

The custom long-prompt reuse benefit reproduced, and the stock installer incompatibility is fixed. Full stock near-limit qualification, the fixed-budget memory breach and possible generation regressions remain unresolved. Fewer reservation IDs are **not** a measured VRAM reduction or a guarantee of safe higher context/concurrency.

**Decision: preserve the fixes in this draft PR, keep default OFF, and resume investigation later.** No default-ON or merge recommendation yet. The current evidence supports a monitored custom opt-in, not a universal speedup claim.

**Next-action owner: plotarmordev / the resumed implementation session.** Start from the saved failures and instrument admission/memory behavior; do not request or run another broad benchmark merely to get a green result. Before further GPU runs, select the configuration and diagnostic question. Unblocking evidence for a later merge recommendation: explain the stock stops or explicitly establish a bounded supported scope, resolve or characterize the observed regressions, and re-review the then-current head. No unrelated maintainer policy decision is being treated as missing author evidence.

TP3/TP4 GPU behavior, other backends/architectures and tensor-level numerical parity remain unqualified. No new experiment was run for this publication checkpoint.

### Evidence and restoration

- Fresh experiments: **392 completed benchmark requests and 197 correct marker/reference answers**, with the failed rollover and safety-stopped arms explicitly retained. Those totals are not an all-pass claim.
- Neutral result receipt SHA-256: `de9dc16deb0aafbbe60bc469d71cd250a988287c52e2f069f62537f6f4e79b46`.
- 419-file raw-evidence manifest SHA-256: `9e901d5db485e1c42e3b2030e080be0b96de7c19e05821ec9b4f9e76fd7aaf0a`. Raw deployment details stay private.
- Earlier valid custom receipt: `35fd5ef14b9e9502116311c5706e0ae874056369f90e38ba2184f2be3257e7e2`. Its 96 manifested files and the 15-file original stock-failure manifest were verified unchanged.
- Original deployment restored: original images, environments, commands, mounts and patch hashes verified; five original source hashes unchanged; health and live inference smoke passed.
- Outgoing eight-commit range: no secret findings. Worktree scan: 32 previously reviewed false positives, all in four files byte-identical to the old published head; not reported as a clean raw scan.

Detailed reports: [fresh custom](https://github.com/MiaAI-Lab/GLM-5.3-Flash-EXL3-2x-DGX-Sparks/pull/238#issuecomment-5756046631), [stock and adjusted stock](https://github.com/MiaAI-Lab/GLM-5.3-Flash-EXL3-2x-DGX-Sparks/pull/238#issuecomment-5759982309), [original installer failure](https://github.com/MiaAI-Lab/GLM-5.3-Flash-EXL3-2x-DGX-Sparks/pull/238#issuecomment-5756351431). Their local/unpublished wording describes their earlier reporting time; the fixes are now pushed at the head above.

plotarmordev:
## Follow-up pushed: Mamba state fixes + qualification results

Pushed at `fd329d2` (fast-forward from `bdf6b9a`). Draft; default remains off.

This push adds two Mamba "align" overlays that apply with **either** compact value:

- `patch_mamba_align_state_free.py` — releases every superseded Mamba state block. Previously one tracker per request could be overwritten during async prefill, leaving blocks referenced for the request's lifetime. Align-mode admission now reserves 10 pages (was 9), matching the resident peak.
- `patch_mamba_align_chunking.py` — aligns prefill chunks to the Mamba block rather than the drafter's smaller block, so valid checkpoints are no longer missed (the stock diagnostic lost 3,584 of 28,672 prefix tokens).
- Compact drafter managers no longer retain the unused EAGLE lookahead block; the KV-capacity log accounts for this (`lookup=boundary`).

Both the OFF and ON arms below include the Mamba fixes, so the comparisons isolate the **compact-page tradeoff**, not the total effect of every fix.

### Custom TP2/DFlash2 configuration — compact ON vs mean of OFF arms

Fresh OFF / ON / OFF boots, 81 requests each; decode cells use nine trials per arm. Faster throughput is positive; lower time is negative.

| Measurement | OFF 1 | ON | OFF 2 | ON vs mean OFF |
|---|---:|---:|---:|---:|
| Reservation IDs per max-length request | 180 | 121 | 180 | **−32.78%** |
| 28,672-token prefix + 64-token tail, repeat TTFT (s) | 2.623 | 0.302 | 3.434 | **−90.02%** |
| C2 code aggregate (tok/s) | 73.521 | 75.957 | 74.230 | +2.82% |
| Code decode (tok/s) | 53.493 | 52.048 | 49.361 | +1.21% |
| Structured decode (tok/s) | 78.589 | 78.380 | 76.390 | +1.15% |
| Two 242,628-token follow-ups, pair wall (s) | 5.486 | 5.718 | 6.357 | −3.43% |
| 100,701-token repeat TTFT (s) | 0.710 | 0.706 | 0.726 | −1.70% |
| Cold 8k TTFT (s) | 7.042 | 7.107 | 7.125 | +0.34% (slower) |
| Cold 100,701-token TTFT (s) | 85.576 | 87.085 | 86.902 | +0.98% (slower) |
| Cold 32k TTFT (s) | 27.731 | 28.642 | 28.258 | +2.31% (slower) |
| Two cold 242,628-token requests, pair wall (s) | 419.578 | 433.644 | 420.289 | +3.26% (slower) |
| Prose decode (tok/s) | 33.188 | 31.131 | 31.200 | −3.30% (slower) |

The prior first-reuse regression is resolved: ON retained the full 28,672-token prefix at all six tested tails (64, 65, 349, 896, 897, 2,048); both OFF arms retained 25,088.

The two ~3.3% regressions were rechecked with a **fixed, prespecified holdout** (three fresh OFF / ON / OFF boots, 45 prose trials, three cold long-C2 pairs per arm; all trials kept):

| Holdout median | OFF 1 | ON | OFF 2 | ON vs mean OFF |
|---|---:|---:|---:|---:|
| Prose decode (tok/s) | 32.684 | 32.286 | 32.495 | −0.93% |
| Cold long-C2 pair wall (s) | 420.493 | 429.522 | 428.543 | +1.18% (slower) |

### Stock/default configuration — 850k context, 4 sequences, 7,168 batching, utilization 0.85

Public InstantTensor image, no fixed-memory override:

| Result | Compact OFF | Compact ON |
|---|---|---|
| Startup | **Fails**: 13.56 GiB KV required vs 10.77 GiB available | **Boots** |
| Full workload | Not runnable | **95/95 requests passed**, zero preemptions |
| 814,571-token cold prompt TTFT | — | 633.711 s |
| Same prompt, repeat TTFT | — | 2.748 s (**−99.57% vs cold**) |
| Repeat prefix-cache hits | — | 813,568 tokens (99.88% of prompt) |
| Minimum sampled available RAM | — | 5.55 GiB (guard: ≥ 2 GiB) |

At the previous head (`9a3aca4`), stock ON answered the cold prompt but recorded two preemptions and could not complete the repeat. With these fixes, cold and repeat both complete without preemption.

Cached-conversation residency is still not guaranteed: two 243k follow-ups took 9.31 s and 170.08 s, with 240,128 of 485,306 queried tokens hit.

### Totals and verification

- Full workloads + holdout: **500 requests, 140 correct reference answers, four 3,200-token rollover checks**, zero preemptions, no safety stops (guards: ≥ 2 GiB available RAM, ≤ 256 MiB new swap per node).
- CPU: focused compact/Mamba/capacity/prefix suites re-run at this head against sources extracted from **both** supported images; launcher parity/numeric/override suites and `bash -n` pass.
- Commit-range scan for secrets and deployment-local content: no findings.

### Limits

- **No universal decode speedup.** Prose decode is about 1% slower in the holdout, and cold long prefills are about 1–3% slower.
- Reservation-ID savings are **not a measured VRAM reduction**.
- Stock OFF cannot fit the unchanged defaults, so no matched stock OFF-vs-ON inference comparison exists.
- TP3/TP4 GPU execution, other backends/architectures, a full image rebuild and full-model bitwise equivalence remain unqualified. **Default remains `0`** in this PR.

Neutralized receipt SHA-256 (raw evidence private): `abee1ec2b2783620a5f42cd397d29c92ce33add653f8dc106ea0103420e4b671`.

plotarmordev:
## Merge qualification at `11e6e29` (PR238 merged with `main` `c1b7d4c`, including #242)

`main` merged cleanly. The only follow-up is a documentation fix: #242's 14 GiB InstantTensor pool figures now include the Mamba align reservation (**536 IDs = 13.56 GiB per 850k request; 876,958 tokens = 1.03×**), matching the merged stock boot's capacity log.

### CPU (exact head)

| Suite | Result |
|---|---|
| Compact / Mamba / capacity-log / prefix-hit / retention, against sources from **both** supported images | 43 passed, 1 skipped per image (skip: coordinator-source retention case, unchanged by this PR) |
| Launcher parity, numeric config, overrides, #242 KV note | 33 passed; `bash -n` clean |

### GPU (TP2, 2× GB10; each trial restored the production deployment and re-verified it)

Runtime vLLM sources on both ranks matched an offline composition of this tree onto each image. Container environments matched exact per-rank tables, and both Mamba overlay markers were present.

| Trial | Configuration | Result |
|---|---|---|
| Stock `main` defaults | compact **off**, 850k, 4 seqs, 7,168 batching, 0.85, #242 fixed 14 GiB pool, public InstantTensor image | Boots. **47/48** checks passed, including the boundary matrix (1–2,049-token tails), cache controls and reference answers; no preemptions; 0 / 0.004 MiB new swap |
| Stock `main` defaults, stress | same | Quick set 7/7 passed. The 816k-token cold prefill was stopped by our host guard (≤ 256 MiB new swap): BOTTOM reached 295 MiB with ~5 GiB still available |
| Custom | compact **on**, 262k, 2 seqs, 1,024 batching, 0.865, InstantTensor | **7/7 passed**; warm and extended-prefix reuse 7,168 tokens; fresh-salt controls 0; 0 swap |

### Findings that are not caused by this PR (FAB review, code evidence)

- **One rollover flake:** the 3,200-token count repeated "202" once. The same signature is already recorded for stock compact-off plus the fixed 14 GiB pool at `9a3aca4`, before the Mamba overlays existed. For a 32-token prompt, the overlays perform exactly `main`'s operations: no chunk split, no state release, no prefix hit.
- **Near-limit host memory:** PR238 adds no host allocation. The +4 reserved IDs are inside the fixed pool, and the chunking change halves per-step prefill work. The same Mamba overlays ran the 816k cold prefill twice with 0 swap under the automatic pool. The pressure comes from #242's fixed 14 GiB pool on the node with less headroom, which `main` already documents (the 15 GiB option measured 0.8–2.2 GiB free). An 816k compact-off request under that pool is **not qualified** by this run.

### Review

FAB (independent @FAB role): **APPROVE**. This covers the PR238 code at the merged head (default-off guarantee, Mamba lifetime and chunking, 14 GiB-pool arithmetic), the qualification harness, and both attributions above.

Hosted CI is not configured for this repository, so there are no hosted checks. The exact-head local suites above are the CI equivalent.

**Default remains `GLM53_DRAFT_KV_COMPACT=0`.** The Mamba fixes apply with either value.

hernaninverso:
Serving observations for `GLM53_DRAFT_KV_COMPACT=1` under a multi-session agentic load (2× GB10, TP=2 over CX7), 2026-09-23.

**Setup (declared from our launcher and `.env`; not all of it is visible in the boot or metrics logs).** Local build of `main` from 2026-09-23 with #238. `MAX_MODEL_LEN=200000` and `GPU_MEM_UTIL=0.85` are confirmed by the logs. Also `MAX_NUM_SEQS=4`, `MAX_NUM_BATCHED_TOKENS=7168`, DFlash2 k=7 with `GLM53_ADAPTIVE_K=ema`, `GLM53_MIXED_PREFILL_CHUNK=fair`, cooperative MoE overlay, `ABLIT=1`. Between arms we changed only `GLM53_DRAFT_KV_COMPACT` 0 → 1, with the same launcher and the same `.env`. Clients were 1–3 concurrent requests from Claude Code agent sessions. Over three days of this workload the client-side prompts had a median of ~84k tokens (p90 ~149k); vLLM's cumulative means over the two windows below are ~60k–63k prompt tokens per request.

**Boot capacity log (`GLM53_KV_CAPACITY_LOG=1`)**
| | compact=0 | compact=1 |
|---|---:|---:|
| ids per 3584-token cached segment | 38 `[1,0,1,1,1,1,33]` | 8 `[1,0,1,1,1,1,3]` |
| cached-conversation capacity (upper bound: nothing running, every reachable block hashed, aligned hits) | ≈43,008 tok (12 seg) | ≈204,288 tok (57 seg) |
| ids per 200k request | 354 (1.39×) | 117 (3.97×) |

**Serving: `/metrics` scraped every ~5 min.** Samples cover 22:08–22:39 UTC for compact=0 and 22:49–23:29 UTC for compact=1.
| | compact=0 | compact=1 |
|---|---|---|
| prefix hit rate per 5-min window (Δhits/Δqueries) | median 62.7 %, range 0–97.5 % (6 windows) | median 96.9 %, range 90.5–97.8 % (8 windows) |
| summed per-request prefill / decode / queue time ÷ summed e2e latency | 45.1 / 37.0 / 15.2 % (125 req) and 50.0 / 34.7 / 13.3 % (183 req) | 19.9 / 78.2 / 1.1 % (112 req) |
| spec acceptance (accepted / draft tokens) | 54.6 % and 53.7 % | 60.9 % |

These time shares are sums of per-request times, which overlap across concurrent requests. They are not an engine-utilization breakdown. An earlier manual reading at 21:50 UTC in the compact=0 arm (61 requests, before our probes started) gave 54 / 22 / 21 %; it is not part of the 5-min scrape series.

**Eviction probe.** A unique 27,128-token prompt was re-sent after growing idle gaps while the sessions kept working. One probe ran at a time, and a probe counted only if Δ`prefix_cache_queries_total` equalled its own prompt tokens. Every re-send also refreshes the prefix.
- compact=0: after ~1.35 min idle, 0 hits. The request took 141 s end to end under `fair`.
- compact=1: 25,088 hits after ~1, 2, 4 and 8 min idle (1.9–2.4 s each), and 0 hits after ~16 min.

**Caveats.**
- This is observational, not an A/B/A. The two windows ran different jobs; in the second, a long translation task raised the decode share.
- Our own cache probes ran from ~22:04 to ~22:40 UTC: re-sends and cold 27k prompts, about 40 requests. They overlap almost all of the scraped compact=0 samples. The cleanest compact=0 figure is therefore the manual 21:50 UTC reading, taken before the probes (54 / 22 / 21 % and a cumulative hit rate of 69.7 %), but it is not part of the 5-min scrape series.
- Each arm has few windows and a single probe series.
- One compact=0 window had Δhits=0 over 98,496 queried tokens. That does not show the global freeze of #106, but it does not rule one out either. The compact=1 window is shorter than the ~1 h that #106 describes.

PR #242, opened 2026-09-21
InstantTensor: ship the 14 GiB KV reservation by default, document the cost, warn at preflight (#204)

### LINKED ISSUE #204: InstantTensor loader costs ~4.4 GiB of KV cache — stock GPU_MEM_UTIL=0.85 no longer boots at 850k (needs ~0.88); cost is undocumented
## Summary

After switching to the new default image `:exl3-instanttensor` (main @ `8f29c6d`), the shipped default `GPU_MEM_UTIL=0.85` can no longer boot at `MAX_MODEL_LEN=850000` on a 2× DGX Spark kit. The KV pool falls to ~10.1–10.6 GiB, below the ~13 GiB hard floor for one max-length request, and vLLM refuses to start.

A controlled experiment shows the cause precisely: **disabling the InstantTensor loader alone recovers 4.43 GiB of KV cache**.

## Environment

- 2× DGX Spark, GB10, 121.69 GiB unified memory per node
- TP=2 (`mp`, `--nnodes 2`), `--kv-cache-dtype fp8`
- Repo: `main` @ `8f29c6d`
- Image: `ghcr.io/miaai-lab/glm-5.3-flash-2x-dgx-sparks:exl3-instanttensor`
- `.env`: `MAX_MODEL_LEN=850000`, `MAX_NUM_SEQS=4`, `MAX_NUM_BATCHED_TOKENS=7168`, `KV_CACHE_DTYPE=fp8`, `SPEC_METHOD=dflash`

## Controlled experiment (single variable: the loader)

Same image, same launcher/overlays (`8f29c6d`), same `GPU_MEM_UTIL=0.85`, on the head node, no other GPU consumers:

| `LOAD_FORMAT` | Available KV | KV tokens | Boot |
|---|---|---|---|
| `instanttensor` (new default) | **10.58 GiB** | – | ❌ `ValueError` below |
| **empty** (`LOAD_FORMAT=`, vLLM auto) | **15.01 GiB** | **934,680** | ✅ |

**Delta: 4.43 GiB.**

For reference, the previous wheel-less image `:exl3` running the *new* code at 0.85 gives 15.54 GiB / 925,093 tokens — i.e. **turning the loader off restores (and slightly exceeds) the old headroom**.

CUDA environment is identical between `:exl3` and `:exl3-instanttensor` (`CUDA_VERSION=13.0.1`, `NV_CUDA_CUDART_VERSION=13.0.88-1`), so this is **not** a base-image/CUDA difference — it is the InstantTensor load path.

## The trade-off (a good feature, but invisible to users)

| `LOAD_FORMAT` | Model loading time |
|---|---|
| `instanttensor` | **37.4 s** |
| empty (vLLM auto) | 305.7 s |
| old `:exl3` image | 301.8 s |

So InstantTensor makes weight loading **~8.2× faster at a cost of ~4.4 GiB of resident memory**. That is a reasonable trade — but it is currently undocumented.

## Why this bites

The README states the shipped default is `850k / util 0.85 / rightsize` and that it "boots with ~0.5 GiB of KV margin". With InstantTensor enabled, 0.85 yields only ~10.5 GiB, i.e. ~3 GiB *below* the 13.46 GiB needed for one 850k request — so the documented default no longer boots.

The "Existing installs: pull the InstantTensor image" section instructs changing only `IMAGE` and `LOAD_FORMAT`, and does not mention that `GPU_MEM_UTIL` must be raised as well.

We had to raise `GPU_MEM_UTIL` to **0.88** to boot (KV = 14.0 GiB, 862k–884k tokens). That works, but host headroom drops to ~2 GiB instead of the ~5 GiB the README targets.

## Suggestions

1. **Document the memory cost** of `--load-format instanttensor` (~4.4 GiB on this kit) alongside its ~8× loading speedup, so users can choose deliberately (`LOAD_FORMAT=` remains a valid opt-out).
2. **Update the "Existing installs" section** to state that `GPU_MEM_UTIL` likely needs raising (e.g. 0.85 → 0.88) when moving to the InstantTensor image at 850k.
3. Consider bumping the shipped default `GPU_MEM_UTIL` for the InstantTensor image, or adding a preflight warning when the projected KV pool cannot hold one max-length request.

## Logs (excerpt)

```
# LOAD_FORMAT=instanttensor, GPU_MEM_UTIL=0.85
ValueError: To serve at least one request with the model's max seq len (850000),
(13.46 GiB KV cache is needed, which is larger than the available KV cache memory (10.58 GiB))

# LOAD_FORMAT= (vLLM auto), GPU_MEM_UTIL=0.85
GPU KV cache size: 934,680 tokens, Maximum concurrency for 850,000 tokens per request: 1.10x

# LOAD_FORMAT=instanttensor, GPU_MEM_UTIL=0.88 (current workaround)
GPU KV cache size: 862,781 tokens
Model loading took 82.05 GiB and 35.98 seconds
```

Happy to provide full startup logs for all three runs if useful. Thanks for the great work on this repo.

hecisaza:
Closes #204.

A fresh install from `.env.example` does not boot. The InstantTensor loader (default on
`:exl3-instanttensor`) leaves ~4.4–5.6 GiB less for the KV pool than vLLM auto on a 2x GB10
kit; at the shipped `MAX_MODEL_LEN=850000` / `GPU_MEM_UTIL=0.85` that is ~12.4 GiB against the
13.46 GiB one request needs, and the engine refuses at KV allocation 5–9 minutes in. The
"Existing installs" section tells users to change only `IMAGE` and `LOAD_FORMAT`, which is the
trap. Measured three times on this commit by three operators (details in #204).

**Changes**

- `.env.example`: `EXTRA_ARGS="--kv-cache-memory-bytes 15032385536"` is now **on by default** (TP=2 only: `.env.tp3.example` / `.env.tp4.example` clear it, since 14 GiB does not hold a 1M-token request and those topologies were not measured).
  It was already there as a comment. 14 GiB reserves 883,552 tokens (1.04x of 850k), keeps
  ~5 GiB host headroom, and booted 3/3 in the A/B runs and once more from a pristine copy of
  this exact template. The comment block says why, and that it must move with `MAX_MODEL_LEN`.
- `README.md`: line 81's "~0.5 GiB of KV margin" now states which loader it applies to; the
  "Existing installs" step gains the third line and a paragraph on why, including that 0.88
  booted 1 of 4 attempts here; a new **InstantTensor and KV memory** section gives the measured
  cost and ranks the three ways to run it.
- `start.sh`: `preflight_instanttensor_kv_note()`, called from `preflight()` after the two
  `preflight_memory` checks. One `warn` line when `LOAD_FORMAT=instanttensor`,
  `MAX_MODEL_LEN>=850000`, `GPU_MEM_UTIL<=0.85` and `EXTRA_ARGS` has no
  `--kv-cache-memory-bytes` — i.e. an existing `.env` that predates this default. Every other
  path is silent and returns 0. It lives inside the memory-guard marker block so harnesses
  that slice `preflight()` by those markers keep working.
- `tests/test_instanttensor_kv_note.py`: fires for the failing combination; silent for
  loader-off, shorter context, higher share, either flag spelling, empty/malformed inputs;
  placement and single call site asserted. Fails on the unpatched tree.

**Effect on existing installs:** none unless they copy the new template over their `.env`. The
launcher reads the user's `.env`, not the example. An existing `.env` with the loader on and no
reservation now gets the NOTE at preflight instead of a failed boot nine minutes later.

**Not changed:** `GPU_MEM_UTIL`. 0.88 × 121.69 = 107.09 GiB sits on the worker's CUDA-free at
check time (105.8–107.1 measured), and it would undo the 2026-09-07 decision to lower the share.

**Verification.** Suite on `775a58b`: `1 failed, 276 passed` → `1 failed, 280 passed` (the four
new tests; nothing else moved). `shellcheck -S warning`: identical to `main`. Preflight run in
place on the 2x GB10 pair: NOTE prints for the failing combination, silent with the reservation
and with the loader off. A boot from a pristine copy of the new `.env.example` (only the
kit-specific IPs/NICs/model filled in) reached `/v1/models` 200 with 883,552 tokens.


plotarmordev (changes requested):
Reviewed current head `4cc45992aefc4faa50e67da0996cc8c9577d613a` against `775a58b`.

**Changes requested — P2: preserve caller-supplied EXTRA_ARGS in the TP3/TP4 configuration path.**

The new filtering at `start-tp3.sh:119-135` and `start-tp4.sh:93-109` runs after sourcing `.env` without capturing/restoring the caller's `EXTRA_ARGS`. This has two observable regressions:

1. With the new template's `EXTRA_ARGS="--kv-cache-memory-bytes 15032385536"`, launching with caller `EXTRA_ARGS="--kv-cache-memory-bytes 21474836480 --max-log-len 7"` results in an empty aggregate: the unrelated `--max-log-len 7` is lost too.
2. With an existing `.env` that does not set EXTRA_ARGS, the same caller value becomes only `--max-log-len 7`: the operator's explicit 20-GiB reservation is silently removed even though it did not come from the TP2 template.

I exercised the actual pre-configuration launcher preambles in temporary directories with synthetic dotenv files; no SSH/docker/GPU actions ran. Both TP3 and TP4 preserve the full caller value on the base and exhibit the losses above on this head. TP2 correctly preserves the caller in both versions.

Please capture caller setness/value before sourcing configuration, remove the inherited TP2 reservation only from the file-derived configuration, then restore explicit caller input (including an explicit empty value) after the topology overlay. Add behavioral coverage through that whole precedence path; tests of the extracted filtering loop alone miss this. Preserve topology-file overrides as well.

The earlier inherited-14-GiB cap, token-boundary detection, and loader-off documentation concerns are addressed at this head; I am not requesting them again. The remaining migration NOTE/example should also say to append the reservation while retaining existing EXTRA_ARGS, rather than suggest replacing the aggregate.

**Validation:** 42 passed across `test_instanttensor_kv_note.py`, `test_start_overrides.py`, `test_launcher_rank_parity.py`, `test_numeric_config.py`, and `test_nccl_multi_hca.py`; `bash -n` passed for all three launchers. `shellcheck -S warning` reports the same two existing SC2034/NFS_RANKS warnings as the base, not new diagnostics. Full-suite and new GPU qualification were not run; the configuration regression is already a merge blocker, and another boot would not resolve it. Existing operator GPU reports in #204 were read, not represented as my own measurements.

**Hold:** concrete caller-precedence regression above. **Next-action owner:** @hecisaza (PR author). **Unblock:** preserve caller and topology precedence, cover it behaviorally, and review the revised head. No rebase or new GPU measurements are requested before the source correction.

plotarmordev (approved):
Re-reviewed current head 543a44e0490662224846e5efd6054610866f482b. The prior caller EXTRA_ARGS precedence request is addressed, and MIAGO implemented the remaining nonblocking follow-ups: TP4 template guidance, corrected PR description and existing-install documentation, explicit shared-cap removal diagnostics without argument-value disclosure, behavioral tests replacing source-text assertions, and command-local locale hardening. Coordinator independently reviewed the final diff and removed one wording-pinning assertion.

Verification: 310 passed, 4 skipped, 70 subtests passed in the isolated CPU environment; exact base 301 passed with the same four pristine-vLLM-source skips. The diagnostic regression tests produce two expected failures on pre-polish e336284 and pass here. bash -n passes all launchers; only the two unchanged SC2034/NFS_RANKS ShellCheck warnings remain. Commit-range secret scan is clean; worktree findings were reviewed false positives. No required CI checks or test workflow are configured. No GPU or multi-node requalification performed; existing hardware results remain author-reported. No unresolved code defects found; approve.

plotarmordev:
Merged normally as `c1b7d4c9f98c16af65640fef05fa493bbde52ccf` after approving head `543a44e0490662224846e5efd6054610866f482b`. The merge tree exactly matches the validated tree (`e1ac7aefb759a77cb3084b32b04458073425bf8f`).

Validation: 310 passed, 4 unchanged pristine-vLLM-source skips, 70 subtests passed; bash syntax passed; no new ShellCheck warnings; commit-range secret scan clean. No GPU/multi-node requalification was performed.

Maintainer follow-up (owner: plotarmordev / repository maintainers): run test CI against this merge commit if/when test CI becomes available. There is currently no hosted test workflow or required test check configured; this is not a claim of green hosted CI.

Issue #205, opened 2026-09-16
Cold boot after start.sh's own rsync: InstantTensor buffer (1,210 MiB) exceeds cuda-free budget because GB10 counts page cache as used (preflight blind spot); CG_ESTIMATE=0 receipt for #204


guangwenz:
## Summary

On a cold first boot (fresh `./start.sh` that rsyncs the 164 GiB to the worker), the **head rank dies 8 s into weight load** with:

```
RuntimeError: buffer_size (1268776960 B) exceeds device memory budget (793399296 B)
```

from `instanttensor/_impl.py:_finalize_buffer_size`, before a single tensor is read. Both ranks fail identically. The cause is the **page cache from start.sh's own weight rsync**, invisible to the kit's preflight. A secondary receipt for #204: `CG_ESTIMATE=0` recovers the KV shortfall at `GPU_MEM_UTIL=0.87` without going to 0.88.

## Environment

- 2× DGX Spark (GB10, 121.63 GiB unified), head `gx10-59ad`, worker over private CX7 pair
- Repo `main` @ `8f29c6d`, image `ghcr.io/miaai-lab/glm-5.3-flash-2x-DGX-Sparks:exl3-instanttensor` (built locally, stamp `4ccaea67e161`)
- Defaults from `.env`: `MAX_MODEL_LEN=850000`, `GPU_MEM_UTIL=0.85`, `SPEC_METHOD=dflash`, `KV_CACHE_DTYPE=fp8`, `LOAD_FORMAT=instanttensor`
- Fresh sync this boot: rsync moved 175,715,896,392 B to the worker minutes before containers started

## Root cause (verified with measurements, not speculation)

Three facts:

1. **The buffer is one single tensor.** vLLM's `instanttensor_weights_iterator` (weight_utils.py:1130) passes all 120 shards to ONE `safe_open(copy=True)` with all defaults. With `buffer_size=None`, InstantTensor uses `max(recommended_buffer_size_for_tensors(sizes), io requirement)`, and that recommendation is floored at the largest tensor. I parsed all 120 shard headers (150,226 tensors): the max is exactly **1,268,776,960 B (1210.0 MiB)** — the fused fat-expert tensor in `model-00001-of-00120.safetensors` — and `recommended_buffer_size_for_tensors` over the full list returns exactly 1,268,776,960. The number in the traceback is deterministic, not noise.

[...]

PR #243, opened 2026-09-21
docs: systemd reboot units, digest pin, independent #205 receipts

joelhooks:
## Description and Motivation

`./start.sh` docker-runs both ranks detached (restart policy **no**), polls `/health`, and exits. That is the right launcher. It is not reboot persistence.

This PR adds optional units under `examples/systemd` copied from a live 2x GB10 pair that survived a dual-node power cycle:

- worker watcher starts first (`Restart=no`, `docker wait`)
- head SSHes `systemctl restart` of that watcher (`StrictHostKeyChecking=yes`), runs `./start.sh`, then `docker wait`s the head container
- `ExecStartPost` matches HTTP `/health` and `/v1/models`, not a log grep
- Docker `--restart` stays off so the two ranks cannot race

It also documents production facts from that pair **without changing launcher defaults**:

- pin `IMAGE` by digest plus `SKIP_PULL=1 SKIP_BUILD=1`; do not `git pull` on the serving checkout
- independent #205 reproduction: InstantTensor `buffer_size (1268776960 B) exceeds device memory budget (693237760 B)` after this kit's own ~164 GiB rsync. GB10 counts page cache as used to `cudaMemGetInfo` while `MemAvailable` still passes preflight
- workaround that booted: drop page cache, `CG_ESTIMATE=0`, `GPU_MEM_UTIL=0.87`. **The shipped `0.85` default stays.** 0.87 is not host-headroom policy (256k prefill OOM on 2026-09-06)
- loader/hygiene belongs in #230, not a competing overlay here
- packed ~119/121 GiB RAM after a successful EXL3 boot is expected
- `/health` was 4-6 minutes on a cached power-cycle start; allow 10-12 minutes; DFlash2 warmup continues after HTTP 200

Client SSH tunnels and operator desktop routing are not in these units.

## Related Issues

- Related: #205 (independent reproduction + operational workaround; this PR does **not** fix the InstantTensor budget)
- Related: #204 (KV shortfall that `CG_ESTIMATE=0` recovered on this pair)
- Does not replace: #230 (UMA cold load / `GLM53_HOST_MEM_HYGIENE`)
- Does not replace: #242 (InstantTensor KV reservation default)

## Testing

- [x] `python3 tests/test_systemd_examples.py` (`bash -n` on the helpers, unit invariants, HTTP readiness, host-key checking, `GPU_MEM_UTIL` default left at 0.85)
- [x] Live pair already running these units: dual-node power cycle, worker then head, containers 0 Docker restarts, no OOM, RoCE up, `/health` 200, TP=2 DFlash2 warmup 24/24
- [x] Defaults unchanged (`GPU_MEM_UTIL=0.85`, `CG_ESTIMATE=1`, `PORT=8888`, container names `glm53-exl3-*`)
- [ ] This PR does **not** restart the live serving pair to re-prove the units against current `main` (that checkout is 51 commits behind and a pull would rebuild)

## Checklist:


joelhooks:
Opened in error. Not intending to upstream operator units.

